Issue #403, opened 2023-06-01
Incorporate measurement operator calibration in qililab


anapalu:
### Feature details

After the binary classification has been performed for each individual qubit and we obtain the respective bitstring that was measured for that shot, we can correct for unwanted correlations present in the hardware set up by applying a linear correction to the resulting bitstring probabilities. We can find the relevant weights for this linear correction by running a calibration experiment, which consists on running circuits that prepare each state of the computational basis and looking at the bitstrings that we actually get. We can fit these to the probabilities that we should have had of binding each bitstrings for each of these circuits, and voilà. 
This further calibration step was not needed until the process tomography experiment, but I think we ended up suspecting it would be relevant for the successful outcome of CHSH too. Thus, in general it may a good idea to have it incorporated.  In any case, definitely talk to Ramiro about this also in order to see how relevant it may be and set it in the right priority in your to-do list. 

### Implementation

The coefficient matrix to be inverted scales as (2^N - 1) x 2^N though, so the exact solution of this problem is not really scalable, but we can probably find a more efficient procedure to do this.

The code produced by the hackathon for the measurement calibration experiment, which Juanjo and Victor helped develop: [](https://github.com/qilimanjaro-tech/HardwareDataTools/blob/main/src/hwdatatools/experiment_portfolio/measurement_calibration_experiment.py)
In that script the weights relate to the Pauli Z-operators rather than back to the bitstrings themselves, but the problem is equivalent and we can jump from one to the other. The analysis part as it is written there also needs to be generalised to the n-qubit case.

A possible work-around the scaling issue could be calibrating pairs of measured bit-strings (or a small number), basically assuming that undesired correlations will decrease as we compare more distant qubits. But again, run this by Ramiro first.  

### How important would you say this feature is?

None

### Additional information

With this correction, as I was mentioning on the previous issue ( **#402 Add measured binary strings by binning in results.acquisitions()**  ), the outcome bitstring of the single-shot measurement can be replaced by an array of probabilities that we are actually looking at any of the possible bitstrings, which can then be naturally carried though to the final probabilities of  obtaining each bitstring for each circuit.

Issue #402, opened 2023-06-01
Add measured binary strings by binning in results.acquisitions()


anapalu:
### Feature details

This feature is relevant for multi-qubit experiments. It would be nice that the first thing that is done after thresholding is collect the individual bits and compose the full binary string, and provide that as the result of that measurement. I don't know if this is automatically done by the QCM when the hardware average is set to more than 1. I assume that it maybe does because if it just averages over single-qubit measurements we will miss the correlations of a single shot, but otherwise I assume that this feature of hardware average is only useful for single-qubit experiments and should be set to 1 otherwise.  

### Implementation

I wrote some dirty code to acquire this data from the results.acquisitions() structure as it is now (I believe, I haven't been able to test it though), which I attach here. 
![Screenshot from 2023-06-01 11-31-52](https://github.com/qilimanjaro-tech/qililab/assets/75974413/c5913965-a768-4534-8ccf-784d91074ffa)
In this example the column named 'bits' is where the classified bits for each qubit goes. I also haven't put the resulting means in the dataframe explicitly because I was ultimately interested in the expectation values of the Pauli observables (I have opened a different issue for this:  **#403: Incorporate measurement operator calibration in qililab** ). 
However, as I mentioned before it would probably be best to do the regrouping of the classified 0s and 1s into bit strings and work with the full string as the output of the shot as the very first thing.

### How important would you say this feature is?

None

### Additional information

I understand this was maybe already in the plans at least with the implementation of binning, but since it had to be done in a rush there probably was no time to set this up nice and pretty. I'll leave here the fields I found I would certainly like the `results.acquisitions()` dataframe to have after last week's emergency testing: `'circuit_number'`,` [f'loop_index_{x}' for x in range(loops)]`, `'bin_index'`, `[x for x in itertools.product(['0', '1'], repeat=num_qubits)]`. This last series of columns would contain a 0 or a 1 depending on whether that turned out to be the recorded outcome of the shot. Alternatively, provided the incorporation of the calibration of the measurement operators (see issue **#403: Incorporate measurement operator calibration in qililab**) maybe it could contain the probabilities that the outcome really was each of those strings.

PR #414, opened 2023-06-08
dfs fixed

GuillermoAbadLopez:
## ORIGINALY
In our DataFrames, perviously we had a normal index, and then columns to represent the different index before concatenating df's, such as:
- `loop_indexes`
- `results_index`
- `acquisition_index`
- `bin_index`

which were just copies of the previous indices, and the name after concatenating should now be read as:
`acquisition_index` = index this sub-df had when we were at acquisition df's level (before concatenating at acquisition level)

## NOW
But after this PR, we will also have columns representing in which of those concatenated objects they were:
- `circuit_index`
- `qubit_index`
- `bins_index`

So now for each level, we have two types of indices, ones that tells you in which `df` of the `list[df]` they were [new added indices], and another one that tells you the previous indices they had inside those df (before concatenation) [old indices].

Notice that `bins_index `is in both, this is because it is the lowest level we can go, if we could go even lower, this would be split into two indices too.

Screenshot 2023-05-26 at 21 23 23

## REDUNDANCY 
And actually now, with these new columns there is redundancy in the information of the dataframe, since:
- `df['qubit_number']=(df['acquisition_index'] - df['bin_index']) / num_bin)`
- `df['circuit_number']=(df['acquisition_index'] - df['bin_index']) / (num_qubits*num_bin)`

But I still consider worth it for small df, since it is way more visual now. (For larger DF, we should consider reducing this redundancy)

## LINKED PRs/Issues

**PR on dummy_pulsar**
https://github.com/qilimanjaro-tech/qililab-tools/pull/13
**Ana's Issues**
https://github.com/qilimanjaro-tech/qililab/issues/402
https://github.com/qilimanjaro-tech/qililab/issues/403

## TO DO:
- [x] Trespass everything from new-main PR to this PR
- [x] Delete `acquisition_index`
- [ ] Think what to do with `result_index`
- [ ] Check `to_dataframe()` method (You removed `result_index`)
- [x] Check that `loop_indexs` work correctly
- [x] Fix tests
- [ ] Understand the structure of the dictionary in `data.py`, to change and make the test more complex 
- [x] Add threshold column, at the same level of the q,i 
(you need to see how to pass `threshold` parameters into `binary_classification` in qblox library)
- [ ] Decide until whihc depth is gonna be threshold or classification.
- [ ] Check if deleting the `sequence_index` & `software_avg_index` columns 
- [ ] In the `test_qblox_result` implement/delete commented code after understanding it
- [ ] Edit the description of this PR and the changelog
- [ ] Check docstring and edit them if necessary
- [x] Check compatibility of the ideas from Ana to this PR!


GuillermoAbadLopez:
Outdated task

tests/unit/experiment/test_qblox_result.py
```diff
@@ -115,7 +132,7 @@ def fixture_qblox_asymmetric_bins_result():
             },
             "bins": {
                 "integration": {"path0": [0.0, 0.0, 0.0], "path1": [0.0, 0.0, 0.0]},
-                "threshold": [1.0, 0.0, 1.0],
+                "binary_classification": [1.0, 0.1, 1.0],
```
JavierSab, line 135: is it supposed to be 1.0 instead of 0.1?

PR #500, opened 2023-08-11
Added methods to execute qprograms

marcoparadina:
This PR adds a method to execute a qprogram from the platform. That method calls other methods, also added in this PR, in the lower level classes: BusDriver, AWG, SequencerQCM. These methods handle the execution of the qpysequence of each bus, which is created by the QBloxCompiler


fedonman:
Closing since these changes were added in a different PR.

PR #550, opened 2023-10-03
Fix gates with negative amplitudes

AlbertMitjans:
For some unknown reason gates with negative amplitudes are not working properly with Qblox.

A workaround we can take is, when transpiling gates into pulses, add `np.pi` to the phase if the amplitude is negative.


AlbertMitjans:
I have tested this in HW 😄

PR #552, opened 2023-10-04
Cphase phase correction

visagim:
- Changed default value for optimize in translate_circuit from `False -> True`
- Add CZ gate phase correction from tomography calibration as virtual z gate

- [x] Fix unit tests
- [x] Add changelog
- [x] Fix docstrings

### DIFF
--- a/src/qililab/settings/gate_event_settings.py
+++ b/src/qililab/settings/gate_event_settings.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from qililab.typings.enums import Parameter
 
@@ -46,6 +46,11 @@ class GateEventSettings:
         phase: float
         duration: int
         shape: dict
+        options: dict = field(default_factory=dict)
+
+        def __post_init__(self):
+            for key, value in self.options.items():
+                setattr(self, key, value)
 
     bus: str
     pulse: GatePulseSettings
--- a/src/qililab/settings/gate_event_settings.py
+++ b/src/qililab/settings/gate_event_settings.py
@@ -40,6 +40,7 @@ class GateEventSettings:
             phase (float): phase of the pulse
             duration (int): pulse duration
             shape (dict): pulse envelope
+            options (dict): optional gate dependant parameters
         """
 
         amplitude: float
--- a/src/qililab/settings/gate_event_settings.py
+++ b/src/qililab/settings/gate_event_settings.py
@@ -50,6 +50,8 @@ class GateEventSettings:
         options: InitVar[dict | None] = None
 
         def __post_init__(self, options):
+            # load options if options are defined. This is done so that if options are undefined then no options parameter is created
+            # and serialization of the runcard is consistend / doesn't create an empty options field.
             if options is not None:
                 self.options = options
 
--- a/src/qililab/settings/gate_event_settings.py
+++ b/src/qililab/settings/gate_event_settings.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from dataclasses import dataclass, field
+from dataclasses import InitVar, dataclass
 
 from qililab.typings.enums import Parameter
 
@@ -46,11 +46,11 @@ class GateEventSettings:
         phase: float
         duration: int
         shape: dict
-        options: dict = field(default_factory=dict)
+        options: InitVar[dict | None] = None
 
-        def __post_init__(self):
-            for key, value in self.options.items():
-                setattr(self, key, value)
+        def __post_init__(self, options):
+            if options is not None:
+                self.options = options
 
     bus: str
     pulse: GatePulseSettings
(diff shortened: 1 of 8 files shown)


visagim:
Forgot to fix docstrings :sweat_smile:

jjmartinezQT (changes requested):
Good work! A minor comment. I would also like to discuss the gate settings with rest of the team before merging this one to at least agree on a standard, as this might not be the only gate with options in the future.

src/qililab/transpiler/transpiler.py
```diff
@@ -68,6 +86,22 @@ def optimize_transpilation(nqubits: int, ngates: list[gates.Gate]) -> list[gates
...
             shift[gate.qubits[0]] += gate.parameters[0]
+        # add CZ phase correction
+        elif isinstance(gate, gates.CZ) and gates_settings is not None:
+            gate_settings = gates_settings.get_gate(name=gate.__class__.__name__, qubits=gate.qubits)
+            control_qubit, target_qubit = gate.qubits
+            corrections = next(
+                (
+                    event.pulse.options
+                    for event in gate_settings
+                    if f"q{control_qubit}_phase_correction" in event.pulse.options
```
AlbertMitjans, line 97: This is the problem of using generic kwargs...

Why did we choose this option over adding this the phase shift option in the dataclass?

src/qililab/settings/gate_event_settings.py
```diff
@@ -40,12 +40,18 @@ class GatePulseSettings:
...
 
         amplitude: float
         phase: float
         duration: int
         shape: dict
+        options: InitVar[dict | None] = None
+
+        def __post_init__(self, options):
+            if options is not None:
+                self.options = options
```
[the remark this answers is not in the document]
visagim: this is to avoid runcard serialization from failing without adding an empty `options` field to every single gate

src/qililab/transpiler/transpiler.py
visagim: it is to keep `gate_settings` general and avoid having a part of the code specifically handling `CZ` gates. I wanted to avoid discriminating between gates because this is what happened when we had `harware_gates` which ended up making gate definitions too rigid.  Perhaps it'd be nice to discuss this in code discussions

src/qililab/settings/gate_event_settings.py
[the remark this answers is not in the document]
AlbertMitjans: Ok! Now I understand 👍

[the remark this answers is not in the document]
AlbertMitjans: Please add a comment explaining this 😄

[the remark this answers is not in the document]
AlbertMitjans: Remember that everything we write should be understandable for other developers!

src/qililab/transpiler/transpiler.py
AlbertMitjans: Why not adding a phase correction as a generic feature in the gate transpilation? Meaning it can be added for any gate the user defines.

AlbertMitjans: Something like the following:

```yml
CZ(0,2):
- bus: flux_line_q2_bus
  pulse:
    amplitude: 1.0
    phase: 0
    duration: 101
    shape:
      name: snz
      t_phi: 1
      b: 0.5
    phase_correction:
      - qubit: 0
         phase: 0.1
       - qubit: 1
         phase: 0.4
```

AlbertMitjans: Where the `phase_correction` can be added to any gate/pulse

visagim: phase correction is specific to 2q gates, and specifically for the qubit gates we're implementing (based on interactions with the qubits' avoided crossing). So it doesn't make a lot of sense to add a phase correction for other gates.
With the generic "options" i wanted to avoid both this:
> it is to keep gate_settings general and avoid having a part of the code specifically handling CZ gates. I wanted to avoid discriminating between gates because this is what happened when we had harware_gates which ended up making gate definitions too rigid. Perhaps it'd be nice to discuss this in code discussions
mentioned above but also the opposite: adding too many parameters to gate_settings which in many cases might be empty.

Personally I don't think it is a problem to allow gate definitions to be as flexible as possible - but I do agree that the `options` method I propose here is uncertain so we'd need documentation to inform and be clear of the possible options.
I envision these `options` parameter as a similar case as options in [`scipy.minimize`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html) or the config options in[ dwave's sampler method](https://github.com/dwavesystems/dwave-system/blob/1.20.0/dwave/system/samplers/dwave_sampler.py#L91) in both cases, there's an arbitrary dictionary of options (because they are method dependent) but this is not a problem because there's a lot of documentation supporting it (more so in scipy's case)

AlbertMitjans: I agree! Let's make sure we add documentation for this then!

I believe it makes sense to mention this in the Runcard page of the documentation. Could you create a task in the backlog to make sure this is mentioned?

AlbertMitjans: If we do this, then I have no problem merging this PR

jjmartinezQT (approved):

PR #553, opened 2023-10-05
Chip documentation/tas 303

visagim:
- [x] changelog


visagim:
I think I just forgot to push the last changes I had (at the `__init__.py` with most of the explanation for the `chip` module). Let me know if there are any classes with missing documentation

src/qililab/chip/chip.py
```diff
@@ -24,7 +27,10 @@
 
 @dataclass
 class Chip:
-    """Chip representation as a graph."""
+    """Chip representation as a graph.
+    This class represents the chip structure of in the runcard and contains all the connectivity information of
+    the chip.
+    """
 
     nodes: list[Node]
```
jjmartinezQT, line 35: This attribute has no definition in the documentation. Could you include it, please?

src/qililab/chip/__init__.py
```diff
@@ -13,7 +13,24 @@
 # limitations under the License.
 
 """
-This module contains the classes needed to define a quantum chip.
+The chip class contains information about the chip and its conections. This information can be reached through the Platform class through a platform object by accessing platform.chip.
+
+Each node in the chip can be a Coil, Coupler, Port, Qubit or Resonator. Each node also has an alias assigned to it. For example, the alias for Qubit nodes is typically ``qubit_n`` where n is the qubit index.
+
+The qubit connectivity (chip topology) can be accessed by calling chip.get_topology(), which returns a networkx graph of the qubits Eg.
```
jjmartinezQT, line 20: the abbreviated of 'for example' is written in lower case and separated by dots in between, often with a comma before it to separate it from the rest of the sentence. I would rewrite this like:

"The qubit connectivity (chip topology) can be accessed by calling chip.get_topology(), which returns a networkx graph of the qubits, e.g."

src/qililab/chip/node.py
```diff
@@ -21,7 +21,9 @@
 
 @dataclass(kw_only=True)
 class Node(Settings, FactoryElement):
-    """Class representing a node of the chip's graph."""
+    """This class is used to represent a node of the chip's graph.
+    Each node is an element of the chip
+    """
 
     alias: str
```
jjmartinezQT, line 28: This attributes are missing definitions in the docs.

jjmartinezQT: Same with Port class and its attributes, Qubit, Resonator, Coupler and Coil. Also this classes could benefit from better definition of what the class is about.

jjmartinezQT (changes requested):

src/qililab/chip/chip.py
jjmartinezQT: Addressed.

src/qililab/chip/__init__.py
jjmartinezQT: done.

src/qililab/chip/node.py
jjmartinezQT: First pass at this done

jjmartinezQT (approved):

Issue #556, opened 2023-10-09
[BUG] system_control.compile() should not try to compile when S4g is in the bus


paulqili:
### Expected behavior

`system_control.compile()` should skip non AWG type instruments

### Actual behavior

`system_control.compile()` tries to compile for all instruments in the bus, now we use S4g instruments in the bus which are not instances of `AWG` and we get an `AttributeError`

See image:

<img width="1459" alt="image" src="https://github.com/qilimanjaro-tech/qililab/assets/113030828/e14967c0-91e0-4b39-9dfa-d928af7488b1">

### Existing GitHub issues

- [X] I have searched existing GitHub issues to make sure the issue does not already exist.

jjmartinezQT:
This has been fixed and merged into main.

PR #557, opened 2023-10-09
returning none if no AWG instrument is present in bus

jjmartinezQT:
This branch fixes

https://github.com/qilimanjaro-tech/qililab/issues/556#top

### DIFF
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -317,5 +317,8 @@ This document contains the changes of the current release.
 
 ### Bug fixes
 
+- Removing raised error if the system control has no AWG instruments and tries to compile and returning no compiled programs instead.
+  [#556](https://github.com/qilimanjaro-tech/qililab/pull/bug-557)
+
 - The `threshold` and `threshold_rotation` parameters of a `QbloxQRM` can now be set using `Platform.set_parameter`.
   [#534](https://github.com/qilimanjaro-tech/qililab/pull/534)
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -300,6 +300,9 @@ This document contains the changes of the current release.
       nodes: [qubit_2, resonator_q0, drive_line_q0, flux_line_q0]
   ```
 
+- `ql.execute` now accepts a list of circuits!
+  [#549](https://github.com/qilimanjaro-tech/qililab/pull/549)
+
 ### Breaking changes
 
 - Old scripts using `Experiment` with circuits should be changed and use `CircuitExperiment` instead.
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -4,6 +4,9 @@ This document contains the changes of the current release.
 
 ### New features since last release
 
+- Update qiboconnection to 0.12.0
+  [#559](https://github.com/qilimanjaro-tech/qililab/pull/559)
+
 - Added phase correction for CZ gates to the optimize step of translate circuit in `qililab.transpiler.transpiler`. Gates now can accept an optional dicionary with additional settings.
   As an example, the CZ phase correction can be added at options for each qubit:
 
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -347,7 +347,7 @@ This document contains the changes of the current release.
 
 ### Bug fixes
 
-- Removing raised error if the system control has no AWG instruments and tries to compile and returning no compiled programs instead.
+- Avoid creating empty sequences for buses that are no flux lines and do for flux ones that do not have any AWG instrument.
   [#556](https://github.com/qilimanjaro-tech/qililab/pull/bug-557)
 
 - The `threshold` and `threshold_rotation` parameters of a `QbloxQRM` can now be set using `Platform.set_parameter`.
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -4,6 +4,29 @@ This document contains the changes of the current release.
 
 ### New features since last release
 
+- Added phase correction for CZ gates to the optimize step of translate circuit in `qililab.transpiler.transpiler`. Gates now can accept an optional dicionary with additional settings.
+  As an example, the CZ phase correction can be added at options for each qubit:
+
+  ```yml
+  CZ(0,2):
+  - bus: flux_line_q2_bus
+    pulse:
+      amplitude: 1.0
+      phase: 0
+      duration: 101
+      shape:
+        name: snz
+        t_phi: 1
+        b: 0.5
+      options:
+        q0_phase_correction: 0.1
+        q2_phase_correction: 0.2
+  ```
+
+  The default value for the `optimize` flag in qililab.transpiler.transpiler.translate_circuit has been changed from `False` to `True`
+
+  [#552](https://github.com/qilimanjaro-tech/qililab/pull/552)
+
 - build_platform() has been extended: [#533](https://github.com/qilimanjaro-tech/qililab/pull/533)
 
   Now appart from passing the runcard YAML file path, you can directly pass an already build dictionary.
@@ -314,6 +337,10 @@ This document contains the changes of the current release.
   [#494](https://github.com/qilimanjaro-tech/qililab/pull/494)
 
 ### Documentation
+
+- Documentation for the Pulse module: [#532](https://github.com/qilimanjaro-tech/qililab/pull/532)
+
+  Includes documentation for all public features of the Pulse module
 
 ### Bug fixes
 
--- a/src/qililab/platform/platform.py
+++ b/src/qililab/platform/platform.py
@@ -477,9 +477,6 @@ class Platform:  # pylint: disable = too-many-public-methods, too-many-instance-
         for pulse_bus_schedule in pulse_schedule.elements:
             bus = self.buses.get(port=pulse_bus_schedule.port)
             bus_programs = bus.compile(pulse_bus_schedule, num_avg, repetition_duration, num_bins)
-            # ensuring that buses without AWG type instruments but with non AWG instruments don't get added to the
-            # programs queue
-            if bus_programs:
-                programs[bus.alias] = bus_programs
+            programs[bus.alias] = bus_programs
 
         return programs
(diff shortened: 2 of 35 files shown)


src/qililab/platform/platform.py
```diff
@@ -477,6 +477,9 @@ def compile(self, program: PulseSchedule | Circuit, num_avg: int, repetition_dur
         for pulse_bus_schedule in pulse_schedule.elements:
             bus = self.buses.get(port=pulse_bus_schedule.port)
             bus_programs = bus.compile(pulse_bus_schedule, num_avg, repetition_duration, num_bins)
-            programs[bus.alias] = bus_programs
+            # ensuring that buses without AWG type instruments but with non AWG instruments don't get added to the
+            # programs queue
+            if bus_programs:
+                programs[bus.alias] = bus_programs
```
AlbertMitjans, line 483: Wait, why is there a pulse_bus_schedule created for the bus without an awg??

AlbertMitjans: The problem might be somewhere else then!

jjmartinezQT: The fix returns None if there is no AWG present, but if we don't check in here it will create an entry in the dictionary with an empty list, I wanted to avoid that.

jjmartinezQT: If that is no problem, then I can just undo this one.

AlbertMitjans: Here is a small summary of what I think is happening:

For the Bluefors setup, we use QbloxQCM instruments to apply offsets to all the qubits. QbloxQCMs are not supposed to do this, that is why we have voltage/current sources (don't ask me why are we using them for this purpose, that's a HW decision 😆 ). 

Whatever, a big problem we saw is that the offsets of the QbloxQCMs are only activated when you are actually running a program (thus they cannot be used as constant voltage/current sources).

To solve this, we decided to create "empty programs", which are Qblox programs that don't do anything, they just have a wait time equal to all the other programs that are being executed in the other sequencers. This way, we can activate the offsets while running pulse sequences.

What I believe is happening here is that we are still creating this empty programs, and we are trying to send them to the SPIRack, which of course complains, because it is not an AWG.

jjmartinezQT: There are two places where this happens and fails, compile level and run level. I have removed the raised error in both, run is fine just by doing return if no AWG instrument is present, as the return type is not used anywhere. However compiling does something with the returned compiled program, which right now is None after the fix, but it could be an empty list. Whatever the return type is, it will get added to the dictionary of programs per bus, so I think we should avoid that.

AlbertMitjans: So, long story short, I wouldn't remove the error from the bus! What this error is saying (or should say) is that `Bus.compile` cannot be called when the bus doesn't contain an AWG. 

The actual error comes from before, during the transpilation:

https://github.com/qilimanjaro-tech/qililab/blob/0d2eb761fbde9e2bfb96cc4c61dbe614a803e3b9/src/qililab/pulse/circuit_to_pulses.py#L124-L129

This 5 lines create the empty pulse schedules that I was talking about. What I would do is add an `if` statement that only runs these lines when the flux buses are connected to AWGs.

jjmartinezQT: Obviously we should avoid to create empty programs, but not sure this hotfix is meant to tackle that bigger task, as I am not sure the implications of removing them, as you mentioned they were created as a way to by-pass certain HW limitations.

AlbertMitjans: We can discuss it in more detail after the backlog refinement meeting!

AlbertMitjans: > Obviously we should avoid to create empty programs, but not sure this hotfix is meant to tackle that bigger task, as I am not sure the implications of removing them, as you mentioned they were created as a way to by-pass certain HW limitations.

It is not a bigger task! It should be an extra 3 lines 😄 

It is very important that you understand the problem we are having here! Otherwise it would be impossible to fix it. The real problem is that we are creating empty schedules when we shouldn't be creating them (because of everything I mentioned before).

jjmartinezQT: Understood after meeting to understand the whole issue better with empty schedules.

tests/pulse/test_circuit_to_pulses.py
```diff
@@ -463,11 +463,7 @@ def test_translate(self, platform):  # pylint: disable=R0914 # disable pyling to
 
         pulse_schedule = pulse_schedules[0]
         # there are 6 different buses + 3 empty for unused flux lines
-        assert len(pulse_schedule) == 9
-        assert all(len(schedule_element.timeline) == 0 for schedule_element in pulse_schedule.elements[-3:])
-
-        # we can ignore empty elements from here on
-        pulse_schedule.elements = pulse_schedule.elements[:-3]
+        assert len(pulse_schedule) == 6
```
AlbertMitjans, line 466: In this test we don't know if the flux buses contains AWGs or not.

I would like to have dedicated tests for this, where you modify the instruments inside a bus by adding/removing an AWG, and you check explicitly that the empty pulse schedules are added or not

jjmartinezQT: fair enough, including those tests.

jjmartinezQT: unittest added to check that adding/removing buses work as expected, please review that the buses added with the circuit in the unittest makes sense in the number of pulse schedules created.

tests/pulse/test_circuit_to_pulses.py
```diff
@@ -462,17 +503,14 @@ def test_translate(self, platform):  # pylint: disable=R0914 # disable pyling to
...
-        # we can ignore empty elements from here on
-        pulse_schedule.elements = pulse_schedule.elements[:-3]
+        # there are 7 pulse schedules
+        assert len(pulse_schedule) == 7
 
         # extract pulse events per bus and separate measurement pulses
         pulse_bus_schedule = {
             pulse_bus_schedule.port: pulse_bus_schedule.timeline for pulse_bus_schedule in pulse_schedule
         }
+
```
AlbertMitjans, line 513: This shouldn't change, right? 🤔

jjmartinezQT: That is what I don't fully understand from the mocked runcard, there are two schedules less being created. If any of the flux lines do not have an AWG there should not be empty schedules added, right?

tests/pulse/test_circuit_to_pulses.py
```diff
@@ -464,15 +512,12 @@ def test_translate(self, platform):  # pylint: disable=R0914 # disable pyling to
         pulse_schedule = pulse_schedules[0]
         # there are 6 different buses + 3 empty for unused flux lines
         assert len(pulse_schedule) == 9
-        assert all(len(schedule_element.timeline) == 0 for schedule_element in pulse_schedule.elements[-3:])
-
-        # we can ignore empty elements from here on
-        pulse_schedule.elements = pulse_schedule.elements[:-3]
```
AlbertMitjans, line 470: Let's move this back! 🙏

tests/pulse/test_circuit_to_pulses.py
```diff
@@ -439,6 +439,54 @@ def get_bus_schedule(self, pulse_bus_schedule: dict, port: str) -> list[dict]:
...
+        circuit.add(M(0))
+        circuit.add(CZ(2, 3))
+        circuit.add(CZ(4, 0))
+        circuit.add(M(*range(4)))
+        circuit.add(Wait(0, t=10))
+        circuit.add(Drag(0, 2, 0.5))
+
+        pulse_schedules = translator.translate(circuits=[circuit])
+        pulse_schedule = pulse_schedules[0]
+        # there should be 8 pulse_schedules in this configuration
```
AlbertMitjans, line 459:
```suggestion
        # there should be 9 pulse_schedules in this configuration
```

tests/pulse/test_circuit_to_pulses.py
```diff
@@ -439,6 +439,54 @@ def get_bus_schedule(self, pulse_bus_schedule: dict, port: str) -> list[dict]:
...
+            "delay": 0,
+        }
+
+        platform.buses.add(
+            Bus(settings=flux_bus_no_awg_settings, platform_instruments=platform.instruments, chip=platform.chip)
+        )
+        pulse_schedules = translator.translate(circuits=[circuit])
+        pulse_schedule = pulse_schedules[0]
+        # there should not be any extra pulse schedule added
+        assert len(pulse_schedule) == 8
```
AlbertMitjans, line 488: Nice test!! 😍

tests/pulse/test_circuit_to_pulses.py
AlbertMitjans: Amazing, thanks!!

tests/pulse/test_circuit_to_pulses.py
jjmartinezQT: addressed.

docs/releases/changelog-dev.md
```diff
@@ -347,5 +347,8 @@ This document contains the changes of the current release.
 
 ### Bug fixes
 
+- Removing raised error if the system control has no AWG instruments and tries to compile and returning no compiled programs instead.
```
AlbertMitjans, line 350: Please update the changelog entry! 🙏

tests/pulse/test_circuit_to_pulses.py
```diff
@@ -462,17 +510,18 @@ def test_translate(self, platform):  # pylint: disable=R0914 # disable pyling to
...
 
         pulse_schedule = pulse_schedules[0]
-        # there are 6 different buses + 3 empty for unused flux lines
-        assert len(pulse_schedule) == 9
         assert all(len(schedule_element.timeline) == 0 for schedule_element in pulse_schedule.elements[-3:])
 
         # we can ignore empty elements from here on
         pulse_schedule.elements = pulse_schedule.elements[:-3]
+        # there are 6 different buses + 3 empty for unused flux lines
+        assert len(pulse_schedule) == 6
```
AlbertMitjans, line 518: This whole test should remain unchanged!

docs/releases/changelog-dev.md
jjmartinezQT: good catch, changed.

AlbertMitjans (approved):
Nicee! Amazing job!

PR #558, opened 2023-10-09
Fix cz in transpiler

visagim:


AlbertMitjans (approved):
Thanks! Could you maybe add one test in `test_transpiler.py` with swapped qubits to make sure this is fixed? 🙏

visagim:
Did we change anything in the configuration or is this just a bug?
image

PR #561, opened 2023-10-11
fix: :bug: Raise error if the alias of an instrument (in SystemControl) is not found in platform.

AlbertMitjans:
This PR fixes a bug where, if the user added an instrument alias inside a system control that was not present in the platform, no error was raised and no instrument was added in the bus.

```yaml
- alias: drive_line_q4_bus
  system_control:
    name: system_control
    instruments: [Unknown-alias]
  port: drive_line_q4
  distortions: []
```
Now an error is raised to make sure the user knows that the alias of the instrument doesn't exist.


jjmartinezQT (approved):
Tested and working. Good work.

PR #562, opened 2023-10-11
Fix gate settings serialization

visagim:
FIxes gate settings serialization so that gate.pulse.options is serialized and fields with None values are removed from the resulting dictionary

### DIFF
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -350,5 +350,8 @@ This document contains the changes of the current release.
 
 ### Bug fixes
 
+- Avoid creating empty sequences for buses that are no flux lines and do for flux ones that do not have any AWG instrument.
+  [#556](https://github.com/qilimanjaro-tech/qililab/pull/bug-557)
+
 - The `threshold` and `threshold_rotation` parameters of a `QbloxQRM` can now be set using `Platform.set_parameter`.
   [#534](https://github.com/qilimanjaro-tech/qililab/pull/534)
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -4,6 +4,9 @@ This document contains the changes of the current release.
 
 ### New features since last release
 
+- Changed gate settings serialization so that fields with None values are not in the resulting dictionary
+  [#562](https://github.com/qilimanjaro-tech/qililab/pull/562)
+
 - Update qiboconnection to 0.12.0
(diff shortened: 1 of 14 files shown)


src/qililab/settings/gate_event_settings.py
```diff
@@ -47,13 +48,7 @@ class GatePulseSettings:
...
         duration: int
         shape: dict
-        options: InitVar[dict | None] = None
-
-        def __post_init__(self, options):
-            # load options if options are defined. This is done so that if options are undefined then no options parameter is created
-            # and serialization of the runcard is consistend / doesn't create an empty options field.
-            if options is not None:
-                self.options = options
+        options: Optional[dict] = None
```
AlbertMitjans, line 51: Yaay! 🎉

docs/releases/changelog-dev.md
```diff
@@ -4,6 +4,9 @@ This document contains the changes of the current release.
 
 ### New features since last release
 
+- Changed gate settings serialization so that fields with None values are not in the resulting dictionary
+  [#562](https://github.com/qilimanjaro-tech/qililab/pull/562)
+
```
AlbertMitjans, line 9: There should be a test testing this!!

AlbertMitjans (commented):
Good job!! Please don't forget to add tests when fixing bugs / adding new functionalities!!

docs/releases/changelog-dev.md
visagim: there is! the runcard in data.galadriel has
https://github.com/qilimanjaro-tech/qililab/blob/18b269e61dabfabaf6f32548a0aabf3c9bcab8ab/tests/data.py#L142-L163
so the object created in `platform.gate_settings.gates` will have one gate containing options, the other containing `None`. If the serialization works as intended then the test at
https://github.com/qilimanjaro-tech/qililab/blob/18b269e61dabfabaf6f32548a0aabf3c9bcab8ab/tests/settings/test_runcard.py#L40C87-L40C87
will pass, otherwise it won't because it will have the extra field `{options: None}` for gate `CZ(0, 2)`

visagim: the problem is that I originally didn't add these `options` field in the `tests` runcard

AlbertMitjans: Nice! I forgot about this test!!

AlbertMitjans (approved):
Approved!

PR #563, opened 2023-10-11
[TAS-386] Create Quantum Machines Compiler for QProgram

fedonman:


AlbertMitjans:
* A link back to this PR.

jjmartinezQT (dismissed):
Some initial comments on the quantum_machines_compiler class only for now.

src/qililab/qprogram/quantum_machines_compiler.py
```diff
@@ -0,0 +1,576 @@
...
+
+        Args:
+            qprogram (QProgram): The QProgram to be compiled
+            bus_mapping (dict[str, str] | None, optional): Optional mapping of bus names. Defaults to None.
+
+        Returns:
+            tuple[qua.Program, dict, list[MeasurementInfo]]: A tuple consisting of the generated QUA program, the generated configuration and a list of measurements' information.
+        """
+
+        def traverse(block: Block):
```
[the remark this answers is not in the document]
fedonman: Why do you say that closure is slower? I have never seen such statement.

src/qililab/qprogram/quantum_machines_compiler.py
```diff
@@ -0,0 +1,576 @@
...
+        variables = []
+        loops = []
+        for loop in element.loops:
+            qua_variable = self._qprogram_to_qua_variables[loop.variable]
+            values = (
+                np.arange(loop.start, loop.stop + loop.step / 2, loop.step)
+                if isinstance(loop, ForLoop)
+                else loop.values
+            )
+            if loop.variable.domain is Domain.Phase:
```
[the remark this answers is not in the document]
fedonman: Fixed, I added class constants

src/qililab/qprogram/quantum_machines_compiler.py
```diff
@@ -0,0 +1,576 @@
...
+            if loop.variable.domain is Domain.Frequency:
+                values = (values * 1e3).astype(int)
+            if loop.variable.domain is Domain.Time:
+                values = np.maximum(values, 4).astype(int)
+
+            variables.append(qua_variable)
+            loops.append(values)
+        return qua.for_each_(tuple(variables), tuple(loops))
+
+    def _handle_for_loop(self, element: ForLoop):
```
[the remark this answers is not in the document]
fedonman: Fixed, I added class constants

src/qililab/qprogram/quantum_machines_compiler.py
```diff
@@ -0,0 +1,576 @@
...
+            qua.align(*buses)
+        else:
+            qua.align()
+
+    def __add_pulse_to_element_operations(self, element: str, pulse: str):
+        if pulse not in self._configuration["elements"][element]["operations"]:
+            self._configuration["elements"][element]["operations"][pulse] = pulse
+        return pulse
+
+    def __add_or_update_control_pulse_to_configuration(
```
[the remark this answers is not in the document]
fedonman: Yes, because the dictionaries returned are different.

src/qililab/qprogram/quantum_machines_compiler.py
```diff
@@ -0,0 +1,576 @@
...
+
+        if not waveform_variables:
+            waveform_I_name = self.__add_waveform_to_configuration(waveform_I)
+            waveform_Q_name = self.__add_waveform_to_configuration(waveform_Q) if waveform_Q else None
+            pulse_name = self.__add_or_update_control_pulse_to_configuration(waveform_I_name, waveform_Q_name, duration)
+            operation_name = self.__add_pulse_to_element_operations(bus, pulse_name)
+            pulse = operation_name * gain if gain is not None else operation_name
+            qua.play(pulse, bus)
+
+    def _handle_measure(
```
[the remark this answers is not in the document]
fedonman: Yeah, I know this method is huge, but I would like to keep it as it is for now, and not try to break it down/optimize it, mainly because I want all the different branches to be very clear when we try and debug this with the HW people.

src/qililab/qprogram/quantum_machines_compiler.py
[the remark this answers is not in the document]
jjmartinezQT: Forget about this one, was thinking in my times encountering closures in JavaScript.

src/qililab/qprogram/quantum_machines_compiler.py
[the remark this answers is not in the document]
jjmartinezQT: Fair enough.

src/qililab/qprogram/quantum_machines_compiler.py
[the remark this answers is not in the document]
jjmartinezQT: This is good as it is right now after your explanation about signatures, will keep things cleaner.

src/qililab/result/qblox_results/qblox_qprogram_result.py
```diff
@@ -0,0 +1,204 @@
...
+        Returns:
+            Counts: Counts object containing the counts of each state.
+        """
+        return self.qblox_bins_acquisitions.counts()
+
+    def counts(self) -> dict:
+        """Returns a Counts object containing the counts of each state.
+
+        Returns:
+            Counts: Counts object containing the counts of each state.
```
GuillermoAbadLopez, line 149: You have `-> dict`, but then Counts on docstring

src/qililab/result/qblox_results/qblox_qprogram_result.py
```diff
@@ -0,0 +1,204 @@
...
+        )
+        self.qblox_bins_acquisitions = QbloxAcquisitionsBuilder.get_bins(
+            integration_lengths=integration_lengths, qblox_raw_results=qblox_raw_results
+        )
+        self._qblox_scope_acquisition_copy = deepcopy(self.qblox_scope_acquisitions)
+        self.data_dataframe_indices = self.qblox_bins_acquisitions.data_dataframe_indices
+
+    def _demodulated_scope(self, frequency: float, phase_offset: float = 0.0) -> QbloxScopeAcquisitions:
+        """Returns the scope acquisitions demodulated in the given frequency with the given phase offset.
+
```
GuillermoAbadLopez, line 67: missing arguments in method

src/qililab/qprogram/variable.py
```diff
@@ -13,135 +13,79 @@
...
+        return instance
+
+    def __init__(self, domain: Domain = Domain.Scalar):
+        Variable.__init__(self)
+        self._value = None
+        self.domain = domain
 
 
 class FloatVariable(Variable, float):  # type: ignore
     """Float variable. This class is used to define a variable of type float, such that Python recognizes this class
```
GuillermoAbadLopez, line 80: Missing argument of the class (domain) on class level documentation

src/qililab/qprogram/variable.py
```diff
@@ -13,135 +13,79 @@
...
-        super().__init__()
+    def __new__(cls, _: Domain = Domain.Scalar):
+        # Create a new int instance
+        instance = float.__new__(cls, 0.0)
+        return instance
+
+    def __init__(self, domain: Domain = Domain.Scalar):
+        Variable.__init__(self)
+        self._value = None
+        self.domain = domain
```
GuillermoAbadLopez, line 91: missing docstrings for the attriutes

src/qililab/qprogram/variable.py
```diff
@@ -13,135 +13,79 @@
...
 
-    def __init__(self, value: int = 0):
-        self.value: int = value
-        super().__init__()
+    def __new__(cls, _: Domain = Domain.Scalar):
+        # Create a new float instance
+        instance = int.__new__(cls, 0)
+        return instance
+
+    def __init__(self, domain: Domain = Domain.Scalar):
```
GuillermoAbadLopez, line 73: Same as in the FloatVariable

src/qililab/qprogram/qprogram.py
```diff
@@ -185,28 +220,50 @@ def play(self, bus: str, waveform: Waveform | IQPair):
...
+        bus: str,
+        waveform: IQPair,
+        weights: IQPair | tuple[IQPair, IQPair] | tuple[IQPair, IQPair, IQPair, IQPair] | None = None,
+        demodulation: bool = True,
+        save_raw_adc: bool = True,
+    ):
+        """Play a pulse and acquire results.
+
+        Args:
+            bus (str): Unique identifier of the bus.
```
GuillermoAbadLopez, line 255: Missing arguments in docstrings, and if they are optional

src/qililab/qprogram/qprogram.py
```diff
@@ -185,28 +220,50 @@ def play(self, bus: str, waveform: Waveform | IQPair):
...
+            weights (IQPair): Weights used during acquisition.
```
GuillermoAbadLopez, line 256:
```suggestion
            weights (IQPair, optional): Weights used during acquisition. Defaults to None.
```

src/qililab/qprogram/operations/measure.py
```diff
@@ -0,0 +1,37 @@
...
+# limitations under the License.
+
+from dataclasses import dataclass
+
+from qililab.qprogram.operations.operation import Operation
+from qililab.waveforms import IQPair, Waveform
+
+
+@dataclass(frozen=True)
+class Measure(Operation):  # pylint: disable=missing-class-docstring
```
GuillermoAbadLopez, line 22: Missing full class docstring!

src/qililab/instruments/qblox/qblox_qrm.py
```diff
@@ -187,6 +187,14 @@ def acquire_result(self) -> QbloxResult:
         """
         return self.get_acquisitions()
 
+    def acquire_qprogram_results(self, acquisitions: list[str]) -> QbloxQProgramResult:
+        """Read the result from the AWG instrument
```
GuillermoAbadLopez, line 191: missing acquisitonts argument

src/qililab/instruments/qblox/qblox_module.py
```diff
@@ -551,6 +551,15 @@ def reset(self):
         self.clear_cache()
         self.device.reset()
 
+    def upload_qpysequence(self, qpysequence: QpySequence, port: str):
+        """Upload the qpysequence to its corresponding sequencers."""
```
GuillermoAbadLopez, line 555: missing arguments in docstring

src/qililab/instruments/instrument.py
```diff
@@ -192,6 +192,11 @@ def acquire_result(self) -> Result | None:
 
         In some cases this method might do nothing."""
 
+    def acquire_qprogram_results(self, acquisitions: list[str]) -> Result | None:
+        """Acquire results of the measurement.
```
GuillermoAbadLopez, line 196: missing argumetns in docstring

GuillermoAbadLopez (commented):
Did a very quick review mostly about docstrings, which is what can be done quick..

Sorry for not being able to help with the implementation, which is the actual difficult part! 🙏

src/qililab/instruments/qblox/qblox_module.py
```diff
@@ -551,6 +551,15 @@ def reset(self):
...
+        sequencers = self.get_sequencers_from_chip_port_id(chip_port_id=port)
+        for sequencer in sequencers:
+            logger.info("Sequence program: \n %s", repr(qpysequence._program))  # pylint: disable=protected-access
+            self.device.sequencers[sequencer.identifier].sequence(qpysequence.todict())
+            self.device.sequencers[sequencer.identifier].sync_en(True)
+            self.sequences[sequencer.identifier] = (qpysequence, True)
+
```
visagim, line 562: would it be interesting to keep the `if not uploaded` logic as in the `upload` method below? Otherwise there's not much sense on keeping the tuple `(sequence: qpysequence,uploaded: bool)` is there?

fedonman: Nope, the thing is that I want to bypass the caching dictionary that `qblox_module` is using here. Every time I call `upload` I want the provided `qpysequence` to be uploaded without checking any cache. This is another argument that `compile` should be done outside the `qblox_module`. (since the original `upload` uses results from the `compile` method)

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -40,7 +40,7 @@
 from qililab.waveforms import IQPair, Waveform
 
 
-class BusInfo:  # pylint: disable=too-many-instance-attributes, too-few-public-methods
+class BusCompilationInfo:  # pylint: disable=too-many-instance-attributes, too-few-public-methods
     """Class representing the information stored by QBloxCompiler for a bus."""
```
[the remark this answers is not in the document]
fedonman: I will rename this class to private, it's not for public use.

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -53,6 +53,7 @@ def __init__(self):
         self.variable_to_register: dict[Variable, QPyProgram.Register] = {}
```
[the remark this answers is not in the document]
fedonman: All these are private, I will rename them.

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -167,7 +178,7 @@ def _append_to_waveforms_of_bus(self, bus: str, waveform_I: Waveform, waveform_Q
         """
 
         def handle_waveform(waveform: Waveform | None, default_length: int = 0):
-            _hash = QbloxCompiler._hash(waveform) if waveform else f"zeros {default_length}"
+            _hash = QbloxCompiler._hash_waveform(waveform) if waveform else f"zeros {default_length}"
```
visagim, line 182: At some discussions with the hardware team they pointed out that an envelope / pulse of 0 amplitude should not be treated as an empty pulse (because the electronics might still introduce noise).
I don't know if it would be too difficult to do like in the current qblox compiler and just add a wait when we want to send an empty envelope (I don't know in which context it would make sense to set the default of an empty pulse to amplitude otherwise)

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -247,33 +255,35 @@ def _handle_parallel(self, element: Parallel):
...
-                component=QPyInstructions.WaitSync(wait_time=QbloxCompiler.minimum_wait_duration),
-                bot_position=len(qpy_loop.components),
-            )
             self._buses[bus].qpy_block_stack[-1].append_component(qpy_loop)
             self._buses[bus].qpy_block_stack.append(qpy_loop)
             self._buses[bus].average_counter += 1
         return True
 
+    def _handle_infinite_loop(self, _: InfiniteLoop):
+        for bus in self._buses:
```
visagim, line 264: what is the purpose of infinite loops?

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -301,18 +311,13 @@ def _handle_reset_phase(self, element: ResetPhase):
...
-            else convert(element.gain_path1)
+        gain = (
+            self._buses[element.bus].variable_to_register[element.gain]
+            if isinstance(element.gain, Variable)
+            else convert(element.gain)
         )
         self._buses[element.bus].qpy_block_stack[-1].append_component(
-            component=QPyInstructions.SetAwgGain(gain_0=gain_0, gain_1=gain_1)
+            component=QPyInstructions.SetAwgGain(gain_0=gain, gain_1=gain)
         )
```
visagim, line 321: why are both gains the same?

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -328,118 +333,137 @@ def _handle_set_offset(self, element: SetOffset):
...
+        # If they are zero or one, return.
+        if len(buses) <= 1:
+            return
+
+        # If there is no bus marked for sync, return.
+        if all(not self._buses[bus].marked_for_sync for bus in buses):
+            return
+
+        # Is there any bus that has dynamic durations?
+        if any(bus for bus in buses if self._buses[bus].dynamic_durations or self._buses[bus].sync_durations):
```
visagim, line 366: Isn't the hanlde wait above adding dynamic durations if treated as a variable?
One might want to add dynamic waits for things like 2q gates (eg make the passive qubit wait while the other has a flux applied and iterate for several duration values)

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -328,118 +333,137 @@ def _handle_set_offset(self, element: SetOffset):
...
+        for bus in buses:
+            duration_diff = max_duration - self._buses[bus].static_duration
+            if duration_diff > 0:
+                self._buses[bus].qpy_block_stack[-1].append_component(
+                    component=QPyInstructions.Wait(wait_time=duration_diff)
+                )
+                self._buses[bus].static_duration += duration_diff
+
+    def __handle_dynamic_sync(self, buses: set[str]):
+        raise NotImplementedError("Dynamic syncing is not implemented yet.")
```
visagim, line 390: the recommendation from the QBLOX guys to handle times / sync between sequencers without running `wait_sync` (in case this helps) was to just copy the Q1ASM string from sequencers being used and substitute the `play` instructions by `wait`.
Maybe something similar could be done here to avoid keeping track of times all the time. For example setting a play instruction with a specific index that we know it's a dummy play instruction and parse that to a wait before uploading (though that feels very hacky, but maybe something along those lines)

src/qililab/instruments/qblox/qblox_module.py
visagim: Doesn't that make execution slower though if we have to re-upload every time even if the sequence is the same?

src/qililab/qprogram/qprogram.py
```diff
@@ -185,28 +220,50 @@ def play(self, bus: str, waveform: Waveform | IQPair):
...
+    @requires_domain("duration", Domain.Time)
+    def wait(self, bus: str, duration: int):
         """Adds a delay on the bus with a specified time.
 
         Args:
             bus (str): Unique identifier of the bus.
             time (int): Duration of the delay.
         """
-        operation = Wait(bus=bus, time=time)
+        operation = Wait(bus=bus, duration=duration)
```
visagim, line 231: do we handle long_waits?

visagim:
It's not very clear to me how we are handling syncs - maybe we could talk about it at some point?

src/qililab/qprogram/qblox_compiler.py
fedonman: Because on IQ buses the gain of both paths must be the same:

image

src/qililab/qprogram/qblox_compiler.py
fedonman: Yes, `wait` operation adds a dynamic wait if a variable is used.

src/qililab/qprogram/qblox_compiler.py
fedonman: I don't understand what the recommendation is.

This method had originally around ~120 lines of code to calculate the timings, but unfortunetely the hardware does not do some of the calculations correctly, so the code was replaced with an error for now.

The only other way to implement this is to "unwrap" the loops, as in c/p all operations within, with pre-calculated values.

src/qililab/qprogram/qprogram.py
fedonman: What is long_wait?

src/qililab/qprogram/qblox_compiler.py
fedonman: What do you mean by purpose? 😅

src/qililab/qprogram/qblox_compiler.py
fedonman: Q1ASM `play` instruction expects two waveforms always. If you don't need the second path (for example you are in a flux line bus) you have to set the output mappings accordingly, but this is not QProgram-related.

image

fedonman:
> It's not very clear to me how we are handling syncs - maybe we could talk about it at some point?

Sure, but what is unclear?

Everything that has a static real-time duration like `wait` and `play`, adds the static duration to the bus. If we use `sync` between buses with static durations, then the syncronization works correctly.

If any of the `wait` or `play` operations use a variable, so the durations change dynamically (that is, each loop iteration the value is different), then it's not possible to do syncronization. This is because of current HW limitations.

As I mentioned above, there is a workaround, to "unwrap" the loop, but the implementation is beyond this PR.

src/qililab/instruments/qblox/qblox_module.py
fedonman: But everytime you run a QProgram, you compile again and upload again, it's not like the compilation of software-loops with Qibo circuit, that you may have the same circuit between iterations.

src/qililab/qprogram/quantum_machines_compiler.py
```diff
@@ -0,0 +1,576 @@
...
+
+        # Return a dictionary with bus names as keys and the compiled Sequence as values.
+        return qua_program, self._configuration, result_handles
+
+    def _process_streams(self):
+        measurements = []
+
+        for i, measurement_compilation_info in enumerate(self._measurements):
+            measurement = MeasurementInfo()
+            if (stream_I := measurement_compilation_info.stream_I) is not None:
```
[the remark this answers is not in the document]
fedonman: We could refactor to a private method, but there are many variables involved, so I don't know if it is worth it for 4 lines.

src/qililab/qprogram/qprogram.py
fedonman: Ah! No, right now the maximum duration is 2^16 ns which is the maximum value of integers in Q1ASM. But you can always add two `wait` operations if you need more time.

src/qililab/instruments/qblox/qblox_module.py
visagim: the current cache for `qblox_module` (qblox compiler in the current workflow) we check whether the pulse schedule is the same or not (note that pulse schedule might not necessarily come from qibo circuits, it could be a custom pulse schedule)
at `qblox_module.compile`:
```python
        compiled_sequences = []
        sequencers = self.get_sequencers_from_chip_port_id(chip_port_id=pulse_bus_schedule.port)
        for sequencer in sequencers:
            if pulse_bus_schedule != self._cache.get(sequencer.identifier):
                sequence = self._compile(pulse_bus_schedule, sequencer)
                compiled_sequences.append(sequence)
            else:
                compiled_sequences.append(self.sequences[sequencer.identifier][0])
        return compiled_sequences
```
I don't know if a similar thing could be done here and I just realize this is done before even compiling so it might save some time to save compiled pulse schedules and check for repetitions at sequential executions of a program. 
But maybe that should be a separate task?

fedonman: But still the cache is about `PulseBusSchedule`, not `qpysequence`:

```
_cache: dict[int, PulseBusSchedule] = {}
```

And I am certain that this is the wrong place to have the cache, and the `compile` method too. The instrument class should just do `upload`, `run`, `get_acquisitions` and similar tasks. The orchestration of the whole flow should be in a much higher level like `platform`.

`QbloxModule` should not know that `PulseBusSchedule` exists, in the same sense that it does not know that a `QProgram` exists.

The weird syntax in `upload_qpysequence`: `self.sequences[sequencer.identifier] = (qpysequence, True)`  is because this line in `__init__` exists and I don't want to change it, since it will break the Qibo flow.

```
# The sequences dictionary contains all the compiled sequences for each sequencer and a flag indicating whether
# the sequence has been uploaded or not
self.sequences: dict[int, tuple[QpySequence, bool]] = {}  # {sequencer_idx: (program, True), ...}
```

fedonman: `self.sequences: dict[int, tuple[QpySequence, bool]] = {}`  should become `self.sequences: dict[int, QpySequence] = {}` in the alternative implementation that I am describing.

fedonman: We can create a separate task to investigate this and discuss with the team.

src/qililab/qprogram/qblox_compiler.py
visagim: I assume it's so we can do something like qblox's infinite loops
```
seq_prog = """
      wait_sync 4

loop: play    0,0,1200
      jmp     @loop
"""
```
but if it's like that, is killing the process the only way to stop it?

src/qililab/qprogram/qblox_compiler.py
visagim: There are experiments which use duration as a variable. Off the top of my head I can think of the [old cz_conditional experiment](https://github.com/qilimanjaro-tech/HardwareDataTools/blob/bsc-working/src/hwdatatools/experiment_portfolio/cz_conditional.py) but also our current [ssro measurement calibration ](https://github.com/qilimanjaro-tech/LabScripts/blob/main/Personal%20folders/Qililab/ssro_duration.ipynb), they both iterate over duration values for some gate / pulse.
I guess it's not too bad since we could still do this like now through software loops but it'd be very nice if we could do this through hardware loops.
Related to all my other comments about waits / syncs / durations below, I understand it is better to do this in the future / in another pr. So my only concern / doubt (from pure naivety) is if it's gonna be easier to add support for this on top of the current code, or if the current code is such that we're gonna have to modify essential parts of it in order to support duration loops in the future (what do you think?)

src/qililab/qprogram/qprogram.py
visagim: for instance if you execute a circuit and you want to wait for the relaxation time of the qubit before the next iteration, or if you have to wait for the remaining time of `repetition_duration` (which is typically 200_000) this will be more than the max time allowed by the wait instruction (from `qpysequence` because I cannot find this info on the qblox docs `INST_MAX_WAIT = 2**16 - 4`)

fedonman: Yes I agree. But `long_wait` in qpysequence's library is outdated. It forces the duration to be a multiple of 4 which is changed in the version of Qblox we use. Now, the durations can be a multiple of 1.

I will mark this as a future task, to support `long_wait` with one instruction, but for now there is the work around of adding multiple `wait` operations.

2^16ns however is plenty for relaxation times, anyway, right?

src/qililab/qprogram/qblox_compiler.py
fedonman: Yes I know. There is already a task for future PR to support this. No, the changes wouldn't be great, this is why there are separated into different methods since now.

src/qililab/qprogram/qblox_compiler.py
fedonman: Yes for Qblox, this is the only way. Either manually or through some timeout mechanism. Quantum Machines support infinite loops natively though.

requirements.txt
```diff
@@ -4,11 +4,13 @@ PyYAML==6.0
 qblox-instruments==0.10.1
 qcodes==0.38.1
 qcodes_contrib_drivers==0.18.0
+qm_qua==1.1.4
```
AlbertMitjans, line 7: Duplicated line!

src/qililab/instruments/qblox/qblox_module.py
```diff
@@ -551,6 +551,20 @@ def reset(self):
...
+        Args:
+            qpysequence (QpySequence): The qpysequence to upload.
+            port (str): The port of the sequencer to upload to.
+        """
+        sequencers = self.get_sequencers_from_chip_port_id(chip_port_id=port)
+        for sequencer in sequencers:
+            logger.info("Sequence program: \n %s", repr(qpysequence._program))  # pylint: disable=protected-access
+            self.device.sequencers[sequencer.identifier].sequence(qpysequence.todict())
+            self.device.sequencers[sequencer.identifier].sync_en(True)
+            self.sequences[sequencer.identifier] = (qpysequence, True)
```
AlbertMitjans, line 566: Do we really need this line?

src/qililab/instruments/qblox/qblox_qrm.py
```diff
@@ -298,6 +309,39 @@ def get_acquisitions(self) -> QbloxResult:
...
+                self.device.get_acquisition_state(
+                    sequencer=sequencer_id, timeout=cast(AWGQbloxADCSequencer, sequencer).acquisition_timeout
+                )
+
+                if sequencer.scope_store_enabled:
+                    for acquisition in acquisitions:
+                        self.device.store_scope_acquisition(sequencer=sequencer_id, name=acquisition)
+
+                for acquisition in acquisitions:
+                    results.append(
```
AlbertMitjans, line 339: This is not covered 😨

src/qililab/instruments/qblox/qblox_module.py
AlbertMitjans: Ah I see you use it for the acquisitions method

AlbertMitjans: Shouldn't we then clean this variable after getting the acquisitions? If you upload two QPrograms (in a row) with different sequencers, in the second execution you would get the results of all sequencers!

src/qililab/platform/platform.py
```diff
@@ -526,6 +528,40 @@ def __str__(self) -> str:
...
+        sequences = qblox_compiler.compile(qprogram=qprogram)
+        buses = {bus_alias: self._get_bus_by_alias(alias=bus_alias) for bus_alias in sequences}
+
+        # Upload sequences
+        for bus_alias in sequences:
+            buses[bus_alias].upload_qpysequence(qpysequence=sequences[bus_alias])
+
+        # Execute sequences
+        for bus_alias in sequences:
+            buses[bus_alias].run()
```
AlbertMitjans, line 550: Do we need to have 3 separated methods: upload, run, acquire?

Why not having one execute, and doing everything inside the driver?

src/qililab/instruments/qblox/qblox_module.py
fedonman: Yes, you are right! I will do that.

fedonman:
A comment on results. As we discussed, I created the `QbloxQProgramMeasurementResult`. I know big name, but it will get shorter when we unify results for both circuits and qprograms. 

As the name implies, this is the result of a **single** measurement. An execution of QProgram might return multiple results from multiple buses. For now, these are stored in a dictionary of type `dict[str, list[Result]`, with the keys being the names of the buses and the list the measurements in chronological order.

Later, we should add a wrapper class e.g. `QProgramResult` that has this dictionary as attribute and may offer additional processing through methods.

Also, for now the measurement result just store the raw_data and returns the I/Q measurements with the `arrray` method. More methods would be added when we decide on an interface. This should be done ideally after creating the result class for Quantum Machines.

src/qililab/constants.py
```diff
@@ -214,6 +214,12 @@ class QBLOXRESULT:
     QBLOX_RAW_RESULTS = "qblox_raw_results"
 
 
+class QBLOXMEASUREMENTRESULT:
```
AlbertMitjans, line 217: I hate using these constants... :(

src/qililab/platform/platform.py
```diff
@@ -528,15 +529,17 @@ def __str__(self) -> str:
         """
         return str(YAML().dump(self.to_dict(), io.BytesIO()))
 
-    def execute_qprogram(self, qprogram: QProgram):
+    def execute_qprogram(self, qprogram: QProgram) -> dict[str, list[Result]]:
```
AlbertMitjans, line 532: Before the official release of `QProgram` (which should be December 1st) can we have a `Results` class instead of a dictionary?

AlbertMitjans: When doing measurements on different qubits this class will be very helpful to get the probabilities of all the qubits with just one call

fedonman: Yes we can! I didn't want to add it in this PR because it is already huge.

src/qililab/constants.py
AlbertMitjans: I use strings directly in my PRs 🤭

src/qililab/platform/platform.py
AlbertMitjans: Completely agree!

src/qililab/instruments/qblox/qblox_qrm.py
```diff
@@ -187,6 +187,33 @@ def acquire_result(self) -> QbloxResult:
...
+            list[QbloxQProgramMeasurementResult]: Acquired Qblox results in chronological order.
+        """
+        return self._get_qprogram_acquisitions(acquisitions=acquisitions)
+
+    @Instrument.CheckDeviceInitialized
+    def _get_qprogram_acquisitions(self, acquisitions: list[str]) -> list[QbloxQProgramMeasurementResult]:
+        results = []
+        for acquisition in acquisitions:
+            for sequencer in self.awg_sequencers:
+                if sequencer.identifier in self.sequences:
```
AlbertMitjans, line 206: Again, we are not cleaning the cache in `self.sequences`! I don't know if this will bring us problems...

Did you try uploading + executing + getting results of >1 QPrograms in a row? I have the feeling that in the last QPrograms we will also get the results of the first ones...

src/qililab/instruments/qblox/qblox_qrm.py
```diff
@@ -187,6 +187,33 @@ def acquire_result(self) -> QbloxResult:
         """
         return self.get_acquisitions()
 
+    def acquire_qprogram_results(self, acquisitions: list[str]) -> list[QbloxQProgramMeasurementResult]:  # type: ignore
```
AlbertMitjans, line 190: Another worry I have: beware that for the old workflow we delete the acquisition data of old programs. We need to do this, otherwise the old data is still stored in each bin, and will be averaged with the new data!

src/qililab/instruments/qblox/qblox_qrm.py
fedonman: I clean it after acquiring results in the `platform.execute_qprogram` method:

```
# Reset instrument settings
           instrument.reset_sequences()
```

src/qililab/instruments/qblox/qblox_qrm.py
AlbertMitjans: With Victor we started thinking about what are all the steps (for Qblox) that we need to do everytime we execute a new program/circuit. This would be a good time to write them down!

fedonman: > Another worry I have: beware that for the old workflow we delete the acquisition data of old programs. We need to do this, otherwise the old data is still stored in each bin, and will be averaged with the new data!

Can you point me to the line that executes this in the old workflow?

AlbertMitjans: Here: https://github.com/qilimanjaro-tech/qililab/blob/83b6ea22f91b10da945a61f99ba56132c585c709/src/qililab/instruments/qblox/qblox_qrm.py#L178

src/qililab/instruments/qblox/qblox_qrm.py
AlbertMitjans: Nice! Then we could clean the acquisition data in here aswell (maybe clean it when calling `reset_sequences()`?

src/qililab/instruments/qblox/qblox_qrm.py
AlbertMitjans: Wait, maybe by re-uploading a program the acquisition data is removed...

fedonman: Yes, you are right! Missed that! I added this to the method and tests.

src/qililab/instruments/qblox/qblox_qrm.py
fedonman: I added the acquisition data deletion to the `qblox_qrm` since it is relevant only for readout instruments. This loop runs for both control and readouts.

src/qililab/instruments/qblox/qblox_qrm.py
AlbertMitjans: Did you test it in HW? Because now I think we don't need it 😆 

If you check the line above, the deletion is done only when we don't re-upload the sequence.

Given that in the QProgram case we always upload a program, maybe we don't even need this...

fedonman: I cannot test in HW again until Monday!

The line above though, is run if a sequence was previously uploaded, not if we re-upload one. So it runs every time after uploading a sequence.

fedonman: I believe your original thought on needing this is correct.

AlbertMitjans: I'm not so sure... 🤔 We can discuss this after the daily if you want!

visagim: also it's probably safer to do
`self.device.delete_acquisition_data(sequencer=sequencer.identifier, all=True)` which clears all acquisition data in the sequencer. Note that deleting acquisitions only sets all the acquisition values to `None` but the sequencer still has the acquisition data entries (which in this case will be something like `{"default: [None, None, ...]}`)

PR #564, opened 2023-10-12
pulse_documentation_derived_error_and_improvements

GuillermoAbadLopez:
Improved the error about Derived appearing as summary of the class, because it was the first line!

And also added images for `pulse_shapes`!


jjmartinezQT (approved):
Great improvements!

PR #565, opened 2023-10-13
Bug serializing QRM/QCM-RF instruments and precision errors

IsaacLA12:
There was a bug serializing the instruments `QbloxQRMRF` and `QbloxQCMRF`. An attribute `out_offsets` was dumped into the saved runcard file for those instruments when those do not have that particular attribute. Encountered this bug testing a corner case where you `build_platform` then `save_platform` and `build_platform` again. A new test was added on this PR to cover this corner case.

This PR solves a precision error with decimals, now the `save_platform` is correctly saving the platform on the runcard maintaining the same precision. Covered in the previously mentioned test.


GuillermoAbadLopez (approved):
Review my changes before merging, and tell me if something seems wrong!

My changes were basically:
- In the QCM/QRM RF test, I've given them the attribute "out_offsets" to the classes, to check that the to_dict actually pops it!
- And also I did the test on the platform serialization, longer with 2 serialization cycles and with two different origins (Galadriel.runcard and the yaml file)!
- FInally removed from the file the dm_build_platform() method you added, since it really wasn't necessary. Check that in the class you are their were not being mocked! (There are two test classes in that file!)

But if nothing, PR approved!

IsaacLA12:
Thanks @GuillermoAbadLopez looks good for me too!

PR #567, opened 2023-10-18
Platform docs / solving errors of list sizes

GuillermoAbadLopez:
CONTEXT:
The sizes of the for loops before are 11, so the results, even though they are still invented, should be 11 as well for consistency!


jjmartinezQT (approved):
Good to go.

PR #573, opened 2023-10-21
docs_improvements_for_Waveforms&PulseShapes

GuillermoAbadLopez:
Improvement on `Waveforms` documentation for the selection process.

During which I did some improvement to the pulse_shape documentation as well.

### DIFF
--- a/src/qililab/pulse/pulse_shape/gaussian.py
+++ b/src/qililab/pulse/pulse_shape/gaussian.py
@@ -27,17 +27,28 @@ from qililab.utils import Factory
 @Factory.register
 @dataclass(frozen=True, eq=True)
 class Gaussian(PulseShape):
-    """Standard centered Gaussian pulse shape:
+    """Standard centered Gaussian pulse shape, symmetrically spanning for ``num_sigmas`` over the pulse duration.
+
+    The normal distribution's parameters :math:`\mu` (mean) and :math:`\sigma` (standard deviation) will be therefore
+    defined by :math:`\mu =` ``duration`` :math:`/ 2` and :math:`\sigma =` ``duration`` :math:`/` ``num_sigmas``:
 
     .. math::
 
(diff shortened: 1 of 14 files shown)


src/qililab/pulse/pulse_shape/gaussian.py
```diff
@@ -53,22 +64,23 @@ class Gaussian(PulseShape):
             :align: center
 
     Args:
-        num_sigmas (float): Sigma number of the gaussian pulse shape. Defines the height of the gaussian pulse.
+        num_sigmas (float): Sigma number of the gaussian pulse shape. Defines the width of the gaussian pulse.
```
jjmartinezQT, line 67: good catch

src/qililab/waveforms/iq_pair.py
```diff
@@ -12,12 +12,15 @@
...
 
+"""IQPair dataclass."""
 from dataclasses import dataclass
 
 from qililab.waveforms.waveform import Waveform
 
 
 @dataclass
 class IQPair:  # pylint: disable=missing-class-docstring
+    """IQPair dataclass, containing the real (I) and imaginary (Q) parts of a signal."""
```
jjmartinezQT, line 23: I am not really confortable with the 'imaginary' and 'real' part descriptions for I and Q after reading about it, I would rather stick to the 'in-phase' and 'quadrature' descriptions, respectively. What do you think?

jjmartinezQT (changes requested):

src/qililab/waveforms/iq_pair.py
GuillermoAbadLopez: Sure yes! It's not really my domain, so I'll go with what you think its best!

GuillermoAbadLopez: Changed!

jjmartinezQT (approved):

PR #575, opened 2023-10-23
TAS-411/Unify native gate and circuit to pulses transpilers

visagim:
- [ ] Fix tests
- [ ] Fix documentation


src/qililab/circuit_transpiler/__init__.py
```diff
@@ -37,6 +37,5 @@
 """
 
 from .gate_decompositions import translate_gates
-from .native_gates import Drag
-from .park_gate import Park
-from .transpiler import translate_circuit
+from .native_gates import Drag, Wait
```
AlbertMitjans, line 40: Why is not `Park` in the `__init__`?

src/qililab/execution/execution_builder.py
```diff
@@ -98,5 +98,5 @@ def _get_bus_info_from_loop_alias(self, platform: Platform, loop: Loop):
             bus: Bus object
         """
         alias = loop.alias
-        bus = platform._get_bus_by_alias(alias=alias)  # pylint: disable=protected-access
+        bus = platform.get_bus_by_alias(alias=alias)  # pylint: disable=protected-access
```
AlbertMitjans, line 101: If I remember correctly, @fedonman made this private in a recent PR.

Even though this is used outside of the class, I like hiding it from the user, given that they should not use it.

src/qililab/experiment/portfolio/t1.py
AlbertMitjans, line 1: We should get rid of the experiment module soon 😆

src/qililab/transpiler/park_gate.py
AlbertMitjans, line 1: Nice!

src/qililab/circuit_transpiler/transpiler.py
AlbertMitjans, line 1: Can we change the name to `circuit_transpiler` to make sure it has the same name as the class?

tests/circuit_transpiler/test_circuit_to_pulses.py
AlbertMitjans, line 1: Remember that we should have a 1-1 correspondence between test files and source code files.

I think all these tests should go within `test_transpiler.py` right? 🤔

AlbertMitjans (commented):
Here are some comments, but looks good so far!!

AlbertMitjans:
> [Improve UX of transpilation](https://www.notion.so/Improve-UX-of-transpilation-d8a262467b304745a82b92f908c902d9?pvs=8&n=github_linkback)

Careful! You linked this PR to a wrong task!

visagim:
> > [Improve UX of transpilation](https://www.notion.so/Improve-UX-of-transpilation-d8a262467b304745a82b92f908c902d9?pvs=8&n=github_linkback)
> 
> Careful! You linked this PR to a wrong task!

Improve [UX transpilation](https://www.notion.so/qilimanjaro/Improve-UX-of-transpilation-d8a262467b304745a82b92f908c902d9) should be the right one. I don't know why it keeps getting linked to [this one instead](https://www.notion.so/qilimanjaro/FUTURE-IDEAS-low-priority-11e0b67b4fef4373ba79ccbb54d922d3)

tests/circuit_transpiler/test_circuit_to_pulses.py
visagim: yes, fixing unit tests is still pending for after we all agree on the new names / organization

AlbertMitjans: I agree with the nomeclature! You have the OK from my side 👍 

But please discuss it with the team in today's daily

visagim:
Careful, there's already a `circuit/circuit_transpiler.py` and this might be confusing!

AlbertMitjans:
We should remove the `circuit` module, given that it will never be used. Sorry @fedonman 😅

fedonman:
> We should remove the `circuit` module, given that it will never be used. Sorry @fedonman 😅 

Never say never ☺️ I am sure we will need it in the future. 

Anyway, if we want to remove it from main, please move the code to a branch so it's not lost 🙏

fedonman:
I can do it if needed.

visagim:
> > We should remove the `circuit` module, given that it will never be used. Sorry @fedonman 😅
> 
> Never say never ☺️ I am sure we will need it in the future.
> 
> Anyway, if we want to remove it from main, please move the code to a branch so it's not lost 🙏

https://github.com/qilimanjaro-tech/qililab/tree/qililab-circuit

AlbertMitjans:
* A link back to this PR.

visagim:
> [FUTURE IDEAS, low priority](https://www.notion.so/FUTURE-IDEAS-low-priority-11e0b67b4fef4373ba79ccbb54d922d3?pvs=8&n=github_linkback)

I'm not even touching anything and this keeps getting linked o.0

src/qililab/circuit_transpiler/circuit_transpiler.py
```diff
@@ -12,72 +12,168 @@
...
-    """Translates a list of circuits into a list of pulse sequences (each circuit to an independent pulse sequence)
-    For each circuit gate we look up for its corresponding gates settings in the runcard (the name of the class of the circuit
-    gate and the name of the gate in the runcard should match) and load its schedule of GateEvents.
-    Each gate event corresponds to a concrete pulse applied at a certain time w.r.t the gate's start time and through a specific bus
-    (see gates settings docstrings for more details).
 
-    Measurement gates are handled in a slightly different manner. For a circuit gate M(0,1,2) the settings for each M(0), M(1), M(2)
-    will be looked up and will be applied in sync. Note that thus a circuit gate for M(0,1,2) is different from the circuit sequence
-    M(0)M(1)M(2) since the later will not be necessarily applied at the same time for all the qubits involved.
+class CircuitTranspiler:
```
fedonman, line 44: I would still rename this to `QiboCircuitTranspiler` since it is a library-specific transpiler and to avoid future breaking issues if we add our own circuit class.

visagim: I agree if indeed at some time we introduce a very different circuit structure. However for circuit representations like qasm it would be pretty easy to adapt the current `CircuitTranspiler` for that as well / instead of qibo. Since we don't know yet what we'll do I'm not too sure about the name change but we can comment this with the rest of the team and see what they think.

visagim:
Will handle the (unrelated to this pr) uncovered code in a separate pr (since it is unrelated)

visagim:
Will handle the (unrelated to this pr) uncovered code in a separate pr (since it is unrelated)

fedonman:
> > [FUTURE IDEAS, low priority](https://www.notion.so/FUTURE-IDEAS-low-priority-11e0b67b4fef4373ba79ccbb54d922d3?pvs=8&n=github_linkback)
> 
> I'm not even touching anything and this keeps getting linked o.0

Correct ID is TAS-441, not TAS-411.

visagim:
> > > [FUTURE IDEAS, low priority](https://www.notion.so/FUTURE-IDEAS-low-priority-11e0b67b4fef4373ba79ccbb54d922d3?pvs=8&n=github_linkback)
> > 
> > 
> > I'm not even touching anything and this keeps getting linked o.0
> 
> Correct ID is TAS-441, not TAS-411.

ooooh well observed!

src/qililab/__init__.py
```diff
@@ -16,14 +16,14 @@
 This is the top level module from which all basic functions and classes of
 Qililab can be directly imported.
 """
+from .circuit_transpiler import CircuitTranspiler, Drag, Wait
```
AlbertMitjans, line 19: Do we need `CircuitTranspiler` here? Are users supposed to use this class directly? They should use platform instead, right?

src/qililab/circuit_transpiler/circuit_transpiler.py
```diff
@@ -12,39 +12,42 @@
...
+from qililab.circuit_transpiler.gate_decompositions import translate_gates
 from qililab.constants import RUNCARD
-from qililab.instruments import AWG
 from qililab.settings.gate_event_settings import GateEventSettings
 from qililab.typings.enums import Line
 from qililab.utils import Factory
 
 from ..pulse.pulse import Pulse
 from ..pulse.pulse_event import PulseEvent
 from ..pulse.pulse_schedule import PulseSchedule
```
AlbertMitjans, line 36: When importing things outside of the current submodule (which in this case is `circuit_transpiler`), we should import it from qililab (just to follow the same notation everywhere).

Also, remember that each submodule should contain in its `__init__` all the classes/functions that are used outside of the submodule, such that other files outside that submodule can import all the needed functionality in just one line. In this example:
```suggestion
from qililab.pulse import Pulse, PulseEvent, PulseSchedule
```

AlbertMitjans: My second comment above doesn't have to be implemented now in this PR, but I mention it such that everyone in the team has this mentality.

If we follow this, the code will become MUCH clearer and we will stop having circular imports.

AlbertMitjans: I will put it in other words: let's try to avoid having two dots in any of the imports, meaning:
```python
from qililab.x.y import z
```
should become:
```python
from qililab.x import z
```

src/qililab/circuit_transpiler/circuit_transpiler.py
```diff
@@ -12,39 +12,42 @@
...
+from .native_gates import Drag, Wait
+
+if TYPE_CHECKING:
```
AlbertMitjans, line 39: I don't like this at all. I prefer not having type checks for this specific case than creating this variable

AlbertMitjans (commented):
Good job!! Added some comments 😄

fedonman:
I don't know if this suggestion should be included in this PR, but since it touches compilation flow, I will add it here, so it can be discussed.

This originated from Victor's [comment](https://github.com/qilimanjaro-tech/qililab/pull/563#discussion_r1383423073) in my PR.

The argument in a nutshell is that instruments and instrument controllers **should not** do compilation stuff. 

For example, a `QbloxModule` should not know that a `PulseBusSchedule`, `Qibo circuit` or `QProgram` exists. They should contain only methods with parameters that are directly related to them. That would be a `qpysequence` for Qblox and QUA for QM.

This can be seen clearly if you follow the flow that happens now. Starting at the `compile` method of `platform` we can see that half the compilation (`Circuit` -> `PulseBusSchedule`) is happening here using the `CircuitTranspiler` and the other half (`PulseBusSchedule` -> `qpysequence`) is done by calling the `compile` method of `Bus` -> `SystemControl` -> `Instrument` which makes no sense.

```
def compile(self, program: PulseSchedule | Circuit, num_avg: int, repetition_duration: int, num_bins: int) -> dict:
        if isinstance(program, Circuit):
            transpiler = CircuitTranspiler(platform=self)
            pulse_schedule = transpiler.transpile_circuit(circuits=[program])[0]
        else:
            pulse_schedule = program

        programs = {}
            bus = self.buses.get(port=pulse_bus_schedule.port)
            bus_programs = bus.compile(pulse_bus_schedule, num_avg, repetition_duration, num_bins)
            programs[bus.alias] = bus_programs

        return programs
```

So, the logic contained at the last level, that is in the following method of `QbloxModule`, should instead be moved to the `compile` method of `platform`. (using a different class like `PulseBusScheduleTranspiler` or as a method in `CircuitTranspiler`.

```
def compile(
        self, pulse_bus_schedule: PulseBusSchedule, nshots: int, repetition_duration: int, num_bins: int
    ) -> list[QpySequence]:
        if nshots != self.nshots or repetition_duration != self.repetition_duration or num_bins != self.num_bins:
            self.nshots = nshots
            self.repetition_duration = repetition_duration
            self.num_bins = num_bins
            self.clear_cache()

        compiled_sequences = []
        sequencers = self.get_sequencers_from_chip_port_id(chip_port_id=pulse_bus_schedule.port)
        for sequencer in sequencers:
            if pulse_bus_schedule != self._cache.get(sequencer.identifier):
                sequence = self._compile(pulse_bus_schedule, sequencer)
                compiled_sequences.append(sequence)
            else:
                compiled_sequences.append(self.sequences[sequencer.identifier][0])
        return compiled_sequences
   ```

**Comment by visagim** [comment]
> I don't know if this suggestion should be included in this PR, but since it touches the compilation flow, I will add it anyway, so it can be discussed.
> 
> This originated from Victor's [comment](https://github.com/qilimanjaro-tech/qililab/pull/563#discussion_r1383423073) in my PR.
> 
> The argument in a nutshell is that instruments and instrument controllers **should not** do compilation stuff.
> 
> For example, a `QbloxModule` should not know that a `PulseBusSchedule`, `Qibo circuit` or `QProgram` exists. They should contain only methods with parameters that are directly related to them. That would be a `qpysequence` for Qblox and QUA for QM.
> 
> This can be seen clearly if you follow the flow that happens now. Starting at the `compile` method of `platform` we can see that half the compilation (`Circuit` -> `PulseBusSchedule`) is happening here using the `CircuitTranspiler` and the other half (`PulseBusSchedule` -> `qpysequence`) is done by calling the `compile` method of `Bus` -> `SystemControl` -> `Instrument` which makes no sense.
> 
> ```
> def compile(self, program: PulseSchedule | Circuit, num_avg: int, repetition_duration: int, num_bins: int) -> dict:
>         if isinstance(program, Circuit):
>             transpiler = CircuitTranspiler(platform=self)
>             pulse_schedule = transpiler.transpile_circuit(circuits=[program])[0]
>         else:
>             pulse_schedule = program
> 
>         programs = {}
>         for pulse_bus_schedule in pulse_schedule.elements:
>             bus = self.buses.get(port=pulse_bus_schedule.port)
>             bus_programs = bus.compile(pulse_bus_schedule, num_avg, repetition_duration, num_bins)
>             programs[bus.alias] = bus_programs
> 
>         return programs
> ```
> 
> So, the logic contained at the last level, that is in the following method of `QbloxModule`, should instead be moved to the `compile` method of `platform`. This can be done by using a different class like `PulseBusScheduleTranspiler` or as a method in `CircuitTranspiler`.
> 
> ```
> def compile(
>         self, pulse_bus_schedule: PulseBusSchedule, nshots: int, repetition_duration: int, num_bins: int
>     ) -> list[QpySequence]:
>         if nshots != self.nshots or repetition_duration != self.repetition_duration or num_bins != self.num_bins:
>             self.nshots = nshots
>             self.repetition_duration = repetition_duration
>             self.num_bins = num_bins
>             self.clear_cache()
> 
>         compiled_sequences = []
>         sequencers = self.get_sequencers_from_chip_port_id(chip_port_id=pulse_bus_schedule.port)
>         for sequencer in sequencers:
>             if pulse_bus_schedule != self._cache.get(sequencer.identifier):
>                 sequence = self._compile(pulse_bus_schedule, sequencer)
>                 compiled_sequences.append(sequence)
>             else:
>                 compiled_sequences.append(self.sequences[sequencer.identifier][0])
>         return compiled_sequences
> ```
> 
> For example, the final `platform.compile` method could be:
> 
> ```
> def compile(self, program: PulseSchedule | Circuit, num_avg: int, repetition_duration: int, num_bins: int) -> dict:
>         transpiler = CircuitTranspiler(platform=self)
>         pulse_schedule = transpiler.transpile_circuit(circuits=[program])[0] if isinstance(program, Circuit) else program
>         programs = transpiler.traspile_pulse_schedule(pulse_schedule=pulse_schedule)
>         return programs
> ```

Indeed I think this is out of the scope of this PR, but I 100% agree with it. 

> So, the logic contained at the last level, that is in the following method of QbloxModule, should instead be moved to the compile method of platform. This can be done by using a different class like PulseBusScheduleTranspiler or as a method in CircuitTranspiler.
This would fix a couple of issues we've already encountered a couple of times related to the current qblox compiler (handled at each bus instead of at the platform level) not having awareness of the other buses (since other buses are not accessible at the bus level). For more advanced algorithms like active reset, combined measurements, pulse optimization we really need this awareness in the compiler and it would make more sense to do as @fedonman says and compile qpysequence / qua / etc at a platform level and then send this sequence to each bus (it doesn't need to be at `platform.py` per se, could be some new class like `Compiler` (though this might give similar type import problems as the ones we have in the transpiler) at the same execution level as `Platform`)

File: src/qililab/circuit_transpiler/circuit_transpiler.py [hunk 7]
```diff
@@ -12,72 +12,161 @@
...
+        if optimize:
+            gates_to_optimize = translate_gates(circuit.queue)
+            new_circuit.add(self.optimize_transpilation(circuit.nqubits, ngates=gates_to_optimize))
+        else:
+            new_circuit.add(translate_gates(circuit.queue))
+
+        return new_circuit
+
+    def optimize_transpilation(self, nqubits: int, ngates: list[gates.Gate]) -> list[gates.Gate]:
+        """Optimizes transpiled circuit by applying virtual Z gates.
```
jjmartinezQT, line 80: Nice explanation.

src/qililab/circuit_transpiler/circuit_transpiler.py
```diff
@@ -12,72 +12,161 @@
...
+                )
+                if corrections is not None:
+                    shift[control_qubit] += corrections[f"q{control_qubit}_phase_correction"]
+                    shift[target_qubit] += corrections[f"q{target_qubit}_phase_correction"]
+                new_gates.append(gate)
+            else:
+                # if gate is drag pulse, shift parameters by accumulated Zs
+                if isinstance(gate, Drag):
+                    # create new drag pulse rather than modify parameters of the old one
+                    gate = Drag(gate.qubits[0], gate.parameters[0], gate.parameters[1] - shift[gate.qubits[0]])
```
jjmartinezQT, line 134: What is the difference between a new Drag or modifying the old one? Are we using the old one later one unchanged?

src/qililab/circuit_transpiler/circuit_transpiler.py
```diff
@@ -107,7 +196,7 @@ def translate(  # pylint: disable=too-many-locals, too-many-branches
                 # apply gate schedule
                 for gate_event in gate_schedule:
                     # find bus
-                    bus = self.platform._get_bus_by_alias(gate_event.bus)  # pylint: disable=protected-access
+                    bus = self.platform.get_bus_by_alias(gate_event.bus)  # pylint: disable=protected-access
```
jjmartinezQT, line 199: If it is not longer a protected method, do we still need the pylint ignore command here?

src/qililab/execution/execution_builder.py
```diff
@@ -98,5 +96,5 @@ def _get_bus_info_from_loop_alias(self, platform: Platform, loop: Loop):
...
```
jjmartinezQT, line 99: same here about the pylint comment.

jjmartinezQT (approved):
Have a look at the pylint but that would be mostly it, and open question for clarification. Good work.

src/qililab/circuit_transpiler/circuit_transpiler.py
visagim: This is mostly a safeguard. Every single qibo gate is a separate object with a single pointer to it. This is useful if you want to do things like setting parameters efficiently, but it also means that you can do the following:
```python
from qibo.gates import RX
from qibo.models import Circuit

c = Circuit(1)
c.add(RX(0,1))

c2 = Circuit(1)
rx = c.queue[0]
rx.parameters = (2,)
c2.add(rx)
print([gate.parameters for gate in c2.queue])
rx.parameters = (1,)
print([gate.parameters for gate in c2.queue])

out: [(2,)]
out: [(1,)]
```
So if I just add the original gate with the parameters changed, if someone then changes the parameters of the original circuit then the parameters of the transpiled circuit would also change.
There's also a `copy` method for qibo gates, but it's had bugs in the past so the best way I found of copying a gate with different parameters was to actually create a new gate with the desired parameters

jjmartinezQT (approved):
Good to go, good work.

PR #578, opened 2023-10-24
[BUG] Error cutting decimals in yaml.dump > ruamel.yaml.YAML().dump()

GuillermoAbadLopez:
## PR using `ruamel.yaml.YAML().dump()` as standard yaml dumping in qililab to not lose decimals, and work with scientific notation (1e-9)!

Since we want to use scientific notation and large decimals, I've added a test to check the serialization works with:
- large decimals
- scientific notation 1e-9
- underscore notaiton 100_000_000
- None's

## Problems/Warnings I see:
- `ruamel.yaml.YAML().dump` needs a stream almost always, so I had to pass `io.BytesIO()` in various places where before `io` was not being used.

- I see Albert had a yaml_representer for **Null values**, aswell, I dunno if those are supported by `ruamel.yaml`, but I have searched and they are not used in the runcards I have, only None's are being used, and those work!

- Its a longer and uglier code, maybe we could define a  `QililabYAML = ruamel.yaml.YAML()` ?? Not necessary though

- If something was previously in scientific notation or in under score notation, they get translated to normal float for the next Runcard. So Runcard don't keep exactly the format they are written with, but everything works together. Also I guess this already happened before, so not really a problem.

## Other problem we are gonna solve in this PR:
- Fixing the version of `ruamel` in the requirements.txt so main doesn't fail in the future.


src/qililab/experiment/base_experiment.py
```diff
@@ -21,6 +21,7 @@
 from threading import Thread
 
 import numpy as np
+import ruamel.yaml
```
AlbertMitjans, line 24: Instead of using this import and then `ruamel.yaml.YAML` we could import the YAML class directly:
```suggestion
from ruamel.yaml import YAML
```

AlbertMitjans (commented):
Looking at the `ruamel` documentation, I don't think we need a context manager anymore 🤔

src/qililab/experiment/base_experiment.py
GuillermoAbadLopez: Indeed! It will look way better that way yes!

GuillermoAbadLopez:
Yeah, I'll leave that to you, since I am not very familiar with it! I guess you are talking about how we can pass the io.ByTestIO() instead than using the context manager?

Another thing, from what I've seen `ruamel` has a very broad acceptation as an improvement for the basic PyYAML!

src/qililab/experiment/base_experiment.py
GuillermoAbadLopez: Quickly applied this change!

src/qililab/experiment/base_experiment.py
```diff
@@ -127,7 +127,7 @@ def _threaded_function():
                 if self.results_path is not None:
                     with open(file=self.results_path / "results.yml", mode="a", encoding="utf8") as data_file:
                         result_dict = result.to_dict()
-                        yaml.safe_dump(data=[result_dict], stream=data_file, sort_keys=False)
+                        YAML().dump(data=[result_dict], stream=data_file)
```
AlbertMitjans, line 130: Can we use `sort_keys=False` with ruamel to make sure the functionality is the same?

src/qililab/experiment/base_experiment.py
```diff
@@ -357,7 +357,7 @@ def prepare_results(self, save_experiment=True, save_results=True) -> tuple[Resu
             if save_experiment:
                 # Dump the experiment data into the created file
                 with open(file=results_path / EXPERIMENT_FILENAME, mode="w", encoding="utf-8") as experiment_file:
-                    yaml.dump(data=self.to_dict(), stream=experiment_file, sort_keys=False)
+                    YAML().dump(data=self.to_dict(), stream=experiment_file)
```
AlbertMitjans, line 360: Same here

src/qililab/experiment/base_experiment.py
```diff
@@ -414,4 +414,4 @@ def _create_results_file(self, path: Path):
             EXPERIMENT.RESULTS: None,
         }
         with open(file=path / RESULTS_FILENAME, mode="w", encoding="utf-8") as results_file:
-            yaml.dump(data=data, stream=results_file, sort_keys=False)
+            YAML().dump(data=data, stream=results_file)
```
AlbertMitjans, line 417: Same here

src/qililab/instrument_controllers/instrument_controllers.py
```diff
@@ -64,4 +65,4 @@ def __str__(self) -> str:
         Returns:
             str: String representation of the Instrument Controllers class.
         """
-        return str(yaml.dump(self.to_dict(), sort_keys=False))
+        return str(YAML().dump(self.to_dict(), io.BytesIO()))
```
AlbertMitjans, line 68: What is this new addition?

tests/test_data_management.py
```diff
@@ -155,3 +155,53 @@ def test_load_results(self, mock_file: MagicMock, mock_makedirs: MagicMock):
 
         mock_file.assert_called_once_with(path, "r")
         mock_makedirs.assert_not_called()
+
+
+class TestPlatformSerializationWithRuamelYaml:
+    """Unit tests to check the serialization is keeping all we want to keep."""
+
+    original_platform = ql.build_platform(Galadriel.runcard)
+    saved_platform = ql.build_platform(save_platform(path="./test.yml", platform=original_platform))
```
AlbertMitjans, line 164: Tests shouldn't create files! We should remove this.

AlbertMitjans: There should be a way of channeling `yaml.dump` into a variable instead of a file

AlbertMitjans (commented):
Added some comments!

src/qililab/experiment/base_experiment.py
GuillermoAbadLopez: In the same post, you can see that there is actually a `.sort()` method, but for the class `YAML()`, so you first would need to have the yaml in such class, order it as you want with lambas, and then dump it.

But as said before, I think we don't need this, since the Default way already conserves the order!

The only problem we would have would be, in the opposite ways, when we actually want a alphabetical order, but I don't think we want that anywhere, no?

src/qililab/experiment/base_experiment.py
GuillermoAbadLopez: answered above

src/qililab/experiment/base_experiment.py
GuillermoAbadLopez: answered above

src/qililab/experiment/base_experiment.py
GuillermoAbadLopez: Did this in my terminal to check it:

src/qililab/instrument_controllers/instrument_controllers.py
GuillermoAbadLopez: `ruamel`, always needs a `stream`, unless you call it from a `context_manager`, so I just added default `stream` from `io` which I found in a stack exchange. 

`ruamel` also says this in their documentation, we could maybe use that:
Screenshot 2023-10-25 at 19 28 11

tests/test_data_management.py
GuillermoAbadLopez: What about removing the file in the end, as I just did, with `os.remove(path)`?

I've seen this from some other parts in qililab, is that acceptable?

src/qililab/experiment/base_experiment.py
AlbertMitjans: We don't want to sort anything, thus we are good! Good job! 🚀 💪

src/qililab/instrument_controllers/instrument_controllers.py
AlbertMitjans: Great!

tests/test_data_management.py
AlbertMitjans: I prefer not do it. I think you already had a bad experience with tests creating files 😆

AlbertMitjans: You can add a stream like you did in other cases!!

AlbertMitjans (commented):
Good job! I will approve when the test stops creating files!

tests/test_data_management.py
GuillermoAbadLopez: Hahha you convinces me already with the first statement 🤣

tests/test_data_management.py
```diff
@@ -161,7 +161,9 @@ class TestPlatformSerializationWithRuamelYaml:
     """Unit tests to check the serialization is keeping all we want to keep."""
 
     original_platform = ql.build_platform(Galadriel.runcard)
-    saved_platform = ql.build_platform(save_platform(path="./test.yml", platform=original_platform))
+    path = save_platform(path="./test.yml", platform=original_platform)
+    saved_platform = ql.build_platform(path)
+    os.remove(path)
```
AlbertMitjans, line 166: Instead of creating and removing a file, why not using `YAML().dump` to dump the information into a stream, and then retrieve it from there?

[See this link for more information.](https://stackoverflow.com/questions/61722242/dump-the-yaml-to-a-variable-instead-of-streaming-it-in-stdout-using-ruamel-yaml)

GuillermoAbadLopez: Yess, that what I said yesterday that I was gonna do! Thanks!!!

I think I have actually done that already above in this discussion, in the console image I pasted! 🙌

GuillermoAbadLopez: @AlbertMitjans I don't see the solution that straight forward, now that I have looked it better... 😬 

Because I wanna test this function:
Screenshot 2023-10-27 at 21 03 15
so I would need to patch in some way the with open, so that it doesn't open a file and saves it into a variable...

Notice I cannot just do what you said, because I would then just not be testing this function that we wanna test 😅

AlbertMitjans: The `save_platform` function is already tested in this file. As you mention in the comments, the tests you added are to make sure ruamel works properly. Thus I see no need of using `save_platform`.

I would suggest using `YAML().dump` directly to obtain the `saved_platform` variable, and then run the tests.

GuillermoAbadLopez: I mean, it makes sense, but I don't know if I completely agree on that. 🤔 

Since the idea I had for this test, was testing that both methods work together properly, kind of like an integration test of both functions. 

What you are proposing is an extended unit test for the `build_platform`, working correctly with the YAML dump function. 

And if in the future someone changes the `save_platform` but not the `build_platform` one, then the test would still pass, when maybe it shouldn't? 🤔  

What do you think? Still, think it's better to not test the integration of both methods?

GuillermoAbadLopez: Since this conversation is going too long, for a single test, I'll do as you say in the next answer hehe 🙃

tests/test_data_management.py
```diff
@@ -155,3 +155,55 @@ def test_load_results(self, mock_file: MagicMock, mock_makedirs: MagicMock):
...
+
+    def test_None_in_serialization(self):
+        """Test that None's are correctly dumped in serialization."""
+        original_firmware = self.original_platform.instruments.elements[4].firmware
+        original_firmware_from_dict = self.original_platform.to_dict()["instruments"][4]["firmware"]
+        saved_firmware = self.saved_platform.instruments.elements[4].firmware
+        saved_firmware_from_dict = self.saved_platform.to_dict()["instruments"][4]["firmware"]
+
+        assert original_firmware == original_firmware_from_dict == saved_firmware == saved_firmware_from_dict
+        assert saved_firmware is None
```
AlbertMitjans, line 209: By the way, could we merge all of these tests in a single one?

Just doing the following assert would suffice, right?
```python
assert original_platform == saved_platform
```

tests/test_data_management.py
AlbertMitjans: Check my new comment. If you want to test that `save_platform` saves the exact same dictionary, I believe one test would suffice, meaning:

```python
original_platform = ql.build_platform(Galadriel.runcard)
path = save_platform(path="./test.yml", platform=original_platform)
saved_platform = ql.build_platform(path)
os.remove(path)
assert original_platform == saved_platform
```

I would still do this test without using `save_platform`:

```python
original_platform = Galadriel.runcard
saved_platform = YAML().dump(original_platform.to_dict(), stream=...)
assert original_platform == saved_platform
```

GuillermoAbadLopez: Bueno, en el primer caldria possar un `to_dict` por ahi, porque sino les dues classes de Platform no seran les mismes

jjmartinezQT: Please Guille, use english so everyone in the team can follow up on these discussions.

GuillermoAbadLopez: Yesyes, I actually, had deleted the comment already!

tests/test_data_management.py
GuillermoAbadLopez: Reduced them by joining them with the already existing tests!

tests/test_data_management.py
GuillermoAbadLopez: In the end, as we commented, I ended up reducing the tests, by joining them with the already existing one very in few lines 👍

tests/test_data_management.py
```diff
@@ -92,20 +92,30 @@ def test_build_method_with_new_drivers(self):
 
     def test_platform_serialization_from_imported_dict(self):
         """Test platform serialization by building a platform, saving it and then load it back again twice. Starting from a given dict."""
-        original_platform = ql.build_platform(Galadriel.runcard)
+        original_dict = Galadriel.runcard
```
AlbertMitjans, line 95: Just in case, can you do a `copy.deepcopy` of this dictionary? To make sure the changes below don't affect other tests.

tests/test_data_management.py
```diff
@@ -92,20 +92,30 @@ def test_build_method_with_new_drivers(self):
...
+
+        original_platform = ql.build_platform(original_dict)
+
         path = save_platform(path="./test.yml", platform=original_platform)
         saved_platform = ql.build_platform(path)
+
         new_path = save_platform(path="./test.yml", platform=saved_platform)
         new_saved_platform = ql.build_platform(new_path)
 
-        with open(file="examples/runcards/galadriel.yml", mode="r", encoding="utf8") as yaml_f:
```
AlbertMitjans, line 101: This file doesn't exist, how is this even working?

AlbertMitjans: Same thing for the test below

tests/test_data_management.py
```diff
@@ -124,7 +134,7 @@ def test_platform_serialization_from_yaml_file(self):
 
         for i in ["name", "device_id", "chip", "instruments", "instrument_controllers"]:
             assert yaml_f_dict[i] == generated_f_dict[i]
-        assert generated_f_dict != Galadriel.runcard
```
AlbertMitjans, line 127: Why did you remove this?

AlbertMitjans (commented):
Added some comments!

tests/test_data_management.py
GuillermoAbadLopez: Agree, I'll do that!

tests/test_data_management.py
GuillermoAbadLopez: What do you mean it doesn't exist? It is there: 
https://github.com/qilimanjaro-tech/qililab/blob/main/examples/runcards/galadriel.yml

tests/test_data_management.py
GuillermoAbadLopez: Explained in the answer above!

tests/test_data_management.py
GuillermoAbadLopez: The thing is, in this test we were testing `Galadriel.runcard` from `data.py`, and we just also read the `yaml` in `galadriel.yaml`, to then check that they are not the same... but that doesn't make any sense to test..

AlbertMitjans: Ok my bad.

AlbertMitjans: I don't like at all using files outside the `test` folder inside tests. Good thing you deleted it

AlbertMitjans: > Like why would they be different? I mean they are, because they are to different dictionaries, but it doesn't;t make sense to add here a line for the sake of checking that, since that is no functionality at all 🙃

Completely agree!

tests/test_data_management.py
AlbertMitjans: Totally agree!

AlbertMitjans (approved):
Good job!! 💪

PR #581, opened 2023-10-26
Allow multi-measurements on one qubit

visagim:


visagim:
wrong branch this one is merging to new main, should be deleted

PR #585, opened 2023-10-27
delete _init_() docstrings!

GuillermoAbadLopez:
PR moving/deleteing the docstrings from _init_ to class level docstrings, since that's what we decided.


src/qililab/drivers/instruments/qblox/pulsar.py
```diff
@@ -25,15 +25,13 @@
 
 @InstrumentDriverFactory.register
 class Pulsar(QcodesPulsar, BaseInstrument):  # pylint: disable=abstract-method
-    """Qililab's driver for QBlox-instruments Pulsar"""
+    """Qililab's driver for QBlox-instruments Pulsar
+    Args:
```
AlbertMitjans, line 29: There should always be a blank line between the description and the arguments.

```suggestion
    """Qililab's driver for QBlox-instruments Pulsar

    Args:
```

GuillermoAbadLopez: Thanks commited!

when tests pass I'll merge it!

PR #587, opened 2023-10-30
add a couple new enums for SNZ gate

paulqili:
i guess you may need to add some tests. It's just two new parameters for the SNZ gate


jjmartinezQT (approved):
Ready to go.

PR #589, opened 2023-10-31
Parameters add gate options

visagim:
Add the gate options to parameters so we can set the value for cz phase corrections through `platform.set_parameter`


AlbertMitjans (approved):
Approved!

PR #595, opened 2023-11-06
Auto-calibration (graph error handling)

IsaacLA12:
This PR adds logic to the `CalibrationController` class in order to handle some errors on the workflow it may ocurre.
We found three possible errors that can happen in an automatic calibration execution:

- The case all your nodes in the automatic calibration graph passes the `check_status`. In that case we could have a situation where the algorithm does not really calibrate anything, this situation can happen when nodes have high drift timeouts specified by the user. The solution for this is to add a flag `force_maintain` that will be set to `True` for the first call of the `maintain` for the leaves in our graph.

                                       (check_status result: ✅pass)        ✅     ✅
                          (check_data result: 🟢in_spec, 🔴bad_data)        🟢----->🔴

- The case a node is classified as `out_spec` when it should be `bad_data` and vice versa (see below). This situations can be detected on the `diagnose` subroutine. The main purpose of `diagnose` is "finding where the `out_specs` are" while our solution finds and actually checks "the nodes where all the dependancies are `in_spec`". Those two statements are equivalent as `out_spec` nodes will have all the dependancies `in_spec` but we actually check them where as the original implementation relies on the fact that the user defined a good comparison models and thresholds to properly classify nodes `out_spec`.

                               (check_data result: 🟢in_spec, 🔴bad_data, 🟣out_spec)
        🟢---->🔴---->🔴---->🔴                 🟢---->🟣---->🟣---->🔴            🟢---->🔴---->🟣---->🔴 

On the solution proposed for `diagnose` we will keep both implementations and the user can choose which one to use with the flag `safe_check`. Running diagnose with `safe_check=True` will not only rely on the `check_data` node classification but also check the dependancies are `in_spec`, therefore the running time and the resources to perform this extra checks will be higher.

A more precise description of the errors to handle can be found in [Notion](https://www.notion.so/qilimanjaro/Automatic-Calibration-53cef01488db4a44a7ecf0f475e08084).

## GUILLE's TODO's for closing this PR:
- [ ] finish tests
- [ ] check you have modified all the documentation, with the new workflow


src/qililab/calibration/calibration_controller.py
```diff
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# pylint: disable=anomalous-backslash-in-string
```
GuillermoAbadLopez, line 15: Docstring no la toque illo! We have to use the ones from the main autocalibration branch, after checking this works in the lab 👍

GuillermoAbadLopez (changes requested):
Docstring no la toque illo! We have to use the ones from the main autocalibration branch, after checking this works in the lab 👍

GuillermoAbadLopez:
Closing this PR, since it is moved to the following:
https://www.notion.so/qilimanjaro/Improve-check_data-and-comparisons-to-do-2d-comparisons-multiple-comparisons-together-and-always-c-23f80d46c61446618bd92fa79392ee73?pvs=4

PR #596, opened 2023-11-06
[TAS-466] Handle results from Quantum Machines

fedonman:


AlbertMitjans:
* A link back to this PR.

src/qililab/qprogram/quantum_machines_compiler.py
```diff
@@ -163,36 +166,40 @@ def traverse(block: Block):
...
-                    input2 = input2.buffer(loop_iteration)
-                input2.save_all(f"adc1_{i}")
-                input2.save_all(f"adc2_{i}")
-                measurement.result_handles.append(f"adc1_{i}")
-                measurement.result_handles.append(f"adc2_{i}")
-
-            measurements.append(measurement)
+                stream_Q.save(f"Q_{index}")
+                result_handles.append(f"Q_{index}")
+            if measurement.stream_raw_adc is not None:
```
jjmartinezQT, line 187: Just a question, I guess this means both adc1 and adc2 always are present together? If one exists so there is the other?

jjmartinezQT (approved):
Very good job on this one, only a question but this is ready to go.

src/qililab/qprogram/quantum_machines_compiler.py
fedonman: Nice catch. I though so, and it is true for the drive buses, but I looked more carefully in the examples and it actually depends on if the element has one or two outputs defined.

```
"rr1": "outputs": {"out1": ("con1", 2)},
```

vs

```
"rr1": "outputs": {"out1": ("con1", 1), "out2": ("con1", 2)},
```

But this information is currently not know to the compiler (since it depends on the runcard provided).

fedonman: I refactored this part, so `input1` and `input2` are processed separately, and added a `TODO` to make the check. (it's better to implement this after having the translation from runcard to dict, so we figure out what is the best way to pass the information to the compiler)

jjmartinezQT: I agree with the refactor and then check once we have the translation.

PR #598, opened 2023-11-06
Feature/real time saving results

jjmartinezQT:
This PR brings the possibility to use real time saving results capability.


src/qililab/result/stream_results.py
```diff
@@ -0,0 +1,82 @@
...
+    This class wraps a numpy array and adds a context manager to save results on real time while they are acquired by
+    the instruments.
+
+    Args:
+        shape (ndarray.shape): results array shape.
+        path (str): path to save results.
+        loops (dict[str, ndarray]): dictionary with each loop name in the experiment as key and numpy array as values.
+    """
+
+    def __init__(self, shape, path, loops):
```
AlbertMitjans, line 17: Don't forget to add types! 🙏

tests/result/test_stream_results.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def test_stream_array_instantiation(self, stream_array: StreamArray):
+        """Tests the instantiation of a StreamArray object."""
+        assert stream_array.results.shape == (2, 2)
+        assert (stream_array.results == np.empty(shape=(2, 2))).all
+        assert stream_array.path == "test_stream_array.hdf5"
+        assert stream_array.loops == {"test_amp_loop": np.arange(0, 1, 2)}
+
+    @patch("qililab.result.stream_results.h5py.File", return_value=MockFile())
+    def test_context_manager(self, mock_h5py: MockFile, stream_array: StreamArray):  # pylint: disable=unused-argument
```
AlbertMitjans, line 63: We should use `mock_h5py` to assert that the right calls are being made

src/qililab/result/stream_results.py
```diff
@@ -0,0 +1,82 @@
...
+        self.results = np.empty(shape=shape)
```
AlbertMitjans, line 20: Let's use zeros instead:
image

src/qililab/result/stream_results.py
jjmartinezQT: added now.

src/qililab/result/stream_results.py
```diff
@@ -0,0 +1,83 @@
...
+        path (str): path to save results.
+        loops (dict[str, np.ndarray]): dictionary with each loop name in the experiment as key and numpy array as values.
+    """
+
+    def __init__(self, shape: tuple, path: str, loops: dict[str, np.ndarray]):
+        self.results = np.empty(shape=shape)
+        self.path = path
+        self.loops = loops
+        self.file: h5py.File | None = None
+        self.dataset = None
```
AlbertMitjans, line 24: Will users use these attributes? If not, let's make them private 🙏

AlbertMitjans: I would keep `path` and `loops` public though, because these are the values they give in the init (this is only my humble opinion).

In other words, I expect to have access to the information I provided to a class. Here's an example:

```pycon
>>> a = A(hello="world")
>>> a.hello
world
```

src/qililab/result/stream_results.py
```diff
@@ -0,0 +1,84 @@
+from typing import Any
+
+import h5py
+import numpy as np
+
+
+class StreamArray:
```
AlbertMitjans, line 7: Finally, I want to ask for a change that is solely to improve the user experience.

This is something that I've been wanting to do for a long time now, and this is the perfect example to start doing it.

If you check popular python libraries (like numpy), you'll see that all of its classes/functions are shown in lowercase to the user, for example:
```pycon
>>> import numpy as np
>>> ar = np.array([1, 2])  # array is a class, but it is lowercase!
```

This is solely to improve user experience. The user shouldn't care if something is a class or not, thus they shouldn't care about lower/upper case letters.

In PennyLane, they solved this in a way that I really liked: they kept the syntax of upper-case letters for classes, but they created "constructors" (a.k.a functions) in lowercase that returned an instance of the class. In the PennyLane main `__init__` file, only these lower-case functions were imported (similar thing for documentation). [See an example in this link](https://github.com/PennyLaneAI/pennylane/blob/f25eb134d1aaca35c9cd886cd38f4283b457728e/pennylane/ops/op_math/pow.py#L39).

What do you think of doing the same thing? In this case it would be something like this:
```suggestion
def stream_results(shape: tuple, path: str, loops: dict[str, np.ndarray]):  # this is the constructor
    return StreamArray(shape=shape, path=path, loops=loops)

class StreamArray:
```

If we do this, the big chunk of the documentation (long explanation + examples) should be moved to this user-facing method, because this is what users will see. 😄 

Let me know if you have any questions!

AlbertMitjans (commented):
Amazing job!! I added some comments 😄

src/qililab/result/stream_results.py
jjmartinezQT: I like the idea actually and it we enforce all along qililab little by little I think it will indeed improve user experience.

src/qililab/result/stream_results.py
jjmartinezQT: addressed.

jjmartinezQT: addressed

src/qililab/result/stream_results.py
jjmartinezQT: agree with this change.

src/qililab/result/stream_results.py
```diff
@@ -0,0 +1,97 @@
...
+def stream_results(shape: tuple, path: str, loops: dict[str, np.ndarray]):
+    """Constructs a StreamArray instance.
+
+    This methods serves as a constructor for user interface of the StreamArray class.
+
+    Args:
+        shape (tuple): results array shape.
+        path (str): path to save results.
+        loops (dict[str, np.ndarray]): dictionary with each loop name in the experiment as key and numpy array as values.
+    """
```
AlbertMitjans, line 16: Can we add an example? 🙏

AlbertMitjans (commented):
Can we import the `stream_results` function in the top init, and add it into the documentation? 🙏 

We should also prepare a demo for next sprint review, such that HW knows how to use this feature!

tests/result/test_stream_results.py
```diff
@@ -0,0 +1,87 @@
...
+    def __init__(self):
+        """Initialize a mock file."""
+        self.dataset = None
+
+    def create_group(self, name: str):  # pylint: disable=unused-argument
+        return MockGroup()
+
+    def create_dataset(self, name: str, data: np.ndarray):  # pylint: disable=unused-argument
+        """Creates a dataset"""
+        return data
```
AlbertMitjans, line 47: You should return a `copy.deepcopy` of this!

src/qililab/result/stream_results.py
```diff
@@ -0,0 +1,168 @@
...
+        self._dataset = None
+
+    def __setitem__(self, key: str, value: float):
+        """Sets and item by key and value in the dataset.
+
+        Args:
+            key (str): key for the item to save.
+            value (float): value to save.
+        """
+        self._results[key] = value
```
AlbertMitjans, line 117: We should also modify the `self.dataset` attribute!

tests/result/test_stream_results.py
AlbertMitjans: This is because h5py creates a copy of the data before putting it inside the dataset

docs/releases/changelog-dev.md
```diff
@@ -2,6 +2,11 @@
 
 ### New features since last release
 
+- Added real time results saving capability.
+  [#598](https://github.com/qilimanjaro-tech/qililab/pull/598)
+
+- Added the two main classes need for automatic-calibration, `CalibrationController` and `CalibrationNode`: [#554](https://github.com/qilimanjaro-tech/qililab/pull/554)
```
AlbertMitjans, line 8: This should not be here, right? 😆

src/qililab/result/stream_results.py
```diff
@@ -0,0 +1,212 @@
...
+    Args:
+        shape (tuple): results array shape.
+        path (str): path to save results.
+        loops (dict[str, np.ndarray]): dictionary with each loop name in the experiment as key and numpy array as values.
+
+    Examples:
+
+        .. note::
+
+            All the following examples are explained in detail in the :ref:`Platform <platform>` section of the documentation. However, here are a few thing to keep in mind:
```
AlbertMitjans, line 34: Juanjo did you copy this from somewhere else? 😅 

Let's try to keep the examples short and concise please! Otherwise nobody will read them 🙏 

I don't think you should be talking about the Platform here... you should only add an example of how this function works. If you want (and if needed), you can always add a **small** note that links to the platform documentation.

So please, remove everything in this example that is not related to the `stream_results` function 🙏

AlbertMitjans: Also, can you add an extra example that explains how to use this with the `%%submit_job` magic method of SLURM?

AlbertMitjans (commented):
Just some minor comments, but almost good to go!

src/qililab/drivers/instruments/qblox/test_cluster.ipynb
AlbertMitjans, line 1: What is this file? Juanjo please make sure to check the PR changes before marking it ready for review!

src/qililab/result/stream_results.py
```diff
@@ -0,0 +1,277 @@
...
+            # Connecting and setting up the platform:
+            platform.connect()
+            platform.initial_setup()
+            platform.turn_on_instruments()
+
+            # Define circuit
+            circuit = Circuit(QUBIT + 1)
+            circuit.add(ql.Drag(QUBIT, theta=np.pi, phase=0))
+            circuit.add(ql.Wait(QUBIT, M_BUFFER_TIME))
+            circuit.add(gates.M(QUBIT))
```
AlbertMitjans, line 63: Again, this example is way too long. It contains a lot of information that is not related to this method's functionality.

All the lines I marked above can be deleted. A good example could be:

```python
import qililab as ql

stream_array = ql.stream_results(shape=(5, 2), loops={"amp": AMP_VALUES}, path="results.h5")

with stream_array:
    for i in range(5):
        stream_array[i, 0] = i
```
```pycon
>>> stream_array
# Show output
>>> ql.load_results("results.h5")
# Show output
```

AlbertMitjans: The example above is super short, copy-pasteable, and concise (it shows the functionality of the feature).

AlbertMitjans: Then, if we want to add more information (for example, related to the slurm magic method) we can add an extra example below:

AlbertMitjans: For example:

> .. note::
> 
>     The output of `stream_results` is not picklable, thus it cannot be returned in a SLURM job. In this scenario we suggest copying the array with the data into a different variable and returning that variable instead:
> 
>     .. code-block:: python
> 
>         %%submit_job -o results -d galadriel
>         stream = ql.stream_results(shape=(5, 2), loops={"amp": AMP_VALUES}, path="results.h5")
>         with stream:
>             for i in range(5):
>                 # Here you can execute any algorithm you want
>                 stream[i] = i
>         results = stream.results

src/qililab/result/stream_results.py
```diff
@@ -0,0 +1,174 @@
...
+        loops (dict[str, np.ndarray]): dictionary with each loop name in the experiment as key and numpy array as values.
+
+    Examples:
+
+        **1. Executing an experiment and saving the results live time:**
+
+
+        To execute an experiment you first need to define the loops and the values you want to loop for, for example, a Rabi experiment
+        where you will loop over amplitude values, measuring the results out of the amplitude in each of the steps.
+        Then you also need to build, connect, set up, and execute the platform, which together look like:
```
AlbertMitjans, line 37: This text is completely unrelated to the code below

src/qililab/result/stream_results.py
```diff
@@ -0,0 +1,174 @@
...
+            [3. 0.]
+            [4. 0.]]
+
+            >>> ql.load_results("results.h5")
+
+            (array([[0., 0.],
+                [1., 0.],
+                [2., 0.],
+                [3., 0.],
+                [4., 0.]]), {'amp': array([0.  , 0.25, 0.5 , 0.75, 1.  ])})
```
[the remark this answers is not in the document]
AlbertMitjans: We might also need to remove the empty space between the command and the output

src/qililab/result/stream_results.py
```diff
@@ -0,0 +1,171 @@
...
+         [0.5  0.  ]
+         [0.75 0.  ]
+         [1.   0.  ]]
+
+        >>> ql.load_results("results.h5")
+        (array([[0., 0.],
+               [0.25, 0.],
+               [0.5, 0.],
+               [0.75, 0.],
+               [1., 0.]]), {'amp': array([0.  , 0.25, 0.5 , 0.75, 1.  ])})
```
AlbertMitjans, line 62:
```suggestion
               [1., 0.]]), {'loop_name': array([0.  , 0.25, 0.5 , 0.75, 1.  ])})
```

src/qililab/result/stream_results.py
```diff
@@ -0,0 +1,171 @@
...
+    .. note::
+
+        The output of `stream_results` is not picklable, thus it cannot be returned in a SLURM job. In this scenario we suggest copying the array with the data into a different variable and returning that variable instead:
+
+        .. code-block:: python
+
+            %%submit_job -o results -d galadriel
+            LOOP_VALUES = np.linspace(0, 1, 5)
+            stream_array = ql.stream_results(shape=(5, 2), loops={"loop_name": LOOP_VALUES}, path="results.h5")
+            with stream:
```
AlbertMitjans, line 73:
```suggestion
            with stream_array:
```

src/qililab/result/stream_results.py
```diff
@@ -0,0 +1,171 @@
...
+                for i, value in enumerate(LOOP_VALUES):
+                    # Here you can execute any algorithm you want
+                    stream_array[(i, 0)] = value
+
+            results = stream.results
```
AlbertMitjans, line 78:
```suggestion
            results = stream_array.results
```

AlbertMitjans (approved):
Good job!! 🚀 

I left some comments but it is ready to go!

PR #600, opened 2023-11-09
Add Ipython magic method to queue cells to SLURM

AlbertMitjans:
https://www.notion.so/qilimanjaro/SLURM-Integration-with-python-nb-c8cd0c3e3b404844bc86acd47c645dfe


src/qililab/slurm.py
```diff
@@ -0,0 +1,62 @@
...
+    executable_code = "\n".join(import_lines + [cell])
+
+    # Create the executor that will be used to queue the SLURM job
+    folder_path = ".slurm_job_data"
+    executor = AutoExecutor(folder=folder_path)
+    executor.update_parameters(slurm_partition=partition)
+
+    # Compile the code defined above
+    code = compile(executable_code, "<string>", "exec")
+    # Take all the variables from the Jupyter Notebook plus the output variable
```
AlbertMitjans, line 37:
```suggestion
    # Take all the variables defined by the user in the Jupyter Notebook and the output variable
```

src/qililab/slurm.py
```diff
@@ -0,0 +1,62 @@
+import os
+from types import ModuleType
+
+from IPython.core.magic import needs_local_scope, register_cell_magic
+from IPython.core.magic_arguments import argument, magic_arguments, parse_argstring
+from submitit import AutoExecutor
+
+
+@magic_arguments()
+@argument("-o", "--output", help=("Name of the variables that will be returned from the SLURM job."))
```
AlbertMitjans, line 10: Suggestion of improvement of the docstring:
```suggestion
@argument("-o", "--output", help=("Output of the SLURM job. This name should correspond to a variable defined in the cell that we want to retrieve after execution. After queuing a cell, this variable will be converted to a `Job` class. To retrieve the results of the job, you need to call `variable.result()`."))
```

src/qililab/slurm.py
```diff
@@ -0,0 +1,76 @@
...
+@argument("-d", "--device", help=("Name of the device where you want to execute the SLURM job."))
+@argument(
+    "-l",
+    "--logs",
+    default="./slurm_job_data",
+    help=(f"Path where you want slurm to write the logs for the last {num_jobs_to_keep} jobs"),
+)
+@argument("-n", "--name", default="submitit", help=("Name of the slurm job"))
```
AlbertMitjans, line 22: Nice!!

src/qililab/slurm.py
```diff
@@ -0,0 +1,76 @@
...
+from IPython.core.magic import needs_local_scope, register_cell_magic
+from IPython.core.magic_arguments import argument, magic_arguments, parse_argstring
+from submitit import AutoExecutor  # pylint: disable=import-error
+
+logger = logging.getLogger()
+num_jobs_to_keep = 10
+
+
+@magic_arguments()
+@argument("-o", "--output", help=("Name of the variables that will be returned from the SLURM job."))
```
AlbertMitjans, line 14: I happened to me (several times now) that the name of the output doesn't correspond to the name of the variable that I use inside the cell.

What do you think of checking that the name inside `output` corresponds to a variable defined inside the cell?

Here you have a potential implementation of this check:

```python
import ast

def is_variable_used_or_defined(code, variable):
    tree = ast.parse(code)
    for node in ast.walk(tree):
        if isinstance(node, ast.Name) and node.id == variable:
            return True
        if isinstance(node, ast.Assign):
            for target in node.targets:
                if isinstance(target, ast.Name) and target.id == variable:
                    return True
    return False

code = """
results = []

for gain in np.arange(0, 1, 100):
    a = np.array([1, 2, 3])
    results.append(a)
"""

print(is_variable_used_or_defined(code, 'results'))
```

AlbertMitjans: The idea is using the function above to check that the `output` name that the user provides is actually used inside the code of the cell.

ziiiki: makes sense, yes!

ziiiki: Finally, I removed the` if isinstance(node, ast.Name) and node.id == variable` because if the output variable is defined in another cell and any values are assigned to it in the magic cell, submitit considers it is None. 
image

AlbertMitjans:
* A link back to this PR.

src/qililab/slurm.py
AlbertMitjans: Nice catch!!

examples/example_using_qililab.ipynb
AlbertMitjans, line 1: Careful, you deleted this notebook!

src/qililab/__init__.py
```diff
@@ -27,3 +30,9 @@
 from .utils import Loop, Wait
 from .utils.load_data import load
 from .waveforms import *
+
+with contextlib.suppress(NameError, ImportError):
+    # Since Ipython magic methods can only be imported from inside a Jupyter Notebook,
+    # here we first check that `get_ipython` exists (which means we are inside a Jupyter Notebook)
+    get_ipython()  # type: ignore  # noqa: F405 # pylint: disable=undefined-variable
+    from .slurm import submit_job
```
AlbertMitjans, line 38: Weird, this line is not covered... This means that the function `submit_job` is never imported! How are the tests passing then? 🤔

src/qililab/slurm.py
```diff
@@ -0,0 +1,113 @@
...
+
+    # Define the function that will be queued as a SLURM job
+    def function(code, variables):
+        # Execute the code and return the output variable defined by the user
+        exec(code, variables)
+        return variables[output]
+
+    # Check if output variables are defined or used in the magic cell
+    if not is_variable_used(executable_code, output):
+        raise ValueError(f"Output variable '{output}' was not assigned to any value inside the cell!")
```
AlbertMitjans, line 93: We need to add a test for this corner case

src/qililab/slurm.py
```diff
@@ -0,0 +1,113 @@
...
+    # Submit slurm job
+    job = executor.submit(function, code, variables)
+    if execution_env == "local":
+        logger.warning("Your slurm job '%s' with ID %s is running locally, without slurm!", job_name, job.job_id)
+    else:
+        logger.info(f"Your slurm job '{job_name}' with ID {job.job_id} was sent to {partition}!")
```
AlbertMitjans, line 99: Do we really need to make this distinction? If so, we will not be able to test the second logger line

src/qililab/slurm.py
```diff
@@ -0,0 +1,113 @@
...
+
+    job_ids_to_keep = list(range(int(job.job_id) - num_jobs_to_keep, int(job.job_id) + 1))
+
+    # Delete info from past jobs
+    for filename in os.listdir(folder_path):
+        file_path = os.path.join(folder_path, filename)
+        try:
+            if int(filename.split("_")[0]) not in job_ids_to_keep:
+                os.remove(file_path)
+        except ValueError:
```
AlbertMitjans, line 111: It would also be nice to test this, by adding a random file before executing a cell.

examples/example_using_qililab.ipynb
ziiiki: wow, sorry for that!

docs/releases/changelog-dev.md
```diff
@@ -1,6 +1,9 @@
 # Release dev (development release)
 
 ### New features since last release
+- Add magic method to run python code as slurm jobs from Jupyter Notebooks.
+[#600](https://github.com/qilimanjaro-tech/qililab/pull/600)
+- Added the two main classes need for automatic-calibration, `CalibrationController` and `CalibrationNode`: [#554](https://github.com/qilimanjaro-tech/qililab/pull/554)
```
AlbertMitjans, line 6:
```suggestion
[#600](https://github.com/qilimanjaro-tech/qililab/pull/600)

- Added the two main classes need for automatic-calibration, `CalibrationController` and `CalibrationNode`: [#554](https://github.com/qilimanjaro-tech/qililab/pull/554)
```

src/qililab/slurm.py
ziiiki: yep, done!

src/qililab/slurm.py
ziiiki: not really

src/qililab/slurm.py
ziiiki: yep, done!

src/qililab/__init__.py
ziiiki: my guess this is because I was doing `import qililab.slurm` instead of `import qililab` because submit_job was not found in the latter case. finally I do `session_ip.run_cell(raw_cell="import qililab.slurm\nimport qililab")` to cove both cases

examples/example_using_qililab.ipynb
AlbertMitjans, line 1: There shouldn't be any changes here! Can you checkout these changes from this branch? 🙏

src/qililab/slurm.py
```diff
@@ -0,0 +1,111 @@
...
+    local_ns[output] = job
+
+    job_ids_to_keep = list(range(int(job.job_id) - num_jobs_to_keep, int(job.job_id) + 1))
+
+    # Delete info from past jobs
+    for filename in os.listdir(folder_path):
+        file_path = os.path.join(folder_path, filename)
+        try:
+            if int(filename.split("_")[0]) not in job_ids_to_keep:
+                os.remove(file_path)
```
AlbertMitjans, line 108: This line is not covered! Can we test this? 🙏

tests/test_slurm.py
```diff
@@ -0,0 +1,61 @@
...
+        line="-o results -d debug -l slurm_job_data_tests -n unit_test -e local",
+        cell="results = a+b ",
+    )
+    assert ip.user_global_ns["results"].result() == 2
+
+
+def test_submit_job_output_not_assigned(ip):
+    """Check ValueError is raised in case values are not assigned to the output variable within the magic cell."""
+    ip.run_cell(raw_cell="a=1\nb=1")
+    with pytest.raises(ValueError):
```
AlbertMitjans, line 33: Always use `match` to catch the error message 🙏 

```suggestion
    with pytest.raises(ValueError, match="Error message here"):
```

tests/test_slurm.py
```diff
@@ -0,0 +1,61 @@
...
+        ip.run_cell_magic(
+            magic_name="submit_job",
+            line="-o results -d debug -l slurm_job_data_tests -n unit_test -e local",
+            cell="a+b",
+        )
+
+
+def test_submit_job_with_random_file_in_logs_folder(ip):
+    """Check ValueError is raised if non-submitit files are found in logs folder"""
```
AlbertMitjans, line 42: We are not raising an error anymore! We should change the docstring.

tests/test_slurm.py
```diff
@@ -0,0 +1,61 @@
...
+    ip.run_cell(
+        raw_cell="import os\nfolder_path='slurm_job_data_tests'\nfile_name='abc.py'\nfile_path = os.path.join(folder_path, file_name)\nif not os.path.exists(folder_path):\tos.makedirs(folder_path)\nopen(file_path, 'w')\na=1\nb=1"
+    )
+    ip.run_cell_magic(
+        magic_name="submit_job",
+        line="-o results -d debug -l slurm_job_data_tests -n unit_test -e local",
+        cell="results=a+b",
+    )
```
AlbertMitjans, line 50: There is no assert here! Thus this test is doing nothing!

We should assert that the file is deleted

tests/test_slurm.py
```diff
@@ -0,0 +1,61 @@
...
+
+def test_submit_job_delete_info_from_past_jobs(ip):
+    """Check ValueError is raised if non-submitit files are found in logs folder"""
+    ip.run_cell(raw_cell="a=1\nb=1")
+    for _ in range(11):
+        ip.run_cell_magic(
+            magic_name="submit_job",
+            line="-o results -d debug -l slurm_job_data_tests -n unit_test -e local",
+            cell="results=a+b",
+        )
```
AlbertMitjans, line 61: Same here, we are not asserting anything!

We could assert the following:
- That up until 10 jobs, we have N files in the logs folder (where N is the iteration index of the for loop times the amount of files created per job).
- That in the last job we only have the files of the last job.

tests/test_slurm.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def test_submit_job(ip):
+    ip.run_cell(raw_cell="a=1\nb=1")
+    ip.run_cell_magic(
+        magic_name="submit_job",
+        line="-o results -d debug -l slurm_job_data_tests -n unit_test -e local",
+        cell="results = a+b ",
+    )
+    assert ip.user_global_ns["results"].result() == 2
```
AlbertMitjans, line 27: Can you also add asserts checking that the arguments work? For example: check that files are created inside the `slurm_job_data_tests` folder.

AlbertMitjans, line 27: Are these tests creating actual files? If so, we should make sure to delete them at the end of every test!

We can do this with [teardown methods](https://stackoverflow.com/questions/26405380/how-do-i-correctly-setup-and-teardown-for-my-pytest-class-with-tests)

examples/example_using_qililab.ipynb
ziiiki: done!

tests/test_slurm.py
ziiiki: in the first case, we should assert there are 5*10=50 files since every job produces 5 different files. In the second one, 5

tests/test_slurm.py
ziiiki: true

tests/test_slurm.py
ziiiki: yes, this seems quite powerful https://docs.pytest.org/en/latest/how-to/fixtures.html#teardown-cleanup-aka-fixture-finalization. But I don't think there's need to implement such a thing for this tests. Thus, I implemented a manual teardown :)

tests/test_slurm.py
ziiiki: In fact, there is a problem I do not know how to solve. When targeting local instead of slurm, job id's don't follow a clear pattern. As those lines are covered in other tests and are simple, I think we can skip this test. To sum up the behaviour depending on the env:

LOCAL: only keep current job files in logs folder
SLURM: keep logs for last 10 jobs

src/qililab/slurm.py
ziiiki: see comments below to understando why I skipped this line

ziiiki (approved):

src/qililab/slurm.py
```diff
@@ -0,0 +1,111 @@
...
+import os
+from types import ModuleType
+
+from IPython.core.magic import needs_local_scope, register_cell_magic
+from IPython.core.magic_arguments import argument, magic_arguments, parse_argstring
+from submitit import AutoExecutor
+
+from qililab.config import logger
+
+num_files_to_keep = 20  # needs to be a multiple of 4 and 5: 20,40,60,80..
```
AlbertMitjans, line 11: Here you use 20, but in the tests you say 60... Can we increase it to 60?

tests/test_slurm.py
```diff
@@ -0,0 +1,92 @@
...
+    ip.run_cell_magic(
+        magic_name="submit_job",
+        line=f"-o results -d debug -l {slurm_job_data_test} -n unit_test -e local",
+        cell="results=a+b",
+    )
+    time.sleep(4)
+    assert (
+        len([f for f in os.listdir(slurm_job_data_test) if os.path.isfile(os.path.join(slurm_job_data_test, f))])
+        == ql.slurm.num_files_to_keep
+    )
```
AlbertMitjans, line 90: Why do we still have 20 files? Shouldn't we have only 4 files (the files from the last job)?

AlbertMitjans (commented):
Almost good to go! Just one comment and one question!

tests/test_slurm.py
ziiiki: I'm only deleting the oldest job. So: 20-4+4=20

AlbertMitjans: Ahhh truee my bad!!

AlbertMitjans: Thanks for the clarification 😄

src/qililab/slurm.py
ziiiki: done!

PR #601, opened 2023-11-10
Autocalibration, Comparison models

GuillermoAbadLopez:
PR for adding the comparison models:

## TODOs:
- [x] Check documentation of the comparisons
- [x] Check test coverage, and that it covers all the cases/logic.
- [x] Add tests, for everything that was changed in `comparison_models`, due to lab_test
- [x] Check why comparison models, are returning huge values and find a logic in the numbers.
-- drag was returning exceptionally high errors, in the result due to how the experiment handles data. inside, so the thresholds in that one will need to be highly increased!


AlbertMitjans (approved):
I haven't looked into much detail. But since we will run "hardware tests" of this every night I believe we can already merge it 🚀

AlbertMitjans:
> @AlbertMitjans also, it doesn't affect anything outside calibration, so if anything fails, it won't be a fire! 👍

Exactly 💯

PR #602, opened 2023-11-10
Add Time of Flight between readout pulse `play` and `acquire`

paulqili:
This adds a `wait` operation between the readout pulse and the acquisition time. This is the time the pulse takes to go to the fridge and come back

### DIFF
--- a/src/qililab/instruments/awg_analog_digital_converter.py
+++ b/src/qililab/instruments/awg_analog_digital_converter.py
@@ -38,7 +38,7 @@ class AWGAnalogDigitalConverter(AWG):
 
         acquisition_delay_time: int  # ns
         awg_sequencers: Sequence[AWGADCSequencer]
-        time_of_flight: int = 146
+        time_of_flight: int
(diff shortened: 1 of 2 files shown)


src/qililab/instruments/awg_analog_digital_converter.py
```diff
@@ -38,6 +38,7 @@ class AWGAnalogDigitalConverterSettings(AWG.AWGSettings):
 
         acquisition_delay_time: int  # ns
         awg_sequencers: Sequence[AWGADCSequencer]
+        time_of_flight: int
```
fedonman, line 41: There is also an `acquisition_delay_time` attribute in the same class. What is the difference?

paulqili: that is a delay between the last control pulse and the readout pulse. TOF is the delay between sending the pulse and startting integration

fedonman:
This has been addressed in another PR.

PR #608, opened 2023-11-17
Add time limit parameter to the slurm command.

AlbertMitjans:


visagim:
I take it there are no tests for this?

visagim (dismissed):
since we don't need to add tests it's good to go

AlbertMitjans:
> since we don't need to add tests it's good to go

We do need to add tests! My bad 😢 

I added them! Can you take a look?

visagim (approved):
cool! good to go

PR #610, opened 2023-11-17
Add option to choose when to execute a job

AlbertMitjans:


src/qililab/slurm.py
```diff
@@ -36,6 +36,14 @@ def is_variable_used(code, variable):
 @argument("-p", "--partition", help="Name of the partition where you want to execute the SLURM job.")
 @argument("-n", "--name", default="submitit", help="Name of the slurm job.")
 @argument("-t", "--time", default=120, help="Time limit (in minutes) of the job.")
+@argument(
+    "-b",
+    "--begin",
+    default="now",
+    help="Submit the batch script to the Slurm controller immediately, like normal, "
+    "but tell the controller to defer the allocation of the job until the specified time. The time format can be"
+    " either `HH:MM:SS`, `now+1hour`, `now+60minutes`, `now+60` (seconds by default).",
```
ziiiki, line 45: can even specify the day with the following format:  2023-11-17T17:22:00

tests/test_slurm.py
```diff
@@ -102,10 +102,10 @@ def test_setting_parameters(ip):
         with patch("qililab.slurm.os"):
```
ziiiki: do we need any assert here? Perhaps I would do now+1 and check the assert the result

ziiiki (approved):

src/qililab/slurm.py
AlbertMitjans: Added example!

PR #612, opened 2023-11-17
Auto calibration/graph error handle and lab first try feedback changes

GuillermoAbadLopez:
# ISAAC SUMMARY of error handling:

This PR adds logic to the `CalibrationController` class in order to handle some errors on the workflow it may ocurre.
We found three possible errors that can happen in an automatic calibration execution:

- The case all your nodes in the automatic calibration graph passes the `check_status`. In that case we could have a situation where the algorithm does not really calibrate anything, this situation can happen when nodes have high drift timeouts specified by the user. The solution for this is to add a flag `force_maintain` that will be set to `True` for the first call of the `maintain` for the leaves in our graph.

                                       (check_status result: ✅pass)        ✅     ✅
                          (check_data result: 🟢in_spec, 🔴bad_data)        🟢----->🔴

- The case a node is classified as `out_spec` when it should be `bad_data` and vice versa (see below). This situations can be detected on the `diagnose` subroutine. The main purpose of `diagnose` is "finding where the `out_specs` are" while our solution finds and actually checks "the nodes where all the dependancies are `in_spec`". Those two statements are equivalent as `out_spec` nodes will have all the dependancies `in_spec` but we actually check them where as the original implementation relies on the fact that the user defined a good comparison models and thresholds to properly classify nodes `out_spec`.

                               (check_data result: 🟢in_spec, 🔴bad_data, 🟣out_spec)
        🟢---->🔴---->🔴---->🔴                 🟢---->🟣---->🟣---->🔴            🟢---->🔴---->🟣---->🔴 

On the solution proposed for `diagnose` we will keep both implementations and the user can choose which one to use with the flag `safe_check`. Running diagnose with `safe_check=True` will not only rely on the `check_data` node classification but also check the dependancies are `in_spec`, therefore the running time and the resources to perform this extra checks will be higher.

A more precise description of the errors to handle can be found in [Notion](https://www.notion.so/qilimanjaro/Automatic-Calibration-53cef01488db4a44a7ecf0f475e08084).

# SOLUTION TO ISAAC ERRORS, DESCRIPTION:

We add two flags to `run_automatic_calibration()`, each one solving one of the above cases:
- `force_maintain`
- `safe_diagnose`

both of this options might make some process slower, but if we are in no rush, they should definitely be used... that why we pass them as a flag, in case of needing a quick calibration with a graph that you know that works okei normally, you might want to skip them in the future.
----------

# GUILLE's TODO's for closing this PR:
### Graph error handle TODOs:
- [ ] Finish tests for error handling workflow!
- [ ] check you have modified all the documentation, with the new workflow, no errors!
-- Specially check the docu on bad behaviours, because this solves one of those I think 
- [x] 🔥🔥 Test that solves the error of not having `out_of_spec`
- [x] 🔥🔥 Test that applies the force_maintain correctly in the end always!

### Lab test, TODOs:
- [ ] 🔥🔥 Check that more complex workflow logics, work correctly!
- [ ] 🔥🔥 Check last changes from `lab_test` are explained in docu and tested!
-- for example the check_state changes
-- and look at source code for more!

### LAST TODO:
- [ ] 🔥🔥🔥 Pillar una runcard calibrada > cambiar uns parametres > veure que ens calibra aquest canvi 👷


GuillermoAbadLopez:
Good luck with this PR @IsaacLA12 ! Try to not make it +1000 lines if possible hehe, if you delete the function `_get_forced_maintain_condition`, which I think you should, for this iteration at least. 

(save it in your local as a dummy file or something, if you want)

Then the PR won't be that huge, but it will surely contain lots of subtleties, so make sure to document all those properly and test them. As Albert said, lets try to make a corner case test for each thing we debugged since the first lab test! 🚀

GuillermoAbadLopez:
Again, coverage and documentation is pretty poor for this PR, but we will cover them in a next PR! We need to merge it to deploy it daily, and see the real errors  🚀 

The code from here only affects calibration, and nothing outside of it! So no problems with that either 👍

PR #620, opened 2023-11-20
[QHC-56] Quantum Machines Config

jjmartinezQT:
This branch allows to translate a runcard into Quantum Machines config file which is needed to use the Quantum Machines Manager and run QUA programs.


tests/experiment/test_base_experiment.py
```diff
@@ -349,7 +349,7 @@ def test_set_reset_true_with_connected_device(
         exp.platform.connect()
         exp.platform.disconnect()
         mock_reset.assert_called()
-        assert mock_reset.call_count == 14
+        assert mock_reset.call_count == 12
```
jjmartinezQT, line 352: What is the reason for this change? Has anything changed so that this test method needs to be readjusted?

jjmartinezQT (commented):
One important question written down as we have had issues with that test in the past.

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -0,0 +1,354 @@
...
+
+    name = InstrumentName.QUANTUM_MACHINES_CLUSTER
+
+    @dataclass
+    class QuantumMachinesClusterSettings(Instrument.InstrumentSettings):
+        """Settings for Quantum Machines Cluster instrument.
+
+        Args:
+            address (str): IP address to connect to the Quantum Machines instruments.
+            cluster (str): Name of the cluster to connect to.
```
jjmartinezQT, line 55:
```suggestion
            cluster (str): Name of the cluster to connect to.
            run_octave_calibration (bool): Flag to indicate to run the octave calibration
```

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -0,0 +1,354 @@
...
+            num_controllers (int): Number of controllers (instruments) in the quantum machines stack.
```
jjmartinezQT, line 56: I believe we not longer need the number of controllers here, do we?

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -0,0 +1,354 @@
...
+        cluster: str
+        run_octave_calibration: bool
+        octaves: list[dict[str, Any]]
+        controllers: list[dict[str, Any]]
+        elements: list[dict[str, Any]]
+
+        def to_qua_config(self) -> Dict[str, Any]:
+            """Creates the Quantum Machines QUA config dictionary.
+
+            Creates, an instance of a dictionary in the format that QuantumMachines expects the config dictionary to be.
```
jjmartinezQT, line 72: We need to expand on this docstring to include mixers and octaves. Maybe modify slightly that the empty fields are the ones that will be fulfilled by the QProgram dictionary configuration.

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -0,0 +1,354 @@
...
+
+    This class allows Qililab control and communication with an instance of the
+    Quantum Machines Manager, which is the central class to interact with Quantum Machines instruments.
+    The manager is responsible, through the use of QUA sequences, of setting the Quantum Machines instruments
+    in the right manner for Quantum Control.
+
+    Args:
+        name (InstrumentName): Name of the Instrument.
+        device (QMMDriver): Instance of the Quantum Machines Manager Driver class.
+        settings (QMMSettings): Settings of the instrument.
```
jjmartinezQT, line 44: The following attributes are missing in the docstring: 
    
    qmm: QuantumMachinesManager
    qm: QuantumMachine
    config: DictQuaConfig
    octave_config: QmOctaveConfig | None = None

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -0,0 +1,354 @@
...
+        """Sets initial instrument settings.
+
+        Creates an instance of the Qilililab Quantum Machines Manager, and sets the configuration dictionary.
+        """
+        if self.settings.octaves:
+            self.octave_config = QmOctaveConfig()
+            self.octave_config.set_calibration_db(os.getcwd())
+            for octave in self.settings.octaves:
+                self.octave_config.add_device_info(octave["name"], self.settings.address, octave["port"])
+        self.qmm = QuantumMachinesManager(
```
jjmartinezQT, line 258: This line is complaining locally for me saying that there is an unexpected keyword argument 'cluster_name' in the constructor call. I have checked the QuantumMachinesManager and indeed it does not have that attribute.

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -0,0 +1,354 @@
...
+        self.config = merge_dictionaries(dict(self.config), compilation_config)
+        # If we are already connected, reopen the connection with the new configuration
+        if self._is_connected_to_qm():
+            self.qm = self.qmm.open_qm(config=self.config, close_other_machines=True)
+
+    def run_octave_calibration(self):
+        """Run calibration procedure for the buses with octaves, if any."""
+        elements = [element for element in self.config["elements"] if "RF_inputs" in self.config["elements"][element]]
+        for element in elements:
+            self.qm.calibrate_element(element)
```
jjmartinezQT, line 294: In here it does complain that the instance of QuantumMachine does not have the calibrate_element attribute. However this calibration was worked in the lab, wasn't it?

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
fedonman: Good catch, you have to update `qm_qua` to latest `1.1.6`. I will update the `requirements.txt`.

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
fedonman: Same here, you have to update to `1.1.6`.

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
fedonman: Should we make everything private? None of these should be accessed by the user directly.

fedonman: Maybe the `config` could be exposed by a get-only property.

jjmartinezQT: Those both are good ideas.

fedonman: Done. I didn't create the config property because it is not needed anywhere.

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
fedonman: Done

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
fedonman: Done

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
fedonman: Done

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
fedonman: done

fedonman (approved):
Approved by Juanjo disguised as me.

PR #621, opened 2023-11-20
Check that circuit execution time is shorter than `repetition_duration`

visagim:
As title says

### DIFF
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -2,6 +2,9 @@
 
 ### New features since last release
 
+- Raise an error is a user requests to run a circuit that is longer than `repetition_duration`
+  [#621](https://github.com/qilimanjaro-tech/qililab/pull/621)
+
 - Add parameter to the SLURM magic method to configure the time limit of a job.
   [#608](https://github.com/qilimanjaro-tech/qililab/pull/608)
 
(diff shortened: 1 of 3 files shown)


src/qililab/platform/platform.py
```diff
@@ -655,6 +658,11 @@ def compile(self, program: PulseSchedule | Circuit, num_avg: int, repetition_dur
 
         programs = {}
         for pulse_bus_schedule in pulse_schedule.elements:
+            if len(pulse_bus_schedule.timeline) != 0:  # can't do only one conditional if list is empty
```
AlbertMitjans, line 661: Will the list ever be empty??

AlbertMitjans: If so, shouldn't we list an error when it is empty?

visagim: it's empty for the flux schedules if there are no flux pulses. I believe we haven't fixed that yet.

docs/releases/changelog-dev.md
```diff
@@ -2,6 +2,9 @@
 
 ### New features since last release
 
+- Raise an error is a user requests to run a circuit that is longer than `repetition_duration`
```
AlbertMitjans, line 5:
```suggestion
- Raise an error if a user requests to run a circuit that is longer than `repetition_duration`
```

AlbertMitjans (approved):
Good job! Just one small suggestion.

PR #624, opened 2023-11-22
TAS-535 Reduce time in slurm tests

visagim:
reduced `ql.slurm.num_files_to_keep` in test and reduced `sleep` time


AlbertMitjans (approved):
Thanks!

PR #628, opened 2023-11-23
Hotfix/empty flux schedule

jjmartinezQT:
This PR brings back code that was removed and introduced a regression bus, where all flux buses will get added an empty schedule independently if the bus has an AWG instrument or not. 

Aside of bringing back the old code as it should work, I have found an issue with the alias of the bus in the unittest that will help this unittest failing if this part of the code gets removed again by error. 

In order to avoid the circular dependency, an isort skip line is introduced in the main init file of qililab, and the import of the circuit transpiler module has been moved to the bottom of imports to ensure instruments have been fully imported by the time the transpiler imports the AWG instrument class.


fedonman (approved):
:100: :rocket:

PR #651, opened 2023-12-22
[QHC-64]Compilation at platform level

visagim:
- [x] Fix coverage
- [x] Changelog

**Summary**
- Compilation tasks moved from `qblox_module`, `qblox_qrm`, `qblox_qcm` to `pulses.qblox_compiler`
- Unified and removed some non-acquisition related settings from qrm / qcm under qblox_module
- Improved / added some of the qblox modules tests which were missing
- Removed bus related compiler methods


src/qililab/instruments/qblox/qblox_module.py
```diff
@@ -567,28 +404,40 @@ def upload_qpysequence(self, qpysequence: QpySequence, port: str):
...
 
-        This method must be called after the method ``compile``."""
-        if self.nshots is None or self.repetition_duration is None:
-            raise ValueError("Please compile the circuit before uploading it to the device.")
+        This method must be called after the method ``compile`` in the compiler
+        Args:
+            port (str): The port of the sequencer to upload to.
+        """
+        # TODO: run upload sequencer-wise and pass sequence to be uploaded as argument
+        # TODO: pass the sequence to be uploaded here and save to self.sequences here as well instead of at the compiler
```
AlbertMitjans, line 423: Why are we skipping this TODOs?

src/qililab/platform/platform.py
```diff
@@ -574,7 +580,7 @@ def execute_qprogram(self, qprogram: QProgram) -> dict[str, list[Result]]:
         # Reset instrument settings
         for instrument in self.instruments.elements:
             if isinstance(instrument, QbloxModule):
-                instrument.reset_sequences()
+                instrument.clear_cache()  # FIXME: allow cache with QProgram
```
AlbertMitjans, line 583: Why is this a `FIXME`? I see this comment more of a new feature rather than a bug

src/qililab/platform/platform.py
```diff
@@ -637,14 +643,11 @@ def execute(
                 instrument.desync_sequencers()
 
         # FIXME: set multiple readout buses
-        if len(results) > 1:
-            logger.error("Only One Readout Bus allowed. Reading only from the first one.")
-        if not results:
-            raise ValueError("There are no readout buses in the platform.")
-
```
AlbertMitjans, line 644: Why are you removing these lines?

AlbertMitjans (commented):
This is a very big PR changing core functionalities of `qililab`. I think we should make 2 people review and approve the PR before merging it.

visagim:
> This is a very big PR changing core functionalities of `qililab`. I think we should make 2 people review and approve the PR before merging it.

Yes, 100% agree with that. I also think testing it for some time on hardware before merging would be a good idea

src/qililab/instruments/qblox/qblox_module.py
```diff
@@ -88,13 +80,14 @@ def __post_init__(self):
...
     device: Pulsar | QcmQrm
     # Cache containing the last compiled pulse schedule for each sequencer
-    _cache: dict[int, PulseBusSchedule] = {}
+    cache: dict[int, PulseBusSchedule] = {}
 
     def __init__(self, settings: dict):
         # The sequences dictionary contains all the compiled sequences for each sequencer and a flag indicating whether
         # the sequence has been uploaded or not
-        self.sequences: dict[int, tuple[QpySequence, bool]] = {}  # {sequencer_idx: (program, True), ...}
+        self.sequences: dict[int, QpySequence] = {}  # {sequencer_idx: (program, True), ...}
```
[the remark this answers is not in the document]
visagim: sequences and cache are now handled at the compiler. From there we can check if a given sequence or pulse_bus_schedule is to be reuploaded or not, so there is no need to have the boolean flag.
For repeating sequences there are different scenarios considered, which are different for the qrm and qcm since the qcm always has a 1 to one bus-sequencer correspondence but the qrm can have a single bus pointing to different sequencers.

Consider the second iteration of a software loop i.e. a circuit has been run and executed, sequences are saved to memory (qblox_module.cache and qblox_module.sequences)
case QCM:
1. If a sequence is uploaded to a new bus, don't do anything new (compile sequence and upload to bus)
2. If a sequence is uploaded to a previously used bus, overwrite the previous sequence (qblox_module.sequences[seq_idx] = program) - this is equivalent to 1
3. If nothing is to be compiled to a previously cached bus / sequencer. The cache for this bus will still contain the previously run sequences but nothing will happen since the execution at platform will only call `run` on the buses involved in the current list of `pulse_bus_schedules`
case QRM:
The only difference is the behaviour at 3. Because the QRM belongs to a single bus, we have to check that the pulse_bus_schedules in the circuit to be run are the same as the ones in cache

3. If the same sequencer is run in consecutive iterations, the first sequence will be overwritten by the new one like in 2. At a middle point in execution (inside the `compiler` method), `qblox_module.sequences` and `qblox_module.cache` will contain the union of sequencers to run in the current execution and sequencers that run in the previous one but are not to run in the current one. Then we remove the ones with sequencers not found in the `pulse_bus_schedules` belonging to the current circuit, see lines
```python
        # check for sequences which where not in the PulseSchedule but are uploaded and erase them from cache
        # this is only needed for the qrm since it has multiple sequencers for the same bus
        missing_seq_ids = [
            cached_seq_id
            for cached_seq_id in qrm.cache.keys()
            if cached_seq_id not in (sequencer.identifier for sequencer, _ in sequencer_qrm_bus_schedules)
        ]
        # pop from cache. qblox_module takes care of popping from sequences
        for seq_id in missing_seq_ids:
            _ = qrm.cache.pop(seq_id)
            _ = qrm.sequences.pop(seq_id)
```
in the compiler

src/qililab/platform/platform.py
visagim: Fair enough

src/qililab/platform/platform.py
visagim: It was easier to implement the logic in the compiler assuming only one qrm is present, which can be changed to multiple qrms but would be out of the scope of this pr. However since I am adding this one qrm restriction in the compiler:
```python
        if (
            sum(qblox_module.name in self.readout_modules for qblox_module in self.qblox_modules) > 1
        ):  # FIXME: only one qrm supported
            raise NotImplementedError(
                f"Only one readout module can be connected, found {sum(qblox_module.name in self.readout_modules for qblox_module in self.qblox_modules)} instead"
            )
```
we dont need to raise an error here, since it will be raised before (at the compiler)

visagim: Note that there is also a separate error being raised at the very beginning of compilation if no QRM is found:
```python
        if all(
            qblox.name not in self.readout_modules for qblox in self.qblox_modules
        ):  # Raise error if qrm is not found
            raise ValueError(
                f"At least one QRM should be connected. Found {sum(qblox.name in self.readout_modules for qblox in self.qblox_modules)} QRM modules"
            )
```

src/qililab/instruments/qblox/qblox_module.py
```diff
@@ -88,13 +80,14 @@ def __post_init__(self):
...
         # TODO: Set this attribute during initialization of the instrument
+        # TODO: delete useless parameters nshots num bins since they are managed at compilation
```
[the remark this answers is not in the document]
visagim: I could do it but then I would have to change more tests and base classes. Since having them there until we clean them doesn't crash anything I prefered to do that in a separate pr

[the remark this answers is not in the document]
visagim: But I can do it here too otherwise

src/qililab/instruments/qblox/qblox_module.py
```diff
@@ -137,168 +121,16 @@ def module_type(self):
...
-        self, loop: Loop, bin_index: Register | int, sequencer_id: int, weight_regs: tuple[Register, Register]
-    ):
-        """Append an acquire instruction to the loop."""
-
     def start_sequencer(self, port: str):
         """Start sequencer and execute the uploaded instructions."""
         sequencers = self.get_sequencers_from_chip_port_id(chip_port_id=port)
         for sequencer in sequencers:
             if sequencer.identifier in self.sequences:
+                # TODO: skip sequencer here if it is not in cache, instead of poping the sequence at upload
```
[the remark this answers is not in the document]
visagim: this is because I wanted to handle cache in the compiler, and sequences here in qblox_module (do decouple compiler and sequencer as much as possible). But in the end we agreed (in the code discussion before Christmas) to handle both at compiler

src/qililab/pulse/qblox_compiler.py
```diff
@@ -0,0 +1,386 @@
...
+        Returns:
+            dict[AWGQbloxSequencer, PulseBusSchedule]: dictionary of the pulse schedule corresponding to a given sequencer
+        """
+        qrm_sequencers = [
+            sequencer
+            for instrument in qblox_instruments
+            for sequencer in instrument.awg_sequencers
+            if instrument.name in self.readout_modules
+        ]
+        feedline_port = qrm_sequencers[0].chip_port_id  # FIXME: only one chip port id supported for qrm
```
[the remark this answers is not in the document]
visagim: This is related to all the buses problems for feedline input currently pointing to 5 diferent sequencers. This FIXME is related to adding support for more than one QRM in the compiler (I'll add a task for that since we are supporting this in the iqm chip)

fedonman (approved):
💯 💯

PR #652, opened 2023-12-22
moving demos to docs

visagim:


docs/demos/rabi.ipynb
```diff
@@ -571,6 +571,7 @@
     "    fig.suptitle(title_label)\n",
     "    return fig, axes\n",
     "\n",
+    "\n",
```
AlbertMitjans, line 574: Was this on purpose?

AlbertMitjans (approved):
Thanks!!

docs/demos/rabi.ipynb
visagim: nope

visagim: apparently it was nbqa-black doing some funny sfuf

fedonman (approved):

PR #653, opened 2023-12-27
including reset parameter in the instruments controller to be saved i…

jjmartinezQT:
This PR fix the bug encountered that instruments controllers won't save the reset attribute in the runcard saving method.


AlbertMitjans:
@jjmartinezQT make sure to tag only one reviewer per PR! If we tag multiple people, it might happen that nobody reviews the PR because they are expecting someone else to review it 😆

src/qililab/instrument_controllers/instrument_controller.py
```diff
@@ -254,9 +254,12 @@ def __str__(self):
...
     def to_dict(self):
         """Return a dict representation of the InstrumentReference class."""
-        return {
+        controller_dict = {
             RUNCARD.NAME: self.name.value,
             RUNCARD.ALIAS: self.alias,
             INSTRUMENTCONTROLLER.CONNECTION: self.connection.to_dict(),
             INSTRUMENTCONTROLLER.MODULES: [module.to_dict() for module in self.settings.modules],
         }
+        if not self.settings.reset:
```
GuillermoAbadLopez, line 263: You are only saving the reset, when its `False`, is that because the default is `True` and it's not necessary in such case?

GuillermoAbadLopez (commented):
Looking nice, I've left a small comment, just to check, but I'll leave the rest to Albert as we said 👍

src/qililab/instrument_controllers/instrument_controller.py
jjmartinezQT: That is the reason yes. First to not overload the runcard with default parameters, as we are already not doing as we are not saving all the parameters but only those that might change, and also not to have to change many unittests to fix this bug.

AlbertMitjans:
> Looking nice, I've left a small comment, just to check, but I'll leave the rest to Albert as we said 👍

The PR looks good to me, but I prefer not to approve any PR given that it's been a while since I look at qililab's code 😆

fedonman (approved):
Great job! 💯 Everything seem ok for me! 🚀

PR #659, opened 2024-01-08
[QHC-21] Make QProgram serializable

fedonman:
Added `DictSerializable` protocol and `from_dict` utility function to enable (de)serialization (from)to dictionary for any class.


jjmartinezQT (approved):
Good work, very neat.

jjmartinezQT (commented):
Good catch.

PR #663, opened 2024-01-15
[DIRTY-FIX] Add `wait_time` to play operation for Qblox

fedonman:
Since `time_of_flight` is not yet added to runcard settings for Qblox, and measurement is not unified under one operation `measure`, users need a way to define a `time_of_flight` wait time between readout pulse and acquisition. This PR adds this functionality:

```
qp.play(bus="readout_bus", waveform=readout_pair, wait_time=4)
```

### DIFF
--- a/src/qililab/qprogram/qblox_compiler.py
+++ b/src/qililab/qprogram/qblox_compiler.py
@@ -454,7 +454,8 @@ class QbloxCompiler:  # pylint: disable=too-few-public-methods
             index_I, index_Q, duration = self._append_to_waveforms_of_bus(
                 bus=element.bus, waveform_I=waveform_I, waveform_Q=waveform_Q
             )
-            duration = element.wait_time or duration  # TODO: Change this in clean fix
+            if element.wait_time is not None:  # TODO: Change this in clean fix
+                duration = element.wait_time
             convert = QbloxCompiler._convert_value(element)
             duration = convert(duration)
             self._buses[element.bus].static_duration += duration
--- a/src/qililab/qprogram/qblox_compiler.py
+++ b/src/qililab/qprogram/qblox_compiler.py
@@ -454,6 +454,7 @@ class QbloxCompiler:  # pylint: disable=too-few-public-methods
             index_I, index_Q, duration = self._append_to_waveforms_of_bus(
                 bus=element.bus, waveform_I=waveform_I, waveform_Q=waveform_Q
             )
+            duration = element.wait_time or duration  # TODO: Change this in clean fix
             convert = QbloxCompiler._convert_value(element)
             duration = convert(duration)
             self._buses[element.bus].static_duration += duration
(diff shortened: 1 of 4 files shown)


src/qililab/qprogram/qblox_compiler.py
```diff
@@ -454,6 +454,7 @@ def _handle_play(self, element: Play):
             index_I, index_Q, duration = self._append_to_waveforms_of_bus(
                 bus=element.bus, waveform_I=waveform_I, waveform_Q=waveform_Q
             )
+            duration = element.wait_time or duration  # TODO: Change this in clean fix
```
GuillermoAbadLopez, line 457: This `or` operator between objects can be sketchy, so I'm just gonna write this here to help we are choosing the good one we wanna prioritize in each case :) 

Def: `variable = x or y` gives `x`, if it evaluates to `True`, otherwise `y`. 

And in our case (`int | None`) it will evaluate to `False` when we have `None` or a `0`

So in this case, if `element.wait_time` is not `None` or a `0` we will keep it, without checking the previous value of `duration`, and if `element.wait_time` is `None` or a `0`, then we will keep whatever `duration` was previously, doesn't matter what it was, right?

GuillermoAbadLopez (commented):
I don't know what we prefer to keep in case of conflicts between both variables `duration` and `element.wait_time`, so I just added a comment to double check the pure logic, if it's what you had in mind perfect!

src/qililab/qprogram/qblox_compiler.py
fedonman: `A = B or C` is the same as writing

```
if B is not None:
   A = B
else:
   B = C
```

In our case, if the user has provided `element.wait_time` I want to use that as duration, otherwise keep duration as it is.

fedonman: You are right about the case `wait_time == 0`. I will change the syntax to normal.

GuillermoAbadLopez: Thats not exactly what I read online, what I read, is more like: 
```
if B not in [None, 0, 0.0, {}, [], ...]:
   A = B
else:
   A = C
```

So for the 0 case, it wouldn't overwrite it, sorry I had edited the comment just when you were answering, since my point wasn't clear enough I think 😬

GuillermoAbadLopez: Okei I just saw your second answer, cool! 🚀 

Sorry for the late edit 🙌

jjmartinezQT (approved):
Good to go now, good work.

GuillermoAbadLopez (approved):
With the new change, the overwrite logic looks good to me!

(Mind, that I haven't checked if the attribute has been added to all the necessary places, or higher qililab stuff, only checked the overwrite logic 🙌 )

PR #665, opened 2024-01-15
add changes added to main

visagim:


src/qililab/platform/platform.py
```diff
@@ -397,6 +397,24 @@ def get_element(self, alias: str):
             element = self.chip.get_node_from_alias(alias=alias)
         return element
 
+    def get_qrm_ch_id_from_qubit(self, alias: str, qubit_index: int) -> int:
+        """Finds a sequencer id for a given qubit given a bus alias
```
jjmartinezQT, line 401: Could we document this one a bit better? What I mean is that we should include the reasons why this is done only for QRM, what motivated it, maybe the ticket id too, so in the future we avoid accidents like the time we deleted another check for this kind of Qblox Instruments that re-introduced an old bug.

jjmartinezQT (changes requested):

jjmartinezQT (approved):
Approved, ready to merge whenever tests passes.

PR #667, opened 2024-01-16
 [QHC-169] Add `_handle_long_wait` to qprogram qblox compiler

visagim:
Closes QHC-188


visagim:
I removed I line I had previously added raising an error if the duration of a dynamic wait was longer than qblox `INST_MAX_WAIT` because the value of the wait is not known until allocated after compilation. I think this is ok however, since the `wait` instruction from qpysequence already raises an error if a duration longer than `INST_MAX_WAIT` is allocated (so an error expliciting what went wrong is already in place)

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -341,15 +342,34 @@ def _handle_wait(self, element: Wait):
         if isinstance(element.duration, Variable):
             duration = self._buses[element.bus].variable_to_register[element.duration]
             self._buses[element.bus].dynamic_durations.append(element.duration)
+            self._buses[element.bus].qpy_block_stack[-1].append_component(
+                component=QPyInstructions.Wait(wait_time=duration)
+            )
```
fedonman, line 347: What if the variable goes over INST_MAX_WAIT? Shouldn't we handle it the same way?

```
with qp.for_loop(variable=duration, start=100, stop=2 * INST_MAX_WAIT, step=1000):
    qp.wait(duration=duration)
```

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -341,15 +342,34 @@ def _handle_wait(self, element: Wait):
...
+        """Wait for longer than QBLOX INST_MAX_WAIT by looping over wait instructions
+
+        Args:
+            element (Wait): wait element
+            duration (int): duration to wait in ns
+        """
+        for _ in range(duration // INST_MAX_WAIT):
+            self._buses[bus].qpy_block_stack[-1].append_component(
+                component=QPyInstructions.Wait(wait_time=INST_MAX_WAIT)
+            )
```
fedonman, line 371: Since this method does not actually handle long waits by itself (since you also need the extra lines at 356-458) I would prefer to have it directly as code inside the `_handle_wait`. It will be also be clearer since it is not a "handle", it's not added to the `_handlers` dictionary.

src/qililab/qprogram/qblox_compiler.py
visagim: Yes I mention this in the comment above

> I removed I line I had previously added raising an error if the duration of a dynamic wait was longer than qblox INST_MAX_WAIT because the value of the wait is not known until allocated after compilation. I think this is ok however, since the wait instruction from qpysequence already raises an error if a duration longer than INST_MAX_WAIT is allocated (so an error expliciting what went wrong is already in place)

QPysequence already raises an error for this. Would you do it otherwise?

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -380,8 +400,12 @@ def __handle_static_sync(self, buses: set[str]):
...
             duration_diff = max_duration - self._buses[bus].static_duration
             if duration_diff > 0:
+                # loop over wait instructions if static duration is longer than allowed qblox max wait time of 2**16 -4
+                if duration_diff > INST_MAX_WAIT:
+                    self._handle_long_wait(bus=bus, duration=duration_diff)
+                # add the remaining wait time (or all of it if the above conditional is false)
                 self._buses[bus].qpy_block_stack[-1].append_component(
-                    component=QPyInstructions.Wait(wait_time=duration_diff)
+                    component=QPyInstructions.Wait(wait_time=duration_diff % INST_MAX_WAIT)
                 )
```
fedonman, line 409: Could you add all of these under one method? Because now we have half of the logic in `_handle_long_wait` and half in the lines 407-409. And many lines are repeated anyway. We could have everything in one method `__add_wait_instructions`.

src/qililab/qprogram/qblox_compiler.py
fedonman: Oh I didn't see the comment.

Well, in general, if we can catch an error in Qililab level it's better. Since we can provide our own meaningful error etc. And user's shouldn't actually know or care what is going on behind the scenes (for example that we use qpysequence).

fedonman:
Add "Closes QHC-188" in the description to link the bug in linear.

fedonman (approved):
Ready to go 🚀

PR #669, opened 2024-01-17
[QHC-174] Add support for more than one QRM on `qililab.pulse.qblox_compiler`

visagim:


visagim:
This adds only the functionality for the compiler to handle more than one QRM, preparing main so we can merge it with `new-qblox-firmware` branch. It does not handle parsing a list of results aquired from multiple readout buses to a single qblox result - that is something that's already being done in `new-qblox-firmware` and for which the code / workflow is the same, independent on the compiler

IsaacLA12 (approved):
Good job! Nice tests 🚀 💯

PR #673, opened 2024-01-19
Linear calibration / Check 2h time_check in calibration problem

GuillermoAbadLopez:


src/qililab/calibration/calibration_controller.py
```diff
@@ -198,7 +198,10 @@ def calibrate_all(self, node: CalibrationNode):
         for n in self._dependencies(node):
             self.calibrate_all(n)
 
-        if node.previous_timestamp is None or self._is_timeout_expired(node.previous_timestamp, 7200.0):
+        # You can skip it from 2h time, but also skip it due to `been_calibrated()`
+        if (
+            node.previous_timestamp is None or self._is_timeout_expired(node.previous_timestamp, 7200.0)
+        ) and not node.been_calibrated:
```
IsaacLA12, line 204: I believe we don't actually need to check anything related to previous timestamps or anything related to if for linear calibration now that we have this boolean to check (nice implementation btw!!). Whenever we decide to run linear calibration we want to run everything, so to avoid repeating calibrations in the same run I think we just have enough checking `not node.been_calibrated`

Also, with that implementation we could have the following undesired case:
We ran 2 consecutive linear calibrations, for the second one the previous timestamps will not be None (as there exist previous executions) but could be that the time between the execution of a node in the first linear calibration and the second is less than 2 hours so even if you have not calibrated it in this run it would get the parameters from the first one and I think this should not happen 🤔

GuillermoAbadLopez: In the PR description, I wrote this, because I know it might came up:

> I didn't remove the 2h check either, because I think it provides a useful feature, such as stoping a calibration, because some problem, and you can run it again, and it starts from where it stopped previously!

GuillermoAbadLopez: And in the case we wanted to do a new calibration from start, which is when this "feature" might give problems, then you would only need to remove the previous executed files, so it starts again from 0!

IsaacLA12: Nvm, I saw your inline comment and this can be a feature to resume your calibration if it was interrupted which can be helpful

IsaacLA12 (approved):
Nice job and fast! 💯 🪨

PR #679, opened 2024-01-30
[QHC-217] Ensure QProgram supports variables in waveforms

fedonman:
Added Domain restrictions to `Drag` pulse, `DragCorrection` waveform and `Gaussian` waveform.

### DIFF
--- a/tests/qprogram/test_qprogram.py
+++ b/tests/qprogram/test_qprogram.py
@@ -1,15 +1,15 @@
 import math
 from collections import deque
+from itertools import product
 
 import numpy as np
 import pytest
 
-from qililab import Domain, IQPair, QProgram, Square
+from qililab import Domain, DragCorrection, DragPair, Gaussian, IQPair, QProgram, Square
 from qililab.qprogram.blocks import Average, Block, ForLoop, InfiniteLoop, Loop, Parallel
 from qililab.qprogram.operations import (
     Acquire,
     Measure,
-    Operation,
     Play,
     ResetPhase,
     SetFrequency,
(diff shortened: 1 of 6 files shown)


tests/qprogram/test_qprogram.py
```diff
@@ -319,14 +319,56 @@ def test_variable_method_raises_error_if_domain_is_not_scalar_and_type_is_set(se
...
+
+        for var in all_types - {frequency}:
+            with pytest.raises(ValueError):
+                qp.set_frequency(bus="drive", frequency=var)
+
+        for var in all_types - {phase}:
+            with pytest.raises(ValueError):
+                qp.set_phase(bus="drive", phase=var)
+
+        for var in all_types - {voltage}:
```
jjmartinezQT, line 338:
```suggestion
       for var in (all_types - {voltage}):
            with pytest.raises(ValueError):
                qp.set_offset(bus="drive", offset_path0=var, offset_path1=var)
                qp.set_gain(bus="drive", gain=var)
```

jjmartinezQT: You could optimise testing with something like this.

jjmartinezQT: In this case it is not much gaining and the comment might seem a bit picky, but for future cases just to know this will test all the same.

jjmartinezQT (changes requested):
Minor code changes when testing

tests/qprogram/test_qprogram.py
fedonman: I hate to break it to you, but this wouldn't actually work. 😅 Because `qp.set_offset` would raise the `ValueError`, thus ending execution, and `qp.set_gain` would never be tested. (I had the same thought as you 🥂 )

jjmartinezQT: Fair enough, I had that conviction, but as you say, this way we know 100% sure. Approved then.

jjmartinezQT (approved):

PR #680, opened 2024-02-01
Merge `new-qblox-firmware` into main

visagim:


fedonman (changes requested):
Awesome job! 💯 Add changelog entries for all changes happening in this PR. I added comments on some changes I found that are not relevant with the PR title.

src/qililab/pulse/pulse_shape/flat_top.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def to_dict(self) -> dict:
+        """Returns dictionary representation of the Rectangular object/shape.
+
+        Returns:
+            dict: Dictionary representation of the pulse shape including its name.
+        """
+        return {
+            "name": self.name.value,
+        }
```
fedonman, line 91: Add a changelog entry for this.

src/qililab/platform/platform.py
```diff
@@ -714,7 +713,7 @@ def execute(
         repetition_duration: int,
         num_bins: int = 1,
         queue: Queue | None = None,
-    ) -> Result:
+    ) -> Result | QbloxResult:
```
fedonman, line 716: Why do we change this?

src/qililab/platform/platform.py
```diff
@@ -768,7 +767,15 @@ def execute(
...
-        # FIXME: set multiple readout buses
+        # Flatten results if more than one readout bus was used for a qblox module
+        if len(results) > 1:
+            return QbloxResult(
+                integration_lengths=[length for result in results for length in result.integration_lengths],  # type: ignore [attr-defined]
+                qblox_raw_results=[raw_result for result in results for raw_result in result.qblox_raw_results],  # type: ignore [attr-defined]
+            )
+        if not results:
+            raise ValueError("There are no readout buses in the platform.")
+
```
fedonman, line 778: Why do we change this?

src/qililab/waveforms/flat_top.py
```diff
@@ -0,0 +1,66 @@
...
+    Args:
+        duration (int): Duration of the pulse (ns).
+        amplitude (float): Maximum amplitude of the pulse.
+        gaussian (float, optional): Sigma number of the gaussian pulse shape. Defaults to 0.5.
+        buffer (float, optional): Buffer of the waveform. Defaults to 3.0.
+    """
+
+    @requires_domain("amplitude", Domain.Voltage)
+    @requires_domain("duration", Domain.Time)
+    def __init__(self, amplitude: float, duration: int, gaussian: float = 0.5, buffer: float = 3.0):
```
fedonman, line 37: add `requires_domain(Domain.Scalar)` for `gaussian` and `buffer` parameters. Also is the `gaussian` the same as `num_sigmas` parameter in the `Gaussian` waveform? If yes, maybe we should rename to be consistent?

src/qililab/waveforms/flat_top.py
```diff
@@ -0,0 +1,66 @@
...
+        super().__init__()
+        self.amplitude = amplitude
+        self.duration = duration
+        self.gaussian = gaussian
+        self.buffer = buffer
+
+    def envelope(self, resolution: int = 1) -> np.ndarray:
+        """Constant amplitude envelope.
```
fedonman, line 45: Change docstring

src/qililab/waveforms/flat_top.py
fedonman: Also, why did we choose these default values? Is it a special case?

src/qililab/platform/platform.py
visagim: this is to merge changes from `main` which are not in this branch, so it's not a change from `new-qblox-firmware`. This change was introduced to support post-processing results in platform when we re-order them at `self._order_result`

fedonman: I am confused. main branch does not contain `QbloxResult`. 

image

visagim: ok apparently I was wrong, these changes are from https://github.com/qilimanjaro-tech/qililab/pull/656, which we did not implement in `main`

src/qililab/platform/platform.py
visagim: same as above https://github.com/qilimanjaro-tech/qililab/pull/656 I'll add a changelog entry for this as well

src/qililab/waveforms/flat_top.py
visagim: yes, these were set by @paulqili

visagim: `gaussian` is not the same as `num_sigmas` though it is similar. According to Paul it regulates how "gaussian" the flat top distribution is. But since it has a different nature I'd keep the name as is.

fedonman (approved):
Ready to go! :rocket:

PR #691, opened 2024-02-19
[QHC-238] Quick fix to qprogram multiple bus acquisitions

visagim:
Temporary fix, assumes 2 things:

- each bus points to only 1 qrm.sequencer
- assumes that the name of the bus contains at least 1 integer (qubit number). If it contains more than one integer, it assumes that the qubit number is the first one, eg `readout_q0_bus` gets `qubit=0`


fedonman (changes requested):
Sorry to ask this after having the implementation, but could we consider a more clean approach, so we don't need to have assumptions about namings? There is an easy and quick solution that should work with any bus/acquisition name:

If we add a `port` parameter in the `get_qprogram_acquisitions()` method of `ReadoutSystemControl` we can move the information of the specific sequencer to acquire from down to the instrument. In the same way we do in the `upload()` method:

```
# Bus calls the system control with port=self.port
self.system_control.upload_qpysequence(qpysequence=qpysequence, port=self.port)

# SystemControl passes the port to the instrument method
def upload_qpysequence(self, qpysequence: QpySequence, port: str):
        """Uploads the qpysequence into the instrument."""
        for instrument in self.instruments:
            if isinstance(instrument, AWG):
                instrument.upload_qpysequence(qpysequence=qpysequence, port=port)
                return

        raise AttributeError("The system control doesn't have any AWG to upload a qpysequence.")
        
# QbloxModule finds the correct sequencer using the port
def upload_qpysequence(self, qpysequence: QpySequence, port: str):
        for sequencer in sequencers:
            ...
```

fedonman:
> Sorry to ask this after having the implementation, but could we consider a more clean approach, so we don't need to have assumptions about namings? There is an easy and quick solution that should work with any bus/acquisition name:
> 
> If we add a `port` parameter in the `get_qprogram_acquisitions()` method of `ReadoutSystemControl` we can move the information of the specific sequencer to acquire from down to the instrument. In the same way we do in the `upload()` method:
> 
> ```
> # Bus calls the system control with port=self.port
> self.system_control.upload_qpysequence(qpysequence=qpysequence, port=self.port)
> 
> # SystemControl passes the port to the instrument method
> def upload_qpysequence(self, qpysequence: QpySequence, port: str):
>         """Uploads the qpysequence into the instrument."""
>         for instrument in self.instruments:
>             if isinstance(instrument, AWG):
>                 instrument.upload_qpysequence(qpysequence=qpysequence, port=port)
>                 return
> 
>         raise AttributeError("The system control doesn't have any AWG to upload a qpysequence.")
>         
> # QbloxModule finds the correct sequencer using the port
> def upload_qpysequence(self, qpysequence: QpySequence, port: str):
>         sequencers = self.get_sequencers_from_chip_port_id(chip_port_id=port)
>         for sequencer in sequencers:
>             ...
> ```

I am asking this, because I think it's super easy to implement. If you think it will complicate things we can move on with the current one. I just want to avoid having restrictions in namings. Because many examples and demos just use a `drive_bus` / `readout_bus`.

jjmartinezQT (approved):
I agree with the method implemented to avoid naming conventions, which would have been likely to introduce multiple errors when working with different runcards. The whole thing has passed the unittests and the implementation is simple to avoid the issue, good job.

fedonman (approved):
images

PR #695, opened 2024-03-05
[QHC-272] Change degrees to radians for QProgram's `set_phase` operation

fedonman:


jordivallsq (approved):
The unit conversion is straight forward and I see no issues.

PR #697, opened 2024-03-06
Extracted sync from upload_qpysequence into an independent line

jordivallsq:
Moved the sync outside of upload_qpysequence in Qblox_module into an independent function sync_sequencers located in the same .py file mirroring the function desync_sequencers and added 3 lines in platform.py to also mirror the desync and sync every time the execution is ran.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -20,4 +20,4 @@ disable=
     too-many-ancestors,
     missing-module-docstring,
     too-many-arguments,
-    R0912
+    too-many-branches
--- a/.pylintrc
+++ b/.pylintrc
@@ -19,5 +19,5 @@ disable=
     duplicate-code,
     too-many-ancestors,
     missing-module-docstring,
-    too-many-arguments
+    too-many-arguments,
     R0912
--- a/.pylintrc
+++ b/.pylintrc
@@ -20,3 +20,4 @@ disable=
     too-many-ancestors,
     missing-module-docstring,
     too-many-arguments
+    R0912
--- a/src/qililab/instruments/qblox/qblox_module.py
+++ b/src/qililab/instruments/qblox/qblox_module.py
@@ -116,7 +116,6 @@ class QbloxModule(AWG):
         sequencers = self.get_sequencers_from_chip_port_id(chip_port_id=port)
         for sequencer in sequencers:
             self.device.sequencers[sequencer.identifier].sync_en(True)
-        return None
 
     # TODO: merge both desyncs or decide if desync_sequencers is enough
     def desync_by_port(self, port: str) -> None:
@@ -124,7 +123,6 @@ class QbloxModule(AWG):
         sequencers = self.get_sequencers_from_chip_port_id(chip_port_id=port)
         for sequencer in sequencers:
             self.device.sequencers[sequencer.identifier].sync_en(False)
-        return None
 
     def desync_sequencers(self) -> None:
         """Desyncs all sequencers."""
--- a/src/qililab/instruments/qblox/qblox_module.py
+++ b/src/qililab/instruments/qblox/qblox_module.py
@@ -70,9 +70,9 @@ class QbloxModule(AWG):
                 )
 
             self.awg_sequencers = [
-                AWGQbloxSequencer(**sequencer)
-                if isinstance(sequencer, dict)
-                else sequencer  # pylint: disable=not-a-mapping
+                (
+                    AWGQbloxSequencer(**sequencer) if isinstance(sequencer, dict) else sequencer
+                )  # pylint: disable=not-a-mapping
                 for sequencer in self.awg_sequencers
             ]
             super().__post_init__()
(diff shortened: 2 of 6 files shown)


jordivallsq:
The code has been checked with 2 different pulses from different sequencers in a software loop.
This is the n step of a software loop where we can see how both pulses are in sync:
imagen

The sequencers used are:
`                wait_sync        4              

main:
                move             200, R0        
avg_0:
                wait             2000           
                wait             4              
                play             0, 1, 4        
                acquire_weighed  0, 0, 0, 0, 2000
                wait             200            
                loop             R0, @avg_0     
                stop                            


[qililab] [0.23.3|INFO|2024-03-06 10:18:25]: Sequence program: 
 setup:
                wait_sync        4              

main:
                move             200, R0        
avg_0:
                play             0, 1, 2000     
                wait             2208           
                loop             R0, @avg_0     
                stop     `

We can see how the sequencers structure follow the pulses distribution

Using Qililab main we are not able to see one of the colors in the image, only one is seen because they desync

src/qililab/platform/platform.py
```diff
@@ -652,6 +652,11 @@ def _execute_qprogram_with_qblox(
                 buses[bus_alias].upload_qpysequence(qpysequence=sequences[bus_alias])
                 self._qpy_sequence_cache[bus_alias] = sequence_hash
 
+        # sync all instruments
+        for instrument in self.instruments.elements:
+            if isinstance(instrument, QbloxModule):
+                instrument.sync_sequencers()
```
fedonman, line 658: If you do this, you will enable syncing to **all** sequencers present. We only want to enable syncing to the sequencers that are relevant to the specific QProgram.

The de-sync is happening for all sequencers, since you don't really care about it (it doesn't affect anyone if you desync an irrelevant sequencer), but you may change that too so it's more consistent.

jordivallsq: Now I set the sync as true inside the bus loop (`for bus_alias in sequences:`) where the `upload_qpysequence` is located (it is not inside the conditional) with:
```
for instrument in buses[bus_alias].system_control.instruments:
        instrument.sync_sequencers(buses[bus_alias].port)
```
All the required information (port and instruments) comes from the bus

And now the desync is also like this:
```
# Reset instrument settings
for bus_alias in sequences:
    for instrument in buses[bus_alias].system_control.instruments:
            instrument.desync_sequencers(buses[bus_alias].port)
```

The functions changed inside QbloxModule (since these are only rellevant here) are:
```
def sync_sequencers(self, port: str) -> None:
    """Syncs all sequencers."""
    for sequencer in sequencers:
        self.device.sequencers[sequencer.identifier].sync_en(True)

def desync_sequencers(self, port: str) -> None:
    """Syncs all sequencers."""
    for sequencer in sequencers:
        self.device.sequencers[sequencer.identifier].sync_en(False)
```

jordivallsq: I have also tested the syncing

imagen

but the code saving still gives me errors by changing the imports

src/qililab/instruments/qblox/qblox_module.py
```diff
@@ -70,9 +70,9 @@ def __post_init__(self):
                 )
 
             self.awg_sequencers = [
-                AWGQbloxSequencer(**sequencer)
-                if isinstance(sequencer, dict)
-                else sequencer  # pylint: disable=not-a-mapping
+                (
+                    AWGQbloxSequencer(**sequencer) if isinstance(sequencer, dict) else sequencer
+                )  # pylint: disable=not-a-mapping
```
fedonman, line 75: why is this changed? I think it was messed by your formatter before you set it up correctly. Could you revert it back?

src/qililab/instruments/qblox/qblox_module.py
```diff
@@ -111,6 +111,19 @@ def initial_setup(self):
         for idx, offset in enumerate(self.out_offsets):
             self._set_out_offset(output=idx, value=offset)
 
+    def sync_by_port(self, port: str) -> None:
+        """Syncs all sequencers."""
+        sequencers = self.get_sequencers_from_chip_port_id(chip_port_id=port)
+        for sequencer in sequencers:
+            self.device.sequencers[sequencer.identifier].sync_en(True)
+
+    # TODO: merge both desyncs or decide if desync_sequencers is enough
```
fedonman, line 120: comment is not necessary here since we keep and use both methods.

tests/.pylintrc
```diff
@@ -19,4 +19,5 @@ disable=
     too-many-public-methods,
     missing-class-docstring,
     missing-function-docstring,
-    import-error
+    import-error,
+    R0912
```
fedonman, line 23: add the rule by its name, not its id.

.pylintrc
```diff
@@ -19,4 +19,5 @@ disable=
     duplicate-code,
     too-many-ancestors,
     missing-module-docstring,
-    too-many-arguments
+    too-many-arguments,
+    R0912
```
fedonman, line 23: add the rule by its name, not its code

fedonman (changes requested):
Awesome job! 💯 Requested some minor changes

fedonman (approved):
Awesome! 💯

Issue #709, opened 2024-03-27
[BUG] Changing waveform using Qprogram for QM causes crash


OscarQili:
### Expected behavior

I want to run the same Qprogram several times in a row, changing the pulse length between runs. This should run without any issue.

### Actual behavior

The first Qprogram runs without issue. After changing the pulse length from 1000 -> 1004 for instance, it throws an error and crashes.

### Additional information

In this case the Qprogram run is "rabi_single" from experiment_portfolio.qprogram.
I have managed to get it to crash through the same procedure for other Qprograms such as resonator spectroscopy however. 

The notebook used can be found in LabScripts/Personal_folders/Oscar/QM_Qprogram/Qprogram_QM_spectroscopy_new_calibration.ipynb

### Source code

```Shell
# Debugging
from experiment_portfolio.qprogram import rabi_amplitude, rabi_single

qprogram = rabi_single(AVERAGES, int(RESON_DUR), 1000, "QM", RESON_AMP, DRIVE_AMP)

results = platform.execute_qprogram(qprogram, bus_mapping={"readout": readout_bus, "drive": drive_bus}).results

----- change value in cell

# Debugging
from experiment_portfolio.qprogram import rabi_amplitude, rabi_single

qprogram = rabi_single(AVERAGES, int(RESON_DUR), 1004, "QM", RESON_AMP, DRIVE_AMP)

results = platform.execute_qprogram(qprogram, bus_mapping={"readout": readout_bus, "drive": drive_bus}).results
```

### Tracebacks

```Shell
Unclosed connection: Channel('192.168.1.67', 10251, ..., path=None)
2024-03-26 14:05:15,544 - qm - ERROR    - CONFIG ERROR in key "pulses.control_b3869bb5_b3869bb5_1000" [pulses.length] : Pulse 'control_b3869bb5_b3869bb5_1000' length doesn't match waveform 'b3869bb5' length
2024-03-26 14:05:15,545 - qm - ERROR    - CONFIG ERROR in key "pulses.control_b3869bb5_b3869bb5_1000" [pulses.length] : Pulse 'control_b3869bb5_b3869bb5_1000' length doesn't match waveform 'b3869bb5' length
Unclosed connection: Channel('192.168.1.67', 10251, ..., path=None)
Unclosed connection: Channel('192.168.1.67', 10251, ..., path=None)

{
	"name": "OpenQmException",
	"message": "Can not open QM. Please see previous errors",
[...]

PR #713, opened 2024-04-03
[BUG] hotfix to prevent the bug, now the duration is considered while defining the hashtag of the waveform

jordivallsq:
This is a temporal Hotfix with the bugfix, some feedback is required from Hardware to see if this is the real solution

Bug reference: https://github.com/qilimanjaro-tech/qililab/issues/709

### DIFF
--- a/src/qililab/qprogram/quantum_machines_compiler.py
+++ b/src/qililab/qprogram/quantum_machines_compiler.py
@@ -526,7 +526,6 @@ class QuantumMachinesCompiler:  # pylint: disable=too-many-instance-attributes, 
         attributes = [
             f"{key}: {(QuantumMachinesCompiler.__hash_waveform(value) if isinstance(value, Waveform) else str(value))}"
             for key, value in waveform.__dict__.items()
-            if key != "duration"
         ]
         string_to_hash = f"{waveform.__class__.__name__}({','.join(attributes)})"
         hash_result = hashlib.md5(string_to_hash.encode("utf-8"), usedforsecurity=False)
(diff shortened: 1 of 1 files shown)


jordivallsq:
After further investigation of the bug the conclusions are these:

QM has a configuration dictionary (https://docs.quantum-machines.co/1.1.6/qm-qua-sdk/docs/Introduction/config/) including all the information of pulses, waveforms, weights...

The bug originated from a conflict between the information contained in pulses and waveforms for non-Square pulses, in `pulses` the duration is saved as an integer, and in `waveforms` the full pulse is saved as an array (with the requirement that the shape of the array must be the same as the duration).

The machine saves internally this configuration using a system of labels and to do so we create hashes based on the characteristics of the pulse (i.e. `Gaussian(amplitude: 0.1,duration: 1008,num_sigmas: 4.0)`) but the code ignored the duration (`Gaussian(amplitude: 0.1,num_sigmas: 4.0)`) presumably because it is not needed (even detrimental because is irrelevant information) for Square waveforms. But for non-Square waveforms, it creates a conflict where duration sizes do not match between `waveforms` and `pulses`.

The solution implemented solves the problem on non-Square waveforms leaving Square waveforms with the same functionality.

jjmartinezQT (dismissed):
Fine from the software perspective. Ping me back when hardware has reviewed this hotfix and I will approve. Also please update with main as this one is out of date. Good work.

jordivallsq:
By hardware the test has been test with rabi experiments where the size of the pulse changed in every iteration. The experiment run smoothly without crashes (unlike prior to the fix).

This image is for a drive of 100 ns:
image

and for 400 ns:
image


we can see how we are able to increase the size of a waveform (through software loops) without any crash

This experiment has been done in collaboration with Oscar

jordivallsq:
I am adding @fedonman as a reviewer because Juanjo is not available this week

PR #722, opened 2024-05-02
Quantum machines 

jordivallsq:
There was a rounding error involving the signs <= and >= in the compilation of the QUA program's for loops, the error caused results not to generate showcasing a bug.

The signs have been replaced by < and > and the stop value has been replaced by stop +- step/2.

This solution has been checked using the experiment located in qililab-portfolio/experiments/qprogram/flux_vs_flux_pulse.ipynb, if the experiment is executed with a gain step number greater than 9 without this hotfix the results will crash and not return anything.


jordivallsq:
@paulqili tested and approved the bug fix with his computer

jjmartinezQT (approved):
Good work!

PR #727, opened 2024-05-15
Qhc 466 bug wait instruction is not compiled proprely on qprogam for

jordivallsq:
Following the description on the QM reference https://docs.quantum-machines.co/1.1.6/qm-qua-sdk/docs/API_references/qua/dsl_main/#qm.qua._dsl.wait, the used by the QM compiler now is divided by 4 which returns the correct amount of ns waited (before it was multiplied by 4)

### DIFF
--- a/src/qililab/qprogram/quantum_machines_compiler.py
+++ b/src/qililab/qprogram/quantum_machines_compiler.py
@@ -451,7 +451,13 @@ class QuantumMachinesCompiler:  # pylint: disable=too-many-instance-attributes, 
             if isinstance(element.duration, Variable)
             else max(element.duration, self.MINIMUM_TIME)
         )
-        qua.wait(duration // self.WAIT_COEFF, bus)
+
+        qua.wait(
+            duration / int(self.WAIT_COEFF)
+            if isinstance(element.duration, Variable)
+            else int(duration / self.WAIT_COEFF),
+            bus,
+        )
 
     def _handle_sync(self, element: Sync):
         if element.buses:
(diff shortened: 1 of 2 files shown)


src/qililab/qprogram/quantum_machines_compiler.py
```diff
@@ -450,7 +451,13 @@ def _handle_wait(self, element: Wait):
             if isinstance(element.duration, Variable)
             else max(element.duration, self.MINIMUM_TIME)
         )
-        qua.wait(duration, bus)
+
+        qua.wait(
+            duration / int(self.WAIT_COEFF)
```
fedonman, line 456: If `duration` is of type `Variable`, then the division `duration / int(self.WAIT_COEFF)` will not make sense. What needs to be done is change the values of the loop instead, inside the `_handle_for_loop`, `_handle_loop` and `_handle_parallel_loop` methods.

For example, the phases in the `_handle_for_loop` change as follows:
```
if element.variable.domain is Domain.Phase:
            start, stop, step = start / self.PHASE_COEFF, stop / self.PHASE_COEFF, step / self.PHASE_COEFF
```

tests/qprogram/test_quantum_machines_compiler.py
fedonman, line 1: Add/Edit the tests for the loops as well, eg. the `def test_for_loop(self, for_loop: QProgram)`.

fedonman (changes requested):
Static timings seem correct, I added some comments for the handling of loops.

src/qililab/qprogram/quantum_machines_compiler.py
fedonman: Nevermind, I just saw that `duration` is a qua variable, so this works of course. Well done

fedonman (approved):
My bad, this actually works perfectly. Well done! 💯

PR #734, opened 2024-05-29
[QHC-262] Add measure method for qprogram qblox

visagim:
- [ ] changelog

### DIFF
--- a/src/qililab/qprogram/qblox_compiler.py
+++ b/src/qililab/qprogram/qblox_compiler.py
@@ -26,6 +26,7 @@ from qpysequence.utils.constants import INST_MAX_WAIT
 from qililab.qprogram.blocks import Average, Block, ForLoop, InfiniteLoop, Loop, Parallel
 from qililab.qprogram.operations import (
     Acquire,
+    Measure,
     Operation,
     Play,
     ResetPhase,
@@ -99,6 +100,7 @@ class QbloxCompiler:  # pylint: disable=too-few-public-methods
             SetOffset: self._handle_set_offset,
             Wait: self._handle_wait,
             Sync: self._handle_sync,
+            Measure: self._handle_measure,
             Acquire: self._handle_acquire,
             Play: self._handle_play,
         }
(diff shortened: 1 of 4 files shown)


src/qililab/qprogram/qblox_compiler.py
```diff
@@ -400,7 +402,22 @@ def __handle_static_sync(self, buses: set[str]):
...
+
+        Args:
+            element (Measure): measure operation
+        """
+        if not isinstance(element.weights, IQPair):
+            raise NotImplementedError("Qblox measure operation only supports weight format as IQPairs")
+        play = Play(bus=element.bus, waveform=element.waveform)
+        acquire = Acquire(bus=element.bus, weights=element.weights)
+        self._handle_play(play)
+        self._handle_acquire(acquire)
```
fedonman, line 417: The deletion of `Acquire` and `_handle_acquire` will be done in a later PR, right?

fedonman: Also in the final version, I would prefer all the logic to be in the `_handle_measure`  directly. (without creating new objects or calling other handlers manually) If there is duplicate code (e.g. making the Q1ASM play instruction) we should create a separate method that does that.

This is both for consistency with the rest of the code and also that probably you will need to do different things in creating the Q1ASM for Play and for Measure. E.g. Play won't have a `time_of_flight` while Measure will have.

fedonman (approved):
Nice 💯 Left some comments to take into account.

fedonman:
For the  changelog, I believe it's best to add one single entry at the last PR, explaining all changes. Because these interim PRs introduce temporary code.

src/qililab/qprogram/qblox_compiler.py
visagim: > The  deletion of Acquire and _handle_acquire will be done in a later PR, right?

this is what I discussed at:
https://linear.app/qilimanjaro/issue/QHC-262/delete-acquire-and-implement-measure-in-qbloxcompiler#comment-fc64d49c

If we add this and ToF at the same time then I don't mind deleting `handle_aqcuire`

PR #738, opened 2024-06-10
[QHC-264] Add `time_of_flight` to Qblox and use it for compiling `measure` operation

fedonman:


src/qililab/instruments/awg_analog_digital_converter.py
```diff
@@ -330,6 +333,19 @@ def _set_scope_store_enabled(self, value: float | str | bool, sequencer_id: int)
         """
         cast(AWGADCSequencer, self.get_sequencer(sequencer_id)).scope_store_enabled = bool(value)
 
+    @Instrument.CheckParameterValueFloatOrInt
+    def _set_time_of_flight(self, value: float | str | bool, sequencer_id: int):
+        """set scope_store_enable
```
jordivallsq, line 338: This docstring is fir the scope_store enable.
Also the value may also be a Bool, is that right? Shouldn't it be an int (float at best if the user sets an int as a float)?

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -435,7 +448,8 @@ def _handle_measure(self, element: Measure):
         Args:
             element (Measure): measure operation
         """
-        play = Play(bus=element.bus, waveform=element.waveform)
+        time_of_flight = self._buses[element.bus].time_of_flight
+        play = Play(bus=element.bus, waveform=element.waveform, wait_time=time_of_flight)
```
jordivallsq, line 452: Do we have a default time of flight? In Q1ASM this parameter in the Play command needs to be filled, what happens if we do not add a time of flight mannually, I believe that the minimum is 4ns, should we set this as a default?

jordivallsq:
Great job with the addition of TOF!!! I left a couple of comments.

Probably you already have this in your scope but is it planned to add the time of flight for QM in a similar way? I suppose the only thing that needs to change is the quantum_machines_compiler (and other minor adjustments) as the basis is already set.

Additionally in QM we also have smearing (among other different variables on measure), could we also add these on your methodology? I am actually asking because it is QM exclusive so I don't know if (or where) we can do the same as with the time of flight.

fedonman:
> Great job with the addition of TOF!!! I left a couple of comments.
> 
> Probably you already have this in your scope but is it planned to add the time of flight for QM in a similar way? I suppose the only thing that needs to change is the quantum_machines_compiler (and other minor adjustments) as the basis is already set.
> 
> Additionally in QM we also have smearing (among other different variables on measure), could we also add these on your methodology? I am actually asking because it is QM exclusive so I don't know if (or where) we can do the same as with the time of flight.

Time of flight is already implemented in Quantum Machines, natively by them. We already provide the parameter in the runcard.

Smearing is a QM-specific parameter, I don't know if it makes sense or if it is even possible to implement something similar for Qblox.

src/qililab/qprogram/qblox_compiler.py
fedonman: Yes we have a default value. It's the `minimum_wait_duration` which is 4.

```
# Time of flight. Defaults to minimum_wait_duration and is updated if times_of_flight parameter is provided during compilation.
        self.time_of_flight = QbloxCompiler.minimum_wait_duration
```

jordivallsq:
> > Great job with the addition of TOF!!! I left a couple of comments.
> > Probably you already have this in your scope but is it planned to add the time of flight for QM in a similar way? I suppose the only thing that needs to change is the quantum_machines_compiler (and other minor adjustments) as the basis is already set.
> > Additionally in QM we also have smearing (among other different variables on measure), could we also add these on your methodology? I am actually asking because it is QM exclusive so I don't know if (or where) we can do the same as with the time of flight.
> 
> Time of flight is already implemented in Quantum Machines, natively by them. We already provide the parameter in the runcard.
> 
> Smearing is a QM-specific parameter, I don't know if it makes sense or if it is even possible to implement something similar for Qblox.

Thank you for the clarification about QM TOF. About smearing, I was referring to QM exclusively (I don't think either that it can be implemented in Qblox) as currently we don't have an option to define this parameter (? at least I haven't seen one) and we should give the user this option in the `qp.quantum_machines.measure()`. Apologies for the confusion, I realize that this is not the best place for a QM comment XD.

On the other hand once the couple of comments are resolved and this sub-branch can be merged without issue

src/qililab/instruments/awg_analog_digital_converter.py
fedonman: fixed

jordivallsq (approved):
Great work!! Aproved

PR #739, opened 2024-06-11
[QHC-543] Remove `path` argument, from `build_platfom()`

GuillermoAbadLopez:
Removing deprecated argument (`path` from `build_platform`), we should have deleted, some versions ago already 👌

### DIFF
--- a/src/qililab/data_management.py
+++ b/src/qililab/data_management.py
@@ -225,7 +225,7 @@ def build_platform(runcard: str | dict, connection: API | None = None, new_drive
     """
     if not isinstance(runcard, (str, dict)):
         raise ValueError(
-            f"Mandatory `runcard` argument in `build_platform()`, is not a supported type: (str, dict), is type: {type(runcard)}."
+            f"Incorrect type for `runcard` argument in `build_platform()`. Expected (str | dict), got: {type(runcard)}"
         )
 
     if new_drivers:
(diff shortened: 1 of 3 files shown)


src/qililab/data_management.py
```diff
@@ -227,17 +223,10 @@ def build_platform(
...
-    if path is not None:
-        if runcard is not None:
-            raise ValueError("Use only the `runcard` argument, `path` argument is deprecated.")
-        warn(
-            "`path` argument is deprecated and will be removed soon. Use the `runcard` argument instead.",
-            DeprecationWarning,
-            stacklevel=2,
+    if not isinstance(runcard, (str, dict)):
+        raise ValueError(
+            f"Mandatory `runcard` argument in `build_platform()`, is not a supported type: (str, dict), is type: {type(runcard)}."
```
IsaacLA12, line 228: I would change the way we say it to something least confusing like "Incorrect type for mandatory argument `runcard` in `build_platform(). Expected (str | dict) but given type: {type(runcard)}`"

IsaacLA12 (approved):
Gj! 🚀

src/qililab/data_management.py
GuillermoAbadLopez: I agree the sentence, its convoluted at the start! Changing it!

GuillermoAbadLopez: Commited! 🚀

PR #740, opened 2024-06-12
[QHC-548] Fix QProgram's bus mapping bug

fedonman:


jordivallsq (approved):
The changes are good, thank you!

One question though, this non_existant_mapping_qp test what is showing? What means putting "non_existant" in the bus mapping?

fedonman:
> The changes are good, thank you!
> 
> One question though, this non_existant_mapping_qp test what is showing? What means putting "non_existant" in the bus mapping?

I wanted to make sure that if you map a bus that doesn't exist in the qprogram, you won't get an error.

PR #743, opened 2024-06-27
[QHC-561] QProgram (qblox) results: Classification

IsaacLA12:
We add general methods for the `QProgramResults` class to obtain the state probabilities of an experiment, we can compute the probabilities via two different methods depending if we received the results via hardware average or not (with HW average == to 1)

### DIFF
--- a/src/qililab/result/qprogram/qprogram_results.py
+++ b/src/qililab/result/qprogram/qprogram_results.py
@@ -13,6 +13,11 @@
 # limitations under the License.
 
 """MeasurementResult class."""
+from itertools import product
+
+import numpy as np
+
+from qililab.result.counts import Counts
 from qililab.result.qprogram.measurement_result import MeasurementResult
 from qililab.yaml import yaml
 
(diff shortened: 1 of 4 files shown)


src/qililab/result/qprogram/qprogram_results.py
```diff
@@ -35,3 +38,53 @@ def append_result(self, bus: str, result: MeasurementResult):
         if bus not in self.results:
             self.results[bus] = []
         self.results[bus].append(result)
+
+
+def probabilities(qprogram_results: QProgramResults, qubit_mapping: list[str] | None = None) -> dict[str, float]:
```
fedonman, line 43: I had the impression from our last meeting that this functionality would be outside of the `QProgramResults`, since it is a special case anyway.

Ramiro called it with a name, but can't remember, let's call it XYZ. The main component of this class is this matrix he described.

Then it's better to have a class called XYZ that contains the representation. The main advantage is that this class can be used independently of QProgram: the user might create similar results with other methods, either circuit or manually.

The class will also contain a class method `from(qprogram_results: QProgramResults, qubit_mapping: list[str] | None = None`) to do what this function implements.

If you think the XYZ class is the `Counts` class that we already have, implement the `from()` method there.

src/qililab/result/qprogram/qprogram_results.py
```diff
@@ -13,6 +13,9 @@
 # limitations under the License.
 
 """MeasurementResult class."""
+import numpy as np
+
+from qililab.result.counts import Counts
```
fedonman, line 18: Another clue for the reorganization I ask below, is the fact that now QProgram module depends on `Counts` which is outside of QProgram module. We try for QProgram to be completely independent.

fedonman (changes requested):
Great job :100: I ask for some re-organization of the code so not to make QProgram dependent on other code, and follow OOP guidelines.

src/qililab/result/qprogram/qprogram_results.py
IsaacLA12: Oh yeah, It is not part of the class, (it is not indented). I just left it there because I didn't know if I should move it to a new file, but yeah, I can definitely move it into a new class or maybe add it to the `Counts` module. 🙌

IsaacLA12: In the end I moved it out to a static method of the `Counts` class

src/qililab/result/counts.py
```diff
@@ -124,3 +126,56 @@ def __iadd__(self, other: Counts) -> Counts:
...
         return self
+
+    @staticmethod
+    def from_qprogram(qprogram_results, qubit_mapping: list[str] | None = None):
+        """Returns `Count` object from a `QProgramResults`.
+
+        Args:
+            qprogram_results (QProgramResults): The QProgramResults object we want to get the `Counts` from.
+            qubit_mapping (list[str], optional): A list containing the name of the busses to map.
+                The buses are map to qubits on the same index, a bus at the i-th element in the list is mapped to the i-th qubit.
```
fedonman, line 137:
```suggestion
            qubit_mapping (list[str], optional): A list containing the names of the buses to map.
                Each bus is mapped to a qubit at the corresponding index; the bus at the i-th position in the list is mapped to the i-th qubit.
```

src/qililab/result/counts.py
```diff
@@ -124,3 +126,56 @@ def __iadd__(self, other: Counts) -> Counts:
...
+                )
+            if not unique_mapping.issubset(set(buses)):
+                raise ValueError(
+                    "No measurements found for all specified buses, check the name of the buses provided with the mapping match all the buses specified in runcard."
+                )
+            buses = qubit_mapping
+
+        # The threshold inside of a qblox bin is the name they use for already classified data as a value between
+        # 0 and 1, not the value used in the comparator to perform such classification.
+        th_matrix = np.array(
```
fedonman, line 170: Can we avoid abbreviations? We can rename this to `threshold_matrix`

src/qililab/result/counts.py
```diff
@@ -124,3 +126,56 @@ def __iadd__(self, other: Counts) -> Counts:
...
+            [
+                np.concatenate([measurement.threshold.astype(int) for measurement in qprogram_results.results[bus]])
+                for bus in buses
+            ]
+        )
+        th_matrix_T = th_matrix.transpose()
```
fedonman, line 176: And this to `threshold_matrix_T`.

src/qililab/result/counts.py
```diff
@@ -124,3 +126,56 @@ def __iadd__(self, other: Counts) -> Counts:
...
+                Defaults to None.
+
+        Raises:
+            ValueError: When the `results` attribute from `qprogram_results` is an empty dictionary.
+            ValueError: When a qubit mapping is incomplete and does not map all qubits.
+            ValueError: When a qubit mapping is specified and any of the busses does not match with any on the runcard.
+
+        Returns:
+            dict[str, float]: Dictionary containing the quantum states as the keys of the dictionary, and the
+                probabilities obtained for each state as the values of the dictionary.
```
fedonman, line 147:
```suggestion
            dict[str, float]: A dictionary where the keys are the quantum states, and the values are the probabilities obtained for each state.
```

src/qililab/result/counts.py
```diff
@@ -124,3 +126,56 @@ def __iadd__(self, other: Counts) -> Counts:
...
+        """Returns `Count` object from a `QProgramResults`.
+
+        Args:
+            qprogram_results (QProgramResults): The QProgramResults object we want to get the `Counts` from.
+            qubit_mapping (list[str], optional): A list containing the name of the busses to map.
+                The buses are map to qubits on the same index, a bus at the i-th element in the list is mapped to the i-th qubit.
+                Defaults to None.
+
+        Raises:
+            ValueError: When the `results` attribute from `qprogram_results` is an empty dictionary.
```
fedonman, line 141:
```suggestion
           ValueError: If the `results` attribute of `qprogram_results` is an empty dictionary.
```

src/qililab/result/counts.py
```diff
@@ -124,3 +126,56 @@ def __iadd__(self, other: Counts) -> Counts:
...
+            ValueError: When a qubit mapping is incomplete and does not map all qubits.
```
fedonman, line 142:
```suggestion
            ValueError: If the qubit mapping is incomplete and does not map all qubits.
```

src/qililab/result/counts.py
```diff
@@ -124,3 +126,56 @@ def __iadd__(self, other: Counts) -> Counts:
...
+            ValueError: When a qubit mapping is specified and any of the busses does not match with any on the runcard.
```
fedonman, line 143:
```suggestion
            ValueError: If a qubit mapping is specified and any of the buses do not match the ones in the runcard.
```

docs/releases/changelog-dev.md
```diff
@@ -103,6 +103,23 @@
 
   [#747](https://github.com/qilimanjaro-tech/qililab/pull/747)
 
+- Added `from_qprogram` method to compute the counts of the quantum states obtained with `QProgram`. The `Counts` object is expected to work for circuits which has only a measurement per bus at the end of the excution of the circuit so is responsability of the user use this method when it makes sense to compute the counts of the states or not of a `QProgram`. Note we can easily obtain the probabilities by calling the method `probabilities()`. See an example below.
```
fedonman, line 106:
```suggestion
- Added `from_qprogram` method to the `Counts` class to compute the counts of quantum states obtained from a `QProgram`. The `Counts` object is designed to work for circuits that have only one measurement per bus at the end of the circuit execution. It is the user's responsibility to ensure that this method is used appropriately when it makes sense to compute the state counts for a `QProgram`. Note that probabilities can easily be obtained by calling the `probabilities()` method. See an example below.
```

fedonman (commented):
Awesome job mate! :100: Left minor comments on grammar and naming suggestions.

src/qililab/result/counts.py
IsaacLA12: Done!

src/qililab/result/counts.py
IsaacLA12: Sure!

fedonman (approved):

PR #745, opened 2024-07-02
Implemented an aditional error on IQPair

jordivallsq:
Found a bug on qilitools where an IQPair contained 2 other IQPair (this was a mistake). The error was difficult to trace back because the class IQPair accepted other options beside Waveform without complaining. I added an exception so if you create an IQPair with something other than a Waveform it throws a clear error.

### DIFF
--- a/src/qililab/waveforms/iq_pair.py
+++ b/src/qililab/waveforms/iq_pair.py
@@ -35,11 +35,11 @@ class IQPair(DictSerializable):  # pylint: disable=missing-class-docstring
     Q: Waveform
 
     def __post_init__(self):
-        if self.I.get_duration() != self.Q.get_duration():
-            raise ValueError("Waveforms of an IQ pair must have the same duration.")
-
         if not isinstance(self.I, Waveform) or not isinstance(self.Q, Waveform):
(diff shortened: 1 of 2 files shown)


jjmartinezQT (commented):
Looks good to me but Vyron should have a final word in case I have missed something about this.

src/qililab/waveforms/iq_pair.py
```diff
@@ -38,6 +38,9 @@ def __post_init__(self):
         if self.I.get_duration() != self.Q.get_duration():
             raise ValueError("Waveforms of an IQ pair must have the same duration.")
 
+        if not isinstance(self.I, Waveform) or not isinstance(self.Q, Waveform):
```
fedonman, line 41: Maybe it's better to make this check first and then check equality of durations? Since we first want to ensure we have the correct class and then call the method of that class.

fedonman (approved):
Nice catch! :100: I added a comment on the order of the checks.

src/qililab/waveforms/iq_pair.py
jordivallsq: Good suggestion, I made the necessary changes.

PR #750, opened 2024-07-05
Solve `CodeQuality` problem

GuillermoAbadLopez:


IsaacLA12 (approved):
Weird that code quality was complaining and with this it gets fixed, gj! 💯

PR #751, opened 2024-07-09
Qhc 588 bug `get_parameter` does not work with QM platform

jordivallsq:
Fixed the bug on QM that crashed the get_parameter function and changed the QM get_parameter to work without connection to the instrument


.pylintrc
```diff
@@ -20,4 +20,5 @@ disable=
     too-many-ancestors,
     missing-module-docstring,
     too-many-arguments,
-    too-many-branches
\ No newline at end of file
+    too-many-branches,
+    too-many-return-statements
```
GuillermoAbadLopez, line 24: Why are we adding this in here? 

If its a single line/case, maybe just add a:
```
# pylint: disable = too-many-return-statements
```

docs/releases/changelog-dev.md
```diff
@@ -27,9 +30,9 @@
 
 ### Deprecations / Removals
 
-- Deleted all the files in `execution` and `experiment` directories (Already obsolete).
```
GuillermoAbadLopez, line 30: Why are you removing this previous change-log? 👀

docs/releases/changelog-dev.md
```diff
@@ -27,9 +30,9 @@
...
-  [#749](https://github.com/qilimanjaro-tech/qililab/pull/749)
-
 ### Documentation
 
 ### Bug fixes
+
+get_parameter for QM did not work due to the lack of the variable bus_alias in self.system_control.get_parameter. The variable has been added to the function and now get parameter does not return a crash.
```
GuillermoAbadLopez, line 37: I think you put the `-` in the wrong place, no?
```suggestion
- get_parameter for QM did not work due to the lack of the variable `bus_alias in self.system_control.get_parameter`. The variable has been added to the function and now get parameter does not return a crash.
```
(also need to be deleted from the next line!)

docs/releases/changelog-dev.md
```diff
@@ -27,9 +30,9 @@
...
+ - [#751] https://github.com/qilimanjaro-tech/qililab/pull/751
```
GuillermoAbadLopez, line 38:
```suggestion 
[#751] https://github.com/qilimanjaro-tech/qililab/pull/751
```

GuillermoAbadLopez: Also need to be deleted from the next line?!

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -66,7 +66,7 @@ class QuantumMachinesClusterSettings(Instrument.InstrumentSettings):
         controllers: list[dict[str, Any]]
         elements: list[dict[str, Any]]
 
-        def to_qua_config(self) -> Dict[str, Any]:
+        def to_qua_config(self) -> DictQuaConfig:
```
GuillermoAbadLopez, line 69: Aren't we missing the `oscillators` key, for it to be a `DictQuaConfig`?

GuillermoAbadLopez: I like the change though! 

Cool concretization 👍

src/qililab/system_control/system_control.py
GuillermoAbadLopez, line 1: Also I just committed here these missing docstring arguments, hope its okei 👍

GuillermoAbadLopez (commented):
Looks nice, and it works perfectly!

I've left some more minors comments, and one less minor concern, that I wanna look into, but please feel free to answer/change anything in the meantime!  GJ! 🚀

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -378,12 +378,39 @@ def get_parameter_of_bus(self, bus: str, parameter: Parameter):
...
+            self._config = self.settings.to_qua_config()
+
+        config_keys = self._config["elements"][bus]
+
         if parameter == Parameter.LO_FREQUENCY:
-            return self._qm._elements[bus].input.lo_frequency  # type: ignore[union-attr] # pylint: disable=protected-access
-        if parameter == Parameter.GAIN:
-            return self._qm._elements[bus].input.gain  # type: ignore[union-attr] # pylint: disable=protected-access
+            if "mixInputs" in config_keys:
+                return self._config["elements"][bus]["mixInputs"]["lo_frequency"]
```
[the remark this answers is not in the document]
GuillermoAbadLopez: Okei, I see before we were using the default classes and attributes from QM.. and now with our own its more complicated, I guess this will work then! 👌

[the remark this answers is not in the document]
GuillermoAbadLopez: In the future we could simplify the code with a class, because right now, we are repeating a lot the:
```self._config["elements"][bus][Parameter]```
 But I guess its acceptable now, since you are not the first to do it even 👍

File: src/qililab/instruments/quantum_machines/quantum_machines_cluster.py [hunk 7]
```diff
@@ -378,12 +378,39 @@ def get_parameter_of_bus(self, bus: str, parameter: Parameter):
             ParameterNotFound: Raised if parameter does not exist
         """
         # TODO: Change private QM API to public when implemented.
+
+        if "_config" not in dir(self):
```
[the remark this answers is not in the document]
GuillermoAbadLopez: I don't think this is the place to solve this of course, but maybe we can decide if we want the `settings` always ruling over the `_config` in this case, just in case, removing the `conditional if` @fedonman??
```suggestion
```
(also we would need to untab next line...)

-----

Or else, if we want the `_config` to rule, for which then its good as it is

GuillermoAbadLopez (commented):
@fedonman @jordivallsq What do you think, what should rule over the other, `_config` or `settings`, is there any reason they would be different?

Or in theory if everybody coded correctly they are aways synch up?

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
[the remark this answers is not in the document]
GuillermoAbadLopez: Depending on what we decide here, that would also affect, how I write it in this other PR, for the `set_parameters` method: https://github.com/qilimanjaro-tech/qililab/pull/752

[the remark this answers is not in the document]
GuillermoAbadLopez: In the `inital_setup` we are directly doing: `self._config = self.settings.to_qua_config()` like the settings rules over the `._config`, maybe we should follow that everywhere? 🤔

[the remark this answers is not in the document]
GuillermoAbadLopez: # AFTER A MEETING, WE DECIDED:

1.a)  Synching or Prioritizing  settings vs _config (#754, or define sync?)
- :white_check_mark: Both are equally important/priorty - So we have to be careful with synching between them

1.b) Do we want to save the changed info in append_configuration or is it just for the execution?
- :white_check_mark:  We need it yes, no temporary!
-- :x: If not -> This would imply we can delete _config as an attribute, only property (at least not now)
-- :x: If we want -> Then we can do reverse to_qua_config() function
-- :white_check_mark: We then have to correctly define and check that both infos are synch all the time:

2.) Initialization of _qm and _qmm is kind of mixed, no? (setup needed before turn_on..)
- :white_check_mark: Move everything into initial_setup , but lets wait into a better moment? (to check/test properly)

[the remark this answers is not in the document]
GuillermoAbadLopez: I'm going to commit some changes to adapt for these decisions 👍
Resolving this comment then

GuillermoAbadLopez:
Did this commit: https://github.com/qilimanjaro-tech/qililab/pull/751/commits/916d5d7bb22cdb9f2ebbbbd7a799637550c511d1, adding:
- `_config_exists` when we `initial_setup` and the `_config` is created
- Make that we don't generate a `_config` in the `get_parameter()`, but only where its supposed to happen, in the `initial_setup`, creating a temporary dictionary if it doesn't exist before hand👍 (This code can be changed by getting parameters from the `settings` instead)

GuillermoAbadLopez:
@jordivallsq appart than the still open comment, this one I think its ready to go!

Check the last changes, when you can, please! 🙏 
(tomorrow I'll address/resolve my still opened silly comments 👍 )

docs/releases/changelog-dev.md
GuillermoAbadLopez: Addressed!

docs/releases/changelog-dev.md
GuillermoAbadLopez: Addressed!

docs/releases/changelog-dev.md
GuillermoAbadLopez: Addressed

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -378,12 +378,39 @@ def get_parameter_of_bus(self, bus: str, parameter: Parameter):
             ParameterNotFound: Raised if parameter does not exist
```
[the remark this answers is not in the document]
GuillermoAbadLopez: Added a short description myself, see if you want to improve it!

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -378,12 +378,39 @@ def get_parameter_of_bus(self, bus: str, parameter: Parameter):
...
+                return self._config["octaves"][port[0]]["RF_outputs"][port[1]]["LO_frequency"]
         if parameter == Parameter.IF:
-            return self._qm._elements[bus].intermediate_frequency  # pylint: disable=protected-access
+            if "intermediate_frequency" in config_keys:
+                return self._config["elements"][bus]["intermediate_frequency"]
+        if parameter == Parameter.GAIN:
+            if "mixInputs" in config_keys and "outputs" in config_keys:
+                port_i = self._config["elements"][bus]["outputs"]["out1"]
+                port_q = self._config["elements"][bus]["outputs"]["out2"]
+                return (
```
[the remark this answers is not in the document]
GuillermoAbadLopez: **Update:** In the return docstring I wrote, I specified all these possible return types:
`float | str | bool | tuple[float]`

.pylintrc
GuillermoAbadLopez: Okei, lets make it general then! 
I'll remove the other places that have this individually!

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
GuillermoAbadLopez: Seems like we are not using the `oscillators` key anywhere, so that why they are not there! Resolving 👌

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -378,12 +378,39 @@ def get_parameter_of_bus(self, bus: str, parameter: Parameter):
...
+
+        config_keys = self._config["elements"][bus]
+
         if parameter == Parameter.LO_FREQUENCY:
-            return self._qm._elements[bus].input.lo_frequency  # type: ignore[union-attr] # pylint: disable=protected-access
-        if parameter == Parameter.GAIN:
-            return self._qm._elements[bus].input.gain  # type: ignore[union-attr] # pylint: disable=protected-access
+            if "mixInputs" in config_keys:
+                return self._config["elements"][bus]["mixInputs"]["lo_frequency"]
+            if "RF_inputs" in config_keys:
```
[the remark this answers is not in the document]
GuillermoAbadLopez: I guess, there should never be both at the same time, right? 
Can you confirm this?

GuillermoAbadLopez (approved):
It looks nice to me now! 🚀🚀🚀 

Please @jordivallsq take a look, since I added several changes to your version,
and let's merge it, whenever the OPX-1000 is working and we can test it 🔥

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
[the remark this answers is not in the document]
jordivallsq: Sorry for the late reply, RF_inputs uses and is linked to the octave and mixInputs does not require any octave to work (essencially is for when the local oscillator is outside. In some case they can coexist in the same experiment though (a readout uses the octave and a drive bus does not), but they will never be defined in the same bus

jordivallsq (commented):
Great changes and additions we can merge this branch

jordivallsq:
Added @GuillermoAbadLopez changes on set_parameter

PR #647, opened 2023-12-20
Make that `platform.set_parameters()` doesnt need connection

GuillermoAbadLopez:
Steps to be done:

- [x] Adding if self.device is not None in all sets/setups of instruments, buses, etc… 

- [x] Remove the @CheckDeviceInitialised decorator from setups

- [x] Make sure platform.initial_setup raises an error when we are not connected

- [x] Add documentation in platform.set_parameter to make sure users understand what happens if they are/aren't connected to the instruments.

- [x] Change tests to work with `if self.device is not None` instead than with the flag:

- [x] Changed `system_controll`, so the `port_id` flag is removed from everywhere else in qililab

PR #752, opened 2024-07-16
[QHC-610] Make that `set_parameter` for QM, works without connection aswell

GuillermoAbadLopez:
# We need to be able to set a parameter, without connection:
We need to do the same we did for `Qblox` (here: https://github.com/qilimanjaro-tech/qililab/pull/647), so that when we set a parameter without connection, it changes the execution-time `Runcard` locally, so when you `connect()` and `initial_setup()`, it gets set up on the actual physical instruments, as the documentation says:
Screenshot 2024-07-16 at 12 15 24

This means be able to run and pass this test (`set` and `get` a parameter without connection):
Screenshot 2024-07-16 at 19 01 24


GuillermoAbadLopez:
Be aware, that this might be problematic, since we have the `self._config` dictionary, that duplicate the information of the runtime `Runcard`/`Settings`, and we would need to update both 😬:
https://qilimanjaro.slack.com/archives/C06R0KWN58W/p1721232248001349

GuillermoAbadLopez:
Depending on what we decide tomorrow in the daily, leave as it is if `_config` rules, or change it to always doing the:
```self._config = self.settings.to_qua_config()```
and maybe another in the end!

**Review by GuillermoAbadLopez — COMMENTED** [review]
Depending on what we decide tomorrow in the daily, leave as it is if _config rules, or change it to always doing the:
self._config = self.settings.to_qua_config()
and maybe another in the end!

**Comment by GuillermoAbadLopez** [comment]
Also, for merging this PR, we still need to check that a posterior `initial_setup`,
 will then set the parameters that have been changed locally
 - [ ] Follow code and check where they will get set in the `initial_setup` (problem with `_qm` and `_qmm` initialization)
 - [ ] Create test to check that this is indeed the case!

**Review by jordivallsq — COMMENTED** [review]
Great job, straightforward and well done. But let's wait for the QUA version before publishing this because we might be changing stuff here

File: tests/instruments/quantum_machines/test_quantum_machines_cluster.py [hunk 1]
```diff
@@ -316,6 +332,32 @@ def test_set_parameter_of_bus_method(
...
+    def test_set_parameter_without_connection_changes_settings(
+        self, mock_qmm, mock_qm, bus: str, parameter: Parameter, value: float | str | bool, qmm: QuantumMachinesCluster
+    ):
+        """Test that both the local `settings` and `_config` are changed by the set method without connection."""
+
+        # Set intermidiate frequency to 17e6 locally
+        qmm.set_parameter_of_bus(bus, parameter, value)
+
+        # Test that both the local `settings` and `_config` have been changed to 17e6:
+        # assert qmm._config == qmm.settings.to_qua_config() # TODO: Solve these dictionaries not being the same
```
GuillermoAbadLopez, line 355: `# TODO: Solve these dictionaries not being the same`

tests/instruments/quantum_machines/test_quantum_machines_cluster.py
```diff
@@ -199,6 +203,9 @@ def test_append_configuration_after_turn_on(
         qmm._qmm.open_qm.call_count == 2
         assert isinstance(qmm._qm, MagicMock)
 
+        # Assert that the settings are still synch:
+        # assert qmm._config == qmm.settings.to_qua_config() # TODO: Solve these dictionaries not being the same
```
GuillermoAbadLopez, line 207: `# TODO: Solve these dictionaries not being the same`

tests/instruments/quantum_machines/test_quantum_machines_cluster.py
```diff
@@ -186,6 +187,9 @@ def test_append_configuration(self, mock_qmm, mock_qm, qmm: QuantumMachinesClust
         assert "445e964c" in qmm._config["waveforms"]
         assert "fb58e912" in qmm._config["waveforms"]
 
+        # Assert that the settings are still synch:
+        # assert qmm._config == qmm.settings.to_qua_config() # TODO: Solve these dictionaries not being the same
```
GuillermoAbadLopez, line 191: `# TODO: Solve these dictionaries not being the same`

tests/instruments/quantum_machines/test_quantum_machines_cluster.py
```diff
@@ -316,6 +332,32 @@ def test_set_parameter_of_bus_method(
         if parameter == Parameter.IF:
             qmm._qm.set_intermediate_frequency.assert_called_once()
 
+        # Assert that the settings are still synch:
+        # assert qmm._config == qmm.settings.to_qua_config()  # TODO: Solve these dictionaries not being the same
```
GuillermoAbadLopez, line 336: `# TODO: Solve these dictionaries not being the same`

GuillermoAbadLopez (commented):
`append_configuration()` unsynchs `_config` and `settings` information... Discuss/solve it before merging:
image
(test failing in comments below)

---
**Update, to solve this:**
- And idea would be to pass `_config` from attribute to property, but then the `append_configuration()` function, wouldn't; update the actual settings instrument, but only update it for that execution! (Maybe this is good enough), like we do here: https://github.com/qilimanjaro-tech/qililab/pull/754

- Else we maybe will need a reverse function to `to_qua_config()`

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -354,52 +351,68 @@ def set_parameter_of_bus(self, bus: str, parameter: Parameter, value: float | st
 
             if parameter == Parameter.LO_FREQUENCY:
                 lo_frequency = float(value)
-                self._qm.octave.set_lo_frequency(element=bus, lo_frequency=lo_frequency)
+                # 1) Change settings:
```
jordivallsq, line 354: Is this really that confusing that needs these comments?

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -354,52 +351,68 @@ def set_parameter_of_bus(self, bus: str, parameter: Parameter, value: float | st
...
-
-            if f"mixer_{bus}" in self._config["mixers"]:
+            # 2) Change config if exists:
+            if self._config_created:
+                self._config["elements"][bus]["intermediate_frequency"] = intermediate_frequency
+            if self._config_created and f"mixer_{bus}" in self._config["mixers"]:
                 self._config["mixers"][f"mixer_{bus}"][0]["intermediate_frequency"] = intermediate_frequency
+            # 3) Change instrument if connected:
+            if self._is_connected_to_qm:
+                self._qm.set_intermediate_frequency(element=bus, freq=intermediate_frequency)
```
jordivallsq, line 397: Bear in mind that this will be changed in the future into job.set_intermediat_frequency (I will work in this basis for qua 1.2.0)

jordivallsq (approved):
Great PR!! I left some small comments but once they are resolved you can merge

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
GuillermoAbadLopez: Yeahh, Im not sure... also thought about removing them :)

 lemme make it simpler!

GuillermoAbadLopez: Addressed! Check again how it looks now!

GuillermoAbadLopez: Addressed @jordivallsq ! Check again how it looks now!

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
GuillermoAbadLopez: Cool! Tell me something if you need a hand with the implementation then!

GuillermoAbadLopez: But I guess this is not problem for this current code, no? 
Resolving then!

PR #753, opened 2024-07-17
adding mixers when doing set_parameter

jordivallsq:
When the elements are set in mix_inputs and we do a set_parameter for the IF the mixers if_frequency do not update correctly and couses a crash from qm side


jjmartinezQT (approved):
Great work!

PR #754, opened 2024-07-18
[QHC-628] Remove duplicated `_config` info

GuillermoAbadLopez:
1) To not have duplicated info in both `settings` and `_config`, we pass _config attribute to property.

2) This arises a problem, which is that in `append_configuration()` we were unsynchin them, so:
- Or we leave that unsynch only for execution without saving it (as `temporary_config`)
- Or we create a reverse: `to_qua_config()` function, such that we can also update the `settings`!

----
PR coming from initial Jordi's PR on the get without connection: https://github.com/qilimanjaro-tech/qililab/pull/751 and then my implementation for the set without connection: https://github.com/qilimanjaro-tech/qililab/pull/752


GuillermoAbadLopez:
We decided that for the moment, the `_config`, will remain as an attribute, so we can have saved the compiled extra configuration information! (And not have to compile everytime)

PR #755, opened 2024-07-18
[QHC-619] Update Runcard to QUA config flow to support opx1000

fedonman:


jjmartinezQT (approved):
I have compared this to the config file that I worked on with Theo and I think it covers everything.

jjmartinezQT (approved):

jjmartinezQT (approved):

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -21,7 +21,7 @@
 from qm import DictQuaConfig, QuantumMachine, QuantumMachinesManager, SimulationConfig
 from qm.jobs.running_qm_job import RunningQmJob
 from qm.octave import QmOctaveConfig
-from qm.qua import Program
+from qm.program import Program
```
jordivallsq, line 24: Was this change documented? I cannot find any references into it. How did you find out that this needed to change

jjmartinezQT (approved):

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
GuillermoAbadLopez: I think its just an improvement, since both get you to the same final class, but the former one, passed at least two extra files, with a lot of imports.. 

(which makes things slower and more circular import error prone I guess 🤔 )

GuillermoAbadLopez: I guess he saw with VSCode checkers, that the import wasn;t being successful..

And then if you go back to the old version of `qm-qua`, that class is the one that was being imported before (after a couple of file jumps 🙃)

jjmartinezQT (approved):

jjmartinezQT (approved):
damn units :D

jjmartinezQT (approved):

GuillermoAbadLopez:
@fedonman @jordivallsq @jjmartinezQT for solving the documentation issues that this PR is having,
 I created this other PR: https://github.com/qilimanjaro-tech/qililab/pull/758, take a look and if we merge it to main, the ones here should disappear 👍

jjmartinezQT (approved):
good idea!

PR #759, opened 2024-07-22
[QHC-605] Adding rotation thresholds to QM

IsaacLA12:
We add the necessary code to support initializing a QM cluster with a rotation threshold on the runcard. The platform now computes the existing thresholds for the readout buses (if any) and sends them to the compiler.

The QM compiler now handles the measurements with the rotation information on the runcard if no rotation is provided explicitly. If no rotation was specified nor found on the runcard, we use a default value of 0.


docs/releases/changelog-dev.md
```diff
@@ -17,6 +17,10 @@
 
   [#747](https://github.com/qilimanjaro-tech/qililab/pull/747)
 
+- Added `threshold_rotations` argument to `compile()` method in `QProgram`. This argument allows to use rotation angles on measurement instructions if not specified. Currently used to use the angle rotations specified on the runcard (if any) so the user does not have to explicitly pass it as argument to the measure instruction.  Used for classification of results in Quantum Machines's modules.
```
jordivallsq, line 20: If this is a new addition to the runcard please show here an example on how to use it as a segment of the runcard (specifying its location)

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -387,6 +391,11 @@ def get_parameter_of_bus(self, bus: str, parameter: Parameter):
             return self._qm._elements[bus].input.gain  # type: ignore[union-attr] # pylint: disable=protected-access
         if parameter == Parameter.IF:
             return self._qm._elements[bus].intermediate_frequency  # pylint: disable=protected-access
+        if parameter == Parameter.THRESHOLD_ROTATION:
```
jordivallsq, line 394: Note that set and get parameters will be changed to be able to do them while disconnected so you also have to take into account that this set and get THRESHOLD_ROTATION should work always.

Based on https://github.com/qilimanjaro-tech/qililab/pull/751

src/qililab/platform/platform.py
```diff
@@ -628,8 +628,18 @@ def execute_qprogram(
...
+                bus.alias: float(bus.get_parameter(parameter=Parameter.THRESHOLD_ROTATION))
+                for bus in buses
+                if isinstance(bus.system_control, ReadoutSystemControl)
+            }  # type: ignore
             return self._execute_qprogram_with_quantum_machines(
-                cluster=cluster, qprogram=qprogram, bus_mapping=bus_mapping, calibration=calibration, debug=debug
+                cluster=cluster,
+                qprogram=qprogram,
+                bus_mapping=bus_mapping,
+                threshold_rotations=threshold_rotations,  # type: ignore
```
jordivallsq, line 640: can this# type: ignore be solved by changing the _execute_qprogram_with_quantum_machines? shouldn't this be a dictionary or a None

src/qililab/qprogram/quantum_machines_compiler.py
```diff
@@ -322,14 +332,21 @@ def _handle_measure(
             if self._buses[element.bus].current_gain is not None
             else None
         )
+        rotation: float = (
```
jordivallsq, line 335: this is the only variable with a type should we remove it if it is causing a conflict?

jordivallsq (changes requested):
Very good job! I left some comments on this that have to be adressed.

I also have some notes: 

- You should wait for https://github.com/qilimanjaro-tech/qililab/pull/751 to be published on main, fetch it and then change the set and get of `Parameter.THRESHOLD_ROTATION` accordingly.
- Also look into QUA's 1.2.0 last changelog (https://docs.quantum-machines.co/1.2.0/qm-qua-sdk/docs/Releases/qm_qua_releases/#120-2024-07-02) and make sure that not of the functions you created enter in conlfict with it (unlikely though).
- You also should test that this rotation is well implemented looking through the generated QUA file through debug=True in the qprogram_execution function. If you haven't done this already

src/qililab/qprogram/quantum_machines_compiler.py
IsaacLA12: Yes you are right! I was testing some things regarding code quality and forgot to change it.

src/qililab/platform/platform.py
IsaacLA12: That is what I thought, the type for that argument in the method `_execute_qprogram_with_quantum_machines()` is indeed what you suggested but was still complaining so I added it.

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
IsaacLA12: Yes the getter and setter for that parameter does not need the instrument to be connected at all

IsaacLA12: Also thanks for pointing out that PR I did not notice we made the same change. I will wait for it to be merged before then

docs/releases/changelog-dev.md
IsaacLA12: Done!

IsaacLA12:
> Very good job! I left some comments on this that have to be adressed.
> 
> I also have some notes:
> 
> * You should wait for [Qhc 588 bug `get_parameter` does not work with QM platform #751](https://github.com/qilimanjaro-tech/qililab/pull/751) to be published on main, fetch it and then change the set and get of `Parameter.THRESHOLD_ROTATION` accordingly.
> * Also look into QUA's 1.2.0 last changelog (https://docs.quantum-machines.co/1.2.0/qm-qua-sdk/docs/Releases/qm_qua_releases/#120-2024-07-02) and make sure that not of the functions you created enter in conlfict with it (unlikely though).
> * You also should test that this rotation is well implemented looking through the generated QUA file through debug=True in the qprogram_execution function. If you haven't done this already

I have just checked the last release of QUA and there aren't any conflicts 👍 
Also I have tested the debug file generated the correct integration weights and to be sure it wont change in the future, the test that I have made checks exactly that! That the matrices A,B,C,D (which are the integration weights) represent what they should be

jordivallsq (commented):
Great changes if you can confirm me that the get and set work only with element and do not need the conditions implemented in the Get/set PR we can already merge the PR

GuillermoAbadLopez:
@IsaacLA12 there you have the conflicts, with the get/set PRs, if you need any help, please tell me! 🙌

GuillermoAbadLopez (commented):
@IsaacLA12 there you have the conflicts, with the get/set PRs, if you need any help, please tell me! 🙌

But my initial thought to solve the source code ones would be this:

Screenshot 2024-07-30 at 12 38 55
Screenshot 2024-07-30 at 12 37 48

Even though then in the get, from Jordi we are getting it from transforming the settings into QUA easier form, and not in yours, maybe it would be cool, to make both ways equal :)

_(we discussed it already and we prefer your way, directly from settings, but it's more complex codewise 😅)_

IsaacLA12:
> @IsaacLA12 there you have the conflicts, with the get/set PRs, if you need any help, please tell me! 🙌
> 
> But my initial thought to solve the source code ones would be this:
> 
> Screenshot 2024-07-30 at 12 38 55 Screenshot 2024-07-30 at 12 37 48
> Even though then in the get, from Jordi we are getting it from transforming the settings into QUA easier form, and not in yours, maybe it would be cool, to make both ways equal :)
> 
> _(we discussed it already and we prefer your way, directly from settings, but it's more complex codewise 😅)_

I see what you meant, I did this implementation in my case not just because it is what we discussed but also because this parameter is not set anywhere on the config nor on the qmm device once it is connected. It is a custom parameter that I added to have the same behaviour/structure we have with Qblox so I only have that parameter defined on the settings

IsaacLA12:
> Very good job! I left some comments on this that have to be adressed.
> 
> I also have some notes:
> 
> * You should wait for [Qhc 588 bug `get_parameter` does not work with QM platform #751](https://github.com/qilimanjaro-tech/qililab/pull/751) to be published on main, fetch it and then change the set and get of `Parameter.THRESHOLD_ROTATION` accordingly.
> * Also look into QUA's 1.2.0 last changelog (https://docs.quantum-machines.co/1.2.0/qm-qua-sdk/docs/Releases/qm_qua_releases/#120-2024-07-02) and make sure that not of the functions you created enter in conlfict with it (unlikely though).
> * You also should test that this rotation is well implemented looking through the generated QUA file through debug=True in the qprogram_execution function. If you haven't done this already

With the latest merge you will se I had to add some type ignores again, nevertheless the rest of the source code remained the same 🤓

GuillermoAbadLopez:
From my side (just checked `quantum_machines_cluster`), looks nice, GJ 🚀

jordivallsq (approved):
Thank you for the changes. Great Job!

PR #760, opened 2024-07-24
allowing more than a QM existing at same time

jjmartinezQT:
Setting the flag to allow multiple QM virtual machines at once. I think the qm.close is already executed when turning off the instrument, so if experimentalist do platform.disconnect() after each experiment, the virtual machine will be wipped off and not left floating around.


jordivallsq (approved):
Is this the only requirement from QM needed to connect multiple experiments? are there some restrictions in terms of no overlaping input channels or some sort of condition that can couse a crash?

jjmartinezQT:
> Is this the only requirement from QM needed to connect multiple experiments? are there some restrictions in terms of no overlaping input channels or some sort of condition that can couse a crash?

As reported by Yifei the way to do it is to set this flag to True and ensure qm.close() is done every time, I assume to not have zombie virtual machines all around or not to reach a limit. The overlapping of input channels I believe it should be a matter of configuration (runcard) but we need to test it with the experimentalists.

fedonman (changes requested):
This change could lead to a hanging QM that is not closed.

Imagine the scenario that you execute a QProgram, the QM is opened, but something goes wrong in the QUA execution in hardware. For example, the user uses a loop with a very large arbitrary array and memory overflow happens (Jordi had this issue at some point). The python execution is halted and `qm.close()` is never called. With `close_other_machines=True` this was not an issue for the next QProgram execution. With `close_other_machines=False` there will be a port conflict when trying to open the second QM.

I believe we should add some kind of `try/finally` to close the QM in any case.

jjmartinezQT:
> This change could lead to a hanging QM that is not closed.
> 
> Imagine the scenario that you execute a QProgram, the QM is opened, but something goes wrong in the QUA execution in hardware. For example, the user uses a loop with a very large arbitrary array and memory overflow happens (Jordi had this issue at some point). The python execution is halted and `qm.close()` is never called. With `close_other_machines=True` this was not an issue for the next QProgram execution. With `close_other_machines=False` there will be a port conflict when trying to open the second QM.
> 
> I believe we should add some kind of `try/finally` to close the QM in any case.

Makes sense, where do you suggest to try close the QM, execute_qprogram method?

jordivallsq (approved):
I agree with the exception introduced to close only the user's connection (although the name turn off is scary for the qm.close ^^

fedonman (approved):
Seems ok for me! :100: Have we test it in hardware with some on purpose exceptions?

Also, this might be changed in the future implementation of Session, since it will be responsible for handling these errors.

jjmartinezQT:
> Seems ok for me! 💯 Have we test it in hardware with some on purpose exceptions?
> 
> Also, this might be changed in the future implementation of Session, since it will be responsible for handling these errors.

Not yet tested in hardware, looking for the way to test it with the minimum impact to the guys measuring. Need to generate the on purpose exceptions, or a meaningful way to test this on hardware.

PR #761, opened 2024-07-24
Domain.Scalar now is defined correctly as an int on QM

jordivallsq:
I added a condition for the Domain.Scalar if the defined type is int


IsaacLA12 (approved):
Seems good to me! Nice catch! 🚀

PR #762, opened 2024-07-25
[QHC-562] Classification results QM

IsaacLA12:
We add the classifications of results in `QuantumMachinesMeasurementResult` by implementing the `threshold` property.
We followed the same classification method Qblox uses internally, see the [documentation](https://qblox-qblox-instruments.readthedocs-hosted.com/en/main/cluster/q1_sequence_processor.html#threshold).


src/qililab/platform/platform.py
```diff
@@ -727,10 +734,12 @@ def _execute_qprogram_with_quantum_machines(  # pylint: disable=too-many-locals
         acquisitions = cluster.get_acquisitions(job=job)
 
         results = QProgramResults()
+        # Doing manual classification of results as QM does not return thresholded values like Qblox
         for measurement in measurements:
             measurement_result = QuantumMachinesMeasurementResult(
-                *[acquisitions[handle] for handle in measurement.result_handles]
+                *[acquisitions[handle] for handle in measurement.result_handles],
             )
+            measurement_result.set_classification_threshold(thresholds.get(measurement.bus, None))
```
jordivallsq, line 742: Is there a default for this thesholds or is it a mandatory field and whenever we do the measurment without the threshold in the runcard it will crash, if that is the case please create an exception to execute an error

jordivallsq (commented):
I left a comment asI don't understand if this is a required field and in case it is now behaving like one I believe that it should be optional to include it in the runcard

src/qililab/platform/platform.py
IsaacLA12: It is not required in the runcard, when we initialize the QM measurement result class we initialize this property to `None`. Then when we ask for the thresholded results if the `threshold` is still None then an array of zeros is returned in order to keep consistency with `Qblox`. But it is true that maybe I could raise a warning and document it better on the docstrings

jordivallsq (approved):
Ok great then I have not more complaints, if you think you could add any exception for easier debugging please go ahead but besides that everything else looks good.
Make sure that it works on qm (as sometimes the tests are not enought) and everything will be done

PR #764, opened 2024-07-30
QHC 621 Modify set_intermediate_frequency for the jobAPI

jordivallsq:
Now the set_parameter(IF) creates a dictionary {bus: frequency} (between connection and execute) and it is called after creating the job (inside the execute qprogram)


jordivallsq:
I also added the function qm.calibrate_element(bus) after setting the IF and LO

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -252,12 +257,14 @@ def _get_elements_and_mixers_config(self) -> tuple:
...
-                        if "fem" in element["single_input"]
-                        else (element["single_input"]["controller"], element["single_input"]["port"]),
+                            (
+                                element["single_input"]["controller"],
+                                element["single_input"]["fem"],
+                                element["single_input"]["port"],
+                            )
+                            if "fem" in element["single_input"]
+                            else (element["single_input"]["controller"], element["single_input"]["port"])
+                        ),
```
fedonman, line 267: same here.

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -597,10 +626,19 @@ def run_compiled_program(self, compiled_program_id: str) -> RunningQmJob:
...
+        self.pending_job = self._qm.queue.add_compiled(  # pylint: disable=attribute-defined-outside-init
+            compiled_program_id
+        )
+
+        # TODO: job.wait_for_execution() is deprecated and will be removed in the future. Please use job.wait_until("Running") instead.
+        self.job = self.pending_job.wait_for_execution()  # type: ignore[return-value]  # pylint: disable=attribute-defined-outside-init
+        if self.controller == "opx1000" and self._intermediate_frequency:
+            for bus, intermediate_frequency in self._intermediate_frequency.items():
+                self.job.set_intermediate_frequency(element=bus, freq=intermediate_frequency)  # type: ignore[union-attr]
+                self._qm.calibrate_element(bus)
```
fedonman, line 639: Shouldn't we reset the `self._intermediate_frequency` after applying the values with the Job API? As it is now, it will apply the settings in every run, right?

fedonman (changes requested):
Great work! :100: Added some comments. The ones for parentheses are applied to other lines as well, but didn't add all of them.

Also, changelog entries are missing.

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
jordivallsq: Thank you for the observation, fixed

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -214,22 +215,26 @@ def _get_octaves_config(self) -> dict[str, Any]:
...
                         },
                         "IF_out2": {
                             "port": (
-                                octave["if_outputs"][1]["controller"],
-                                octave["if_outputs"][1]["fem"],
-                                octave["if_outputs"][1]["port"],
-                            )
-                            if "fem" in octave["if_outputs"][1]
-                            else (octave["if_outputs"][1]["controller"], octave["if_outputs"][1]["port"]),
+                                (
```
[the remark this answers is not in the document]
jordivallsq: This was done by black formater automatically while saving. In fact this changes has been applied on another PR and all the checks have accepted it. I am trying to revert the changes but Black in the checks complains

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -475,9 +478,34 @@ def run_octave_calibration(self):
...
+            controller_name = self._config["elements"][bus]["mixInputs"]["I"][0]
+        elif "singleInput" in self._config["elements"][bus]:
+            controller_name = self._config["elements"][bus]["singleInput"]["port"][0]
+
+        for controller in self.settings.controllers:
+            if controller["name"] is controller_name:
+                return controller["type"] if "type" in controller else "opx1"
+        raise AttributeError(f"Controller with bus {bus} does not exist")
+
+    def set_parameter_of_bus(  # pylint: disable=too-many-locals, too-many-statements  # noqa: C901ç
```
jordivallsq, line 506:
```suggestion
    def set_parameter_of_bus(  # pylint: disable=too-many-locals, too-many-statements  # noqa: C901
```

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -401,6 +402,8 @@ def _get_elements_and_mixers_config(self) -> tuple:
     _config: DictQuaConfig
     _octave_config: QmOctaveConfig | None = None
     _is_connected_to_qm: bool = False
+    _intermediate_frequency: dict[str, float] = {}
+    _controller: str | None = None
```
fedonman, line 406: Since this is used only in a single method, we don't need it as class attribute. We can convert it to a variable inside the method instead.

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -475,9 +478,34 @@ def run_octave_calibration(self):
         for element in elements:
             self._qm.calibrate_element(element)
 
-    def set_parameter_of_bus(  # pylint: disable=too-many-locals
+    def get_controller_from_bus(self, bus: str) -> str | None:
```
fedonman, line 481: Renaming since it returns just the type, instead of the actual controller.

```suggestion
    def get_controller_type_from_bus(self, bus: str) -> str | None:
```

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -646,10 +679,20 @@ def run_compiled_program(self, compiled_program_id: str) -> RunningQmJob:
...
-        pending_job = self._qm.queue.add_compiled(compiled_program_id)
-        # CHANGES: job.wait_for_execution() is deprecated and will be removed in the future. Please use job.wait_until("Running") instead.
-        return pending_job.wait_for_execution()  # type: ignore[return-value]
+        # TODO: qm.queue.add_compiled() -> qm.add_compiled()
+        self.pending_job = self._qm.queue.add_compiled(  # pylint: disable=attribute-defined-outside-init
+            compiled_program_id
+        )
+
+        # TODO: job.wait_for_execution() is deprecated and will be removed in the future. Please use job.wait_until("Running") instead.
+        self.job = self.pending_job.wait_for_execution()  # type: ignore[return-value]  # pylint: disable=attribute-defined-outside-init
```
fedonman, line 688: Do we need these two variables as attributes? They are not used elsewhere.

src/qililab/instruments/quantum_machines/quantum_machines_cluster.py
```diff
@@ -401,6 +402,8 @@ def _get_elements_and_mixers_config(self) -> tuple:
     _config: DictQuaConfig
     _octave_config: QmOctaveConfig | None = None
     _is_connected_to_qm: bool = False
+    _intermediate_frequency: dict[str, float] = {}
```
fedonman, line 405: Renaming suggestion for more clarity of the atrribute usage.
```suggestion
    _pending_set_intermediate_frequency: dict[str, float] = {}
```

fedonman (changes requested):
Awesome job :100: Minor details left to be addressed.

jordivallsq (commented):
I implemeted the suggested changes, thank you!

jjmartinezQT (approved):
good work.

PR #765, opened 2024-07-31
hotfix

jjmartinezQT:
Hotfix to always delete the acquisitions between measurements.


fedonman (approved):
Seems ok for me! Let the guys test in hardware if this is what they need and we are good to go. :100:

PR #766, opened 2024-08-01
[QHC-660] Add one qubit analog transpiler

visagim:


visagim:
Pylint is being unreasonable but other than that it's ready for review

src/qililab/analog/one_qubit_transpiler.py
```diff
@@ -0,0 +1,63 @@
...
+# limitations under the License.
+
+from typing import Any, Callable
+
+from qililab.analog import Parameter
+
+
+class Qubit2LevelTranspiler:  # pylint: disable=too-few-public-methods
+    """Implementation of the transpiler for the 2 level qubit. This is done mainly by inverting the same functions
+    used int he single_qubit_2level emulator model.
```
jjmartinezQT, line 22:
```suggestion
    used in the single_qubit_2level emulator model.
```

src/qililab/analog/one_qubit_transpiler.py
```diff
@@ -0,0 +1,63 @@
...
+        self.delta = Parameter(name="delta", set_method=self._set_delta)
+        # flux parameters
+        self.phiz = Parameter(name="phiz")
+        self.phix = Parameter(name="phix")
+
+    def __call__(self, delta: float, epsilon: float) -> tuple[Any, Any]:
+        """Transpiles Delta and Epsilon to phix, phiz"""
+        self.delta(delta)
+        self.epsilon(epsilon)
+        return self.phix(), self.phiz()  # type: ignore[func-returns-value]
```
jjmartinezQT, line 50: Just out of curiosity, what was the mypy issue before you included the type ignore here?

jjmartinezQT (approved):
Looks good to me. Great work on the parameter structure, eventually qililab should move into that direction.

src/qililab/analog/one_qubit_transpiler.py
visagim: It was complaining about `parameter()` not returning a value (function call does not return a value),  Now that I think about it maybe the issue was that I hadn't added type return to `parameter.get()`, I think it should work fine without the ignore now (trying this in the next commit)

jjmartinezQT: Yep, sounds right what you are saying.

src/qililab/analog/one_qubit_transpiler.py
```diff
@@ -0,0 +1,63 @@
...
+
+from qililab.analog import Parameter
+
+
+class Qubit2LevelTranspiler:  # pylint: disable=too-few-public-methods
+    """Implementation of the transpiler for the 2 level qubit. This is done mainly by inverting the same functions
+    used in the single_qubit_2level emulator model.
+
+    Args:
+        eps_model (Callable): epsilon model
```
fedonman, line 25: typo, should be `epsilon_model`.

src/qililab/analog/one_qubit_transpiler.py
```diff
@@ -0,0 +1,63 @@
...
+        delta_model (Callable): delta model
+        qubitData (DataClass): dataclass containing info about the physical parameters of the qubit. Defaults to None.
```
fedonman, line 27: Is this still relevant? I can't see the argument anywhere.

fedonman (commented):
Great work. :100: Added two comments on docstrings.

PR #767, opened 2024-08-01
[QHC-636] Create AnnealingSchedule class in Qililab.

visagim:


src/qililab/analog/annealing_program.py
```diff
@@ -0,0 +1,71 @@
...
+                anneal_step[circuit_element]["phiz"] = phiz
+
+    def get_waveforms(self) -> dict[str, tuple[Bus, ArbitraryWave]]:
+        """Returns a dictionary containing (bus, waveform) for each flux control from the transpiled fluxes. `AnnealingProgram.transpile` should be run first. The waveform is an arbitrary waveform obtained from the transpiled fluxes.
+
+        Returns:
+            dict[str,tuple[Bus,ArbitraryWave]]: Dictionary containing (bus, waveform) for each flux control (i.e. phix or phiz).
+        """
+        # parse names from algorithm notation (e.g. qubit_0) to runcard notation (e.g. q0)
+        element_name_map = {"qubit": "q", "coupler": "c"}
```
jjmartinezQT, line 52: this is enforcing a runcard standard for naming, no?

jjmartinezQT: I mean, what happens if the experimentalist that writes the runcard makes a consistent typo naming 'qubi' all along the runcard?

src/qililab/platform/platform.py
```diff
@@ -567,6 +590,42 @@ def __str__(self) -> str:
...
+        anneal_program = AnnealingProgram(self, anneal_program_dict)
+        anneal_program.transpile(transpiler)
+        anneal_waveforms = anneal_program.get_waveforms()
+
+        qp_anneal = QProgram()
+        with qp_anneal.average(averages):
+            for bus, waveform in anneal_waveforms.values():
+                qp_anneal.play(bus=bus.alias, waveform=waveform)
+
+        # TODO: define readout
```
jjmartinezQT, line 625: Should we include measurement as a parameter then for the execute function? what do you suggest? The question goes along where in the whole procedure we create an instance of the new measurement block and append it to this qprogram. I guess there is room for discussion with the team.

src/qililab/analog/annealing_program.py
visagim: yes true. I could also load the mappings from the runcard so that the standard is set at the runcard (`flux_to_bus_topology`) but some consistency from the user side will still be required when they send the anneal program dictionary however.

src/qililab/platform/platform.py
visagim: I think measurement needs to be defined as a qprogram block / standard procedure, since it is not just a single pulse (though we don't 100% know that yet) - but for the time being we can just add a normal measurement (i.e. `qp.meausre(...)` if we want things "to look pretty"

jjmartinezQT: Yes, I agree in that is a full block and not just a pulse, but the question for me is to when it is included as a parameter, in execute_anneal() or somewhere else.

src/qililab/analog/annealing_program.py
jjmartinezQT: that is a good point, worth explaining it in the changeloc and documentation, that the annealing schedule mapping needs to be consistent with runcard keys nomenclature.

visagim: let me clarify this so we can determine the best course of action. The way it's written there are 3 different strings:
- `qubit_0` which contains some attribute / dictionary with keys `phix`. This is used at the higher level because the structure of the entry point (from H algorithm) is of the kind `qubit_0` with some key `Delta` for the ising coefficients. So the transpiler just adds keys `phix` `phiz` under `qubit_0` to the pre-existing dictionary
- `phix_q0` single string used to point to a unique control line. Here we don't do the nested dict structure above because we want to allow flexibility in the runcard (i.e. in the real chip there might not be a "qubit" with unique lines per se, so we want to avoid grouping things under elements like qubit
- `flux_phix_q0` the name of the bus pointing to the `phix_q0` control.
The 1st and 2nd notations are enforced i.e. not following them will make the code break. The 3rd is completely arbitrary, the bus name can be whatever as in the gate based case (as long as there exists a bus with the same alias under `platform.bus`.

jjmartinezQT: Cool, let's include this clarifications in the changelog please, so ideally an user that digs into it gets a clear picture.

src/qililab/platform/platform.py
jjmartinezQT: As we do not know if the S-curve will even happen to be a reality, could we include here for the moment (September demo) a normal measurement pulse? It should be a preloaded pulse, either a square pulse or a flat top, so we should have flexibility on using a pre-calibrated pulse.

jjmartinezQT (changes requested):

src/qililab/analog/annealing_program.py
```diff
@@ -0,0 +1,85 @@
...
+        },      # time=0ns
+        {...},  # time=1ns
+    .
+    .
+    .
+    ]
+    """
+
+    platform: Any
+    anneal_program: list[dict[str, dict[str, float]]]
```
fedonman, line 40: If we know the format of each time step (sigma_x, sigma_y, sigma_z values) why don't we use a class to define each time step, so it's more readable and easy for the user to code it? Dictionaries are useful if we don't know the keys beforehand, but in other case OOP is preferred.

fedonman (changes requested):
Great work Victor! :100: Requested a couple of changes, mainly on OOP concepts, and naming conventions.

src/qililab/qprogram/qprogram.py
```diff
@@ -101,6 +102,54 @@ def __init__(self) -> None:
...
+                temp_dict["weights"] = [  # type: ignore[assignment]
+                    (type(program.weights.I).__name__, program.weights.I.envelope()),
+                    (type(program.weights.Q).__name__, program.weights.Q.envelope()),
+                ]
+
+        for key, value in temp_dict.copy().items():
+            if "UUID" in str(value):
+                temp_dict[key] = None
+
+        return temp_dict
```
fedonman, line 151: Also, why do we remove several fields like `UUID`? `__dict__` should return the dictionary representation of the class as it is. In other words we could create a `from_dict()` method that takes this dictionary and correctly builds the class instance again. This will not be the case with this implementation.

src/qililab/analog/annealing_program.py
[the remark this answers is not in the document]
visagim: I chose to use a dictionary because i'm assuming that ultimately the annealing schedule is something that's gonna come from an upper layer outside qililab (either from the algorithms team library or from a notebook, etc. ), so I wanted to go for some easy to convert format that's serializable / can be sent through the net, easy to store, etc. So for that the dictionary type seemed ideal. Once it's more clear what the entry point (how we convert from a qibo hamiltonian to this) we can implement some more rigid class if relevant

src/qililab/analog/annealing_program.py
```diff
@@ -0,0 +1,85 @@
...
+         "coupler_1_0 : {...},
+        },      # time=0ns
+        {...},  # time=1ns
+    .
+    .
+    .
+    ]
+    """
+
+    platform: Any
```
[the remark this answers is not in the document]
visagim: The annealing program might be accessed from outside through the class object. Platform won't.

src/qililab/platform/platform.py
```diff
@@ -567,6 +591,41 @@ def __str__(self) -> str:
         """
         return str(YAML().dump(self.to_dict(), io.BytesIO()))
 
+    def execute_anneal_program(
```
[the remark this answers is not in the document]
visagim: Hmm yes there's some degeneracy in the name usage sorry about that, I'll change it to be consistent. `anneal` is not a shortcut of `annealing` here though, I was just using it as the action of "performing an anneal". I'll check out which is more correct ("running an anneal" vs "running an annealing schedule")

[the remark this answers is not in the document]
visagim: Hmm yes there's some degeneracy in the name usage sorry about that, I'll change it to be consistent. `anneal` is not a shortcut of `annealing` here though, I was just using it as the action of "performing an anneal". I'll check out which is more correct ("running an anneal" vs "running an annealing schedule")

src/qililab/platform/platform.py
```diff
@@ -567,6 +591,41 @@ def __str__(self) -> str:
...
+        self, anneal_program_dict: list[dict[str, dict[str, float]]], transpiler: Callable, averages=1
```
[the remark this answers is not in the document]
visagim: Refer to https://github.com/qilimanjaro-tech/qililab/pull/767#discussion_r1713447189. Ofc this might be addressed in a separate pr building on top of this.

> In QProgram, the parameter for the number of averages is called shots, following other frameworks that use the same name. Should we rename averages to shots this to be consistent?

Using shots when we are actually running averages is misleading. Other frameworks allow you to recover statistics about shots (perhaps not all of them) which cannot be accessed when running averages (eg standard deviation). I think it's much more clear to name it average since we are actually running an average, and not retrieving each individual "shot"

src/qililab/qprogram/qprogram.py
```diff
@@ -101,6 +102,54 @@ def __init__(self) -> None:
         self._variables: list[Variable] = []
         self._block_stack: deque[Block] = deque([self._body])
 
+    def __dict__(self) -> dict:  # type: ignore[override]
+        """Dictionary representation"""
+        return self._to_dict(self)
+
+    def _to_dict(self, program) -> dict:
```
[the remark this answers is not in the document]
visagim: Implementing the method as `__dict__` crashed many other parts of the workflow (outside of this pr). I just wanted to implement a method to have a readable qprogram string, for the moment I avoided clashes with other parts of the workflow but if we like this method we can integrate it more in the future.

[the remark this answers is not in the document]
fedonman: If we need a readable string, then `__str__` is the method that needs implementation.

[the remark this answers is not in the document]
fedonman: The best way btw to do it is to implement `__str__` to all operations and blocks and call `str(element)` when you traverse the qprogram.

src/qililab/qprogram/qprogram.py
fedonman: I also commented above. If we need the strings we should implement the `__str__` method, which is ok to hide information and display only the useful ones. `__dict__` by python documentation should return all attributes of the class.

src/qililab/analog/annealing_program.py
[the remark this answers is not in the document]
fedonman: Then it's better to define both `_platform` and `_anneal_program` as private, and  provide a public attribute `anneal_program` to read-only the annealing program (but not set it, since this is done in `__init__` only)

src/qililab/analog/annealing_program.py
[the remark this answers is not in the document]
fedonman: I understand what you are saying, it's probably unlikely that someone will create a schedule manually, but still each layer should be independent.

Also, having a concrete class with fields will help the algorithms team as well, since they will have a reference. (Imagine renaming "sigma_x" or altering something, right now we would have to manually relay the change to everyone)

Anyway, this is not a hard requirement, but I think the change will help the whole implementation of the flow.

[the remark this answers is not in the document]
visagim: I've created an issue for this at https://linear.app/qilimanjaro/issue/QHC-689/create-entry-point-class-for-annealingprogram-instead-of-dictionary

src/qililab/platform/platform.py
visagim: will be dealt here to avoid making this pr too big
https://linear.app/qilimanjaro/issue/QHC-691/add-default-measurement-pulse-to-annealing-program

jjmartinezQT: Agree.

jjmartinezQT (approved):

src/qililab/analog/annealing_program.py
```diff
@@ -0,0 +1,93 @@
...
+    Args:
+        platform (Any): platform
+        annealing_program (list[dict[str, dict[str, float]]]): dictionary with the annealing program with the above structure.
+    """
+
+    def __init__(self, platform: Any, annealing_program: list[dict[str, dict[str, float]]]):
+        """Init method"""
+        self._platform = platform
+        self._annealing_program = annealing_program
+        self.annealing_program = annealing_program  # TODO: implement as frozenDataclass
```
fedonman, line 48: why do we have a duplicate attribute here?

fedonman: If this is meant to be the public property then we should use
```
@property
def annealing_program(self)
```

fedonman (commented):
Added a new comment and there still one unresolved question from previous review and is good to go :100: :rocket:

src/qililab/platform/platform.py
```diff
@@ -567,6 +600,46 @@ def __str__(self) -> str:
...
+        from the bus to flux mapping given by the runcard.
+
+        Args:
+            annealing_program_dict (list[dict[str, dict[str, float]]]): annealing program to run
+            transpiler (Callable): ising to flux transpiler. The transpiler should take 2 values as arguments (delta, epsilon) and return 2 values (phix, phiz)
+            averages (int, optional): Amount of times to run and average the program over. Defaults to 1.
+        """
+        annealing_program = AnnealingProgram(self, annealing_program_dict)
+        annealing_program.transpile(transpiler)
+        annealing_waveforms = annealing_program.get_waveforms()
```
fedonman, line 632: Also, why have we chosen the transpiler to be a Callable and not a class? Now the method `transpile()` belongs to `AnnealingProgram` when it should belong to the `Transpiler` class.

Each class should have a single responsibility, and we should thrive to have a similar structure as circuit and qprogram: the Transpiler class receives an AnnealingProgram and transpiles it.

visagim: I don't see a problem moving it into a class in the future when we need some more complicated functionality for the transpilaiton process. Right now `AnnealingProgram.transpile()` is just calling the transpiler (`transpiler: Callable` in the arguments) at each time step (so the transpiler _is_ a different class, this is just calling it to transpile the program)

src/qililab/analog/annealing_program.py
```diff
@@ -0,0 +1,93 @@
...
+        circuit_element_map = {
+            (element, flux): f"{flux}_{element_name_map[element.split('_', 1)[0]]}{element.split('_', 1)[1]}"
+            for element in self._annealing_program[0].keys()
+            for flux in self._annealing_program[0][element].keys()
+            if "phi" in flux
+        }  # {(element, flux): flux_line}
+
+        # Initialize dictionary with flux_lines pointing to (corresponding bus, waveform)
+        annealing_waveforms = {  # type: ignore[var-annotated]
+            flux_line: (self._platform.get_element(flux_line), []) for flux_line in circuit_element_map.values()
```
fedonman, line 84: Another important thing I just saw: Since we just need the Platform to retrieve the mapping of flux lines, we should avoid having this dependency.

Instead the parameter at the `__init__` method should be this mapping.

Also, if the `transpile()` method moves to the `Transpiler` class this dependency moves away from the `AnnealingProgram` completely.

Which makes sense, since our Programs (circuit / qprogram / annealing) are just representations and the compilation / execution are handled by other classes.

visagim:
> Added a new comment and there still one unresolved question from previous review and is good to go 💯 🚀

what's the unresolved comment?

fedonman:
> > Added a new comment and there still one unresolved question from previous review and is good to go 💯 🚀
> 
> what's the unresolved comment?

this, you can find it if you scroll higher.

image

src/qililab/analog/annealing_program.py
visagim: There are 2 different transpilers here (technically). One is the one that handles ising to fluxes (this is the one I call transpiler), the other one handles fluxes to amplitudes (waveforms). We should ensure modularity of these 2 since they belong to 2 different workflows, even if they are both supposed to get data from calibration (but they should absolutely be run independently).
The `get_element` method is the standard method to get information for platform objects. Getting the mapping as an object would be better but that would require parsing the class input as a previous step in order to get the `{flux_control_line : bus}` dictionary (note that the current input expects `{chip_element : {ising_coeff : value}}`
This is related to https://github.com/qilimanjaro-tech/qililab/pull/767#discussion_r1713813800 probably; again we can standarize the input once we have more information about the upper stack, I don't see the point in doing this now

src/qililab/platform/platform.py
fedonman: `Callable` is not a class though, its a function.

Question: How do we expect users to create/use these transpilers? Do we have a number of predefined algorithms for transpilation?

src/qililab/platform/platform.py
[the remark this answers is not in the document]
visagim: Oh maybe I messed up the answers, but this was answered here I believe https://github.com/qilimanjaro-tech/qililab/pull/767#discussion_r1713813800 (tldr added a task for this)
Also related to this https://github.com/qilimanjaro-tech/qililab/pull/767#discussion_r1716738156 (tldr yes we should do that, I wouldnt do it this sprint though)

visagim:
Also a further note on this
> Also, why have we chosen the transpiler to be a Callable and not a class? Now the method `transpile()` belongs to `AnnealingProgram` when it should belong to the `Transpiler` class.

The flux to ising transpiler is very much at its infancy still, and there is a lot of uncertainty about how will it evolve further than the one qubit case. That is why I am reluctant about starting to make classes and big structures for a workflow that is sure to change in the short term (1-2months) - I think it's better to keep something minimal and make it more rigid (efficient) as we consolidate each step and are sure that that's how it's going to be in the mid term (5-10months).
But I do believe at some point we'll definitely make a transpiler class with more complicated logic, just think that right now it's overkill

src/qililab/platform/platform.py
visagim: Yes, right now the transpiler is a callable class (we have only a one qubit simple implementation though) https://github.com/qilimanjaro-tech/qililab/blob/main/src/qililab/analog/one_qubit_transpiler.py that takes other functions (methods from fittings, stored in a calibration / transpiler_calibration file, see an example at https://github.com/qilimanjaro-tech/qili-fluqe/blob/f7fe294d2b69b1515f9cbcd87f89523d6ef93bd2/notebooks/single_qubit_2level_algorithm_demo.ipynb)

visagim: Again, this is kept quite open because we have yet to test any of this in hardware

fedonman:
> Also a further note on this
> 
> > Also, why have we chosen the transpiler to be a Callable and not a class? Now the method `transpile()` belongs to `AnnealingProgram` when it should belong to the `Transpiler` class.
> 
> The flux to ising transpiler is very much at its infancy still, and there is a lot of uncertainty about how will it evolve further than the one qubit case. That is why I am reluctant about starting to make classes and big structures for a workflow that is sure to change in the short term (1-2months) - I think it's better to keep something minimal and make it more rigid (efficient) as we consolidate each step and are sure that that's how it's going to be in the mid term (5-10months). But I do believe at some point we'll definitely make a transpiler class with more complicated logic, just think that right now it's overkill

It's not an overkill if we know that we will need it eventually. Making the change after 2+ months of additional development will be much more difficult than it is to make it now.

fedonman:
> > Also a further note on this
> > > Also, why have we chosen the transpiler to be a Callable and not a class? Now the method `transpile()` belongs to `AnnealingProgram` when it should belong to the `Transpiler` class.
> > 
> > 
> > The flux to ising transpiler is very much at its infancy still, and there is a lot of uncertainty about how will it evolve further than the one qubit case. That is why I am reluctant about starting to make classes and big structures for a workflow that is sure to change in the short term (1-2months) - I think it's better to keep something minimal and make it more rigid (efficient) as we consolidate each step and are sure that that's how it's going to be in the mid term (5-10months). But I do believe at some point we'll definitely make a transpiler class with more complicated logic, just think that right now it's overkill
> 
> It's not an overkill if we know that we will need it eventually. Making the change after 2+ months of additional development will be much more difficult than it is to make it now.

My main concern is that this is public API since it is in `platform.execute_annealing()` and it will be much difficult to change the API once we start using it.

If you answer the other question above, `How do we expect users to create/use these transpilers? Do we have a number of predefined algorithms for transpilation?`, we can try to find an minimal OOP design that provides a friendly UX API.

My current guess is an `AnnealingTranspiler` generic class, that can be sub-classed to provide different pre-defined algorithms, or also initialized using a custom `Callable` as it is now.

That way the signature of `platform.execute_annealing()` would be `(annealing_program: AnnealingProgram, transpiler: AnnealingTranspiler)` which makes it more UX friendly than `(anneal_program_dict: list[dict[str, dict[str, float]]], transpiler: Callable)`

src/qililab/platform/platform.py
jjmartinezQT: > `Callable` is not a class though, its a function.
> 
> Question: How do we expect users to create/use these transpilers? Do we have a number of predefined algorithms for transpilation?

We should assume for now that on a higher level someone will want to have a predefined algorithm yes, that being the user running the demo. Will have a deeper answer on the full user story by the end of the week, but we should also consider experimental users wanting to use this building up all the necessary pieces without having a pre-defined complete algorithm.

src/qililab/analog/annealing_program.py
fedonman: > There are 2 different transpilers here (technically). One is the one that handles ising to fluxes (this is the one I call transpiler), the other one handles fluxes to amplitudes (waveforms). We should ensure modularity of these 2 since they belong to 2 different workflows, even if they are both supposed to get data from calibration (but they should absolutely be run independently).

This makes me want the usage of classes even more!

> The `get_element` method is the standard method to get information for platform objects. Getting the mapping as an object would be better but that would require parsing the class input as a previous step in order to get the `{flux_control_line : bus}` dictionary.

Yes, I am not against using `get_element()`, this is the correct way indeed. I am against having a dependency on platform to build an `AnnealingProgram` instance. The annealing program should be independent of the platform. If it needs a mapping `{flux_control_line : bus}` information then this should be the parameter of the constructor, not the whole platform.

There is a similar situation in `QbloxCompiler` where we need the `time_of_flight` for each bus, which we get from platform. The signature of the `compile()` method expects just that: `times_of_flight: dict[str, int] | None = None`. The user might create the dictionary manually (providing total flexibility) or in the case that you execute qprogram from `platform.execute_qprogram`, we retrieve the mapping from platform:

```
times_of_flight = {
                bus.alias: int(bus.get_parameter(Parameter.TIME_OF_FLIGHT))
                for bus in buses
                if isinstance(bus.system_control, ReadoutSystemControl)
            }
```

src/qililab/platform/platform.py
fedonman: Oh now I get it from seeing fluqe code. You do have a class indeed, and you implement the `__call__` method so it is also a `Callable`.

OK, then this is acceptable for now.

fedonman: But we should add a `TODO` to discuss this transpilers structure at some point.

visagim:
I added a ticket to address (I believe) all of the suggestions above and marked related conversations as resolved (as I believe they have been). Once @fedonman gives the OK I'll merge

PR #772, opened 2024-08-06
[QHC-666] Execute qprogram docstring

GuillermoAbadLopez:
Screenshot 2024-08-06 at 14 52 01

### DIFF
--- a/src/qililab/platform/platform.py
+++ b/src/qililab/platform/platform.py
@@ -574,7 +574,11 @@ class Platform:  # pylint: disable = too-many-public-methods, too-many-instance-
         calibration: Calibration | None = None,
         debug: bool = False,
     ) -> QProgramResults:
-        """Execute a QProgram using the platform instruments. The execution is done in the following steps:
+        """Execute a QProgram using the platform instruments.
+
+        |
+
+        **The execution is done in the following steps:**
 
         1. Compile the QProgram.
         2. Upload the sequences to the instruments.
@@ -582,7 +586,9 @@ class Platform:  # pylint: disable = too-many-public-methods, too-many-instance-
         4. Acquire the results.
         5. Reset the instrument settings.
 
-        The execution is done in the following way:
+        |
+
+        **The execution is done in the following way:**
 
         - If all the buses in the QProgram are associated with Qblox modules, the execution is done using the QbloxCompiler.
         - If all the buses in the QProgram are associated with Quantum Machines Clusters, the execution is done using the QuantumMachinesCompiler.
--- a/src/qililab/platform/platform.py
+++ b/src/qililab/platform/platform.py
@@ -588,9 +588,9 @@ class Platform:  # pylint: disable = too-many-public-methods, too-many-instance-
 
         **The execution is done in the following way:**
 
-        - If all the buses in the :class:`.QProgram` are associated with Qblox modules, the execution is done using the :class:`.QbloxCompiler`.
-        - If all the buses in the :class:`.QProgram` are associated with Quantum Machines Clusters, the execution is done using the :class:`.QuantumMachinesCompiler`.
-        - If the buses in the :class:`.QProgram` are associated with a mixture of Qblox modules and Quantum Machines Clusters, the execution is not supported.
+        - If all the buses in the :class:`.QProgram` are associated with Qblox modules, the compilation is done using the :class:`.QbloxCompiler`.
+        - If all the buses in the :class:`.QProgram` are associated with Quantum Machines Clusters, the compilation is done using the :class:`.QuantumMachinesCompiler`.
+        - If the buses in the :class:`.QProgram` are associated with a mixture of Qblox modules and Quantum Machines Clusters, the compilation is not supported.
 
         Args:
             qprogram (QProgram): The :class:`.QProgram` to execute.
--- a/src/qililab/platform/platform.py
+++ b/src/qililab/platform/platform.py
@@ -588,7 +588,7 @@ class Platform:  # pylint: disable = too-many-public-methods, too-many-instance-
 
         **The execution can be done for (buses associated to) two different type of clusters:**
 
-        - For ``Qblox`` modules, the compilation is done using the :class:`.QbloxCompiler`. Which compiles the :class:`.QProgram` into``Q1ASM`` for multiple sequencers based on each bus, uploads and executes the sequences (desynching buses if necessary), and acquires the results.
+        - For ``Qblox`` modules, the compilation is done using the :class:`.QbloxCompiler`. Which compiles the :class:`.QProgram` into``Q1ASM`` for multiple sequencers based on each bus, uploads and executes the sequences, and acquires the results.
         - For ``Quantum Machines`` clusters, the compilation is done using the :class:`.QuantumMachinesCompiler`. This compiler transforms the :class:`.QProgram` into ``QUA``, the programming language of ``Quantum Machines`` hardware. It then executes the resulting ``QUA`` program and returns the results, organized by bus.
 
         Args:
--- a/src/qililab/platform/platform.py
+++ b/src/qililab/platform/platform.py
@@ -600,7 +600,7 @@ class Platform:  # pylint: disable = too-many-public-methods, too-many-instance-
             debug (bool, optional): Whether to print debug information. Defaults to False.
 
         Returns:
-            QProgramResults: The results of the execution. ``QProgramResults.results()`` returns a dictionary (`dict[str, list[Result]]`) of measurement results.
+            QProgramResults: The results of the execution. ``QProgramResults.results()`` returns a dictionary (``dict[str, list[Result]]``) of measurement results.
                 The keys correspond to the buses a measurement were performed upon, and the values are the list of measurement results in chronological order.
         """
         bus_aliases = {bus_mapping[bus] if bus_mapping and bus in bus_mapping else bus for bus in qprogram.buses}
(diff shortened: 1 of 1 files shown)


src/qililab/platform/platform.py
```diff
@@ -576,11 +576,34 @@ def execute_qprogram(  # pylint: disable=too-many-locals
...
+
+        **The execution is done in the following way:**
+
+        - If all the buses in the QProgram are associated with Qblox modules, the execution is done using the QbloxCompiler.
+        - If all the buses in the QProgram are associated with Quantum Machines Clusters, the execution is done using the QuantumMachinesCompiler.
+        - If the buses in the QProgram are associated with a mixture of Qblox modules and Quantum Machines Clusters, the execution is not supported.
+
         Args:
             qprogram (QProgram): The QProgram to execute.
+            bus_mapping (dict[str, str], optional): A dictionary mapping the buses in the QProgram to the buses in the platform.
```
jordivallsq: The statement is correct but the concepts are 2 sentences that repeat each other.
bus_mapping (dict[str, str], optional): A dictionary mapping the buses in the QProgram (keys) to the buses in the platform (values). It is useful for mapping specific experiments to generic Qprograms. Defaults to None.
something like this?

src/qililab/platform/platform.py
```diff
@@ -576,11 +576,34 @@ def execute_qprogram(  # pylint: disable=too-many-locals
...
 
+        |
+
+        **The execution is done in the following steps:**
+
+        1. Compile the QProgram.
+        2. Upload the sequences to the instruments.
+        3. Execute the sequences.
+        4. Acquire the results.
+        5. Reset the instrument settings.
```
jordivallsq: One weird thing I found is the 5th point:
5. Reset the instrument settings.
as this shouldn't be in the execute, more like inside execute_Qblox as it is the source of this last point. I suggest leaving it out and then specifying in each machine what is resetted.

jordivallsq: If execute_Qblox is an internal method then we should explain here the two clusters Qprogram is able to work with and explain what are we doing on each just here. Maybe two more detailed lists should be better, one for each machine

src/qililab/platform/platform.py
```diff
@@ -576,11 +576,34 @@ def execute_qprogram(  # pylint: disable=too-many-locals
...
+
+        - If all the buses in the QProgram are associated with Qblox modules, the execution is done using the QbloxCompiler.
+        - If all the buses in the QProgram are associated with Quantum Machines Clusters, the execution is done using the QuantumMachinesCompiler.
+        - If the buses in the QProgram are associated with a mixture of Qblox modules and Quantum Machines Clusters, the execution is not supported.
+
         Args:
             qprogram (QProgram): The QProgram to execute.
+            bus_mapping (dict[str, str], optional): A dictionary mapping the buses in the QProgram to the buses in the platform.
+                The keys are the buses in the QProgram, and the values are the buses in the platform. Defaults to None.
+            calibration (Calibration, optional): The calibration to use for the execution. Defaults to None.
```
jordivallsq, line 599: This exaplanation is too broad, this calibration makes reference to the yaml file containing information of the experiment (i.e. pulse duration, amp...)

I would say something like:
calibration (Calibration, optional): Calibration file containing information of te experiment in the YAML format. Defaults to None.

jordivallsq (commented):
These are some of my comments

src/qililab/platform/platform.py
GuillermoAbadLopez: I like it, changed it!
Thanks 🔥

src/qililab/platform/platform.py
GuillermoAbadLopez: You are right, I'll delete the case concrete bits, and leave it general, 
if you feel like doing the separation, please free and commit it yourself 🙌

src/qililab/platform/platform.py
GuillermoAbadLopez: Cool! I'll change it, thanks!

src/qililab/platform/platform.py
```diff
@@ -574,13 +574,34 @@ def execute_qprogram(  # pylint: disable=too-many-locals
...
+        2. Run the compiled QProgram.
+        3. Acquire the results.
+
+        |
+
+        **The execution is done in the following way:**
+
+        - If all the buses in the :class:`.QProgram` are associated with Qblox modules, the compilation is done using the :class:`.QbloxCompiler`.
+        - If all the buses in the :class:`.QProgram` are associated with Quantum Machines Clusters, the compilation is done using the :class:`.QuantumMachinesCompiler`.
+        - If the buses in the :class:`.QProgram` are associated with a mixture of Qblox modules and Quantum Machines Clusters, the compilation is not supported.
```
jordivallsq, line 593: I would replace this part (as I think that it is not saying much) for a more specific comment regarding the specific operations inside Qblox and QM

```suggestion
        **execute_qprogram operates with two different clusters independently:**

        - For buses assosiated with Qblox modules, the compilation is done using the :class:`.QbloxCompiler`. This compiler compiles the Qprogram in Q1ASM into multiple sequencers based on each bus, uploads and executes the sequences, acquires the results and resets the settings by desyncing each bus.
        - For buses assosiated with Quantum Machines Clusters, the compilation is done using the :class:`.QuantumMachinesCompiler`. This compiler transforms the Qprogram in QUA, the compilation language for Quantum Machines, runs the job created by the QUA execution and returns the results ordering them by bus.
```
This can be definetly improved as each person has different ways to explain themselves
BTW I really like the :class:`.something`.

src/qililab/platform/platform.py
```diff
@@ -574,13 +574,34 @@ def execute_qprogram(  # pylint: disable=too-many-locals
...
 
         Args:
-            qprogram (QProgram): The QProgram to execute.
+            qprogram (QProgram): The :class:`.QProgram` to execute.
+            bus_mapping (dict[str, str], optional): A dictionary mapping the buses in the :class:`.QProgram` (keys )to the buses in the platform (values).
+                It is useful for mapping specific experiments to generic :class:`.QProgram` 's. Defaults to None.
+            calibration (Calibration, optional): :class:`.Calibration` file containing information of the experiment in the YAML format. Defaults to None.
+            debug (bool, optional): Whether to print debug information. Defaults to False.
```
jordivallsq, line 600: I would expand the debug a bit
```suggestion
            debug (bool, optional): Whether to create debug information. For QBlox cluster all the program information is printed on screen. For Quantum Machines cluster a .py file is created containing the QUA and config compilation. Defaults to False.
```

jordivallsq (commented):
I added a couple of comments but I suggest that @fedonman checks that as he might provide better insight an opinion on what should be written here

src/qililab/platform/platform.py
```diff
@@ -574,13 +574,34 @@ def execute_qprogram(  # pylint: disable=too-many-locals
...
+
+        1. Compile the QProgram.
+        2. Run the compiled QProgram.
+        3. Acquire the results.
+
+        |
+
+        **The execution can be done for (buses associated to) two different type of clusters:**
+
+        - For ``Qblox`` modules, the compilation is done using the :class:`.QbloxCompiler`. Which compiles the :class:`.QProgram` into``Q1ASM`` for multiple sequencers based on each bus, uploads and executes the sequences, acquires the results and resets the settings by desynching each bus.
```
fedonman, line 591: `resets the settings`: This is not true. The settings of the instruments are not touched. The only setting that is changed is the syncing between buses that are involved in the qprogram. But this is only during execution.

src/qililab/platform/platform.py
```diff
@@ -574,13 +574,34 @@ def execute_qprogram(  # pylint: disable=too-many-locals
...
+        - For ``Quantum Machines`` clusters, the compilation is done using the :class:`.QuantumMachinesCompiler`. Which transforms the :class:`.QProgram` into ``QUA``, the compilation language for ``Quantum Machines``, runs the job created by the ``QUA`` execution and returns the results ordering them by bus.
```
fedonman, line 592: Grammar error, `which` cannot start a sentence.

```suggestion
        - For ``Quantum Machines`` clusters, the compilation is done using the :class:`.QuantumMachinesCompiler`. This compiler transforms the :class:`.QProgram` into ``QUA``, the programming language of ``Quantum Machines`` hardware. It then executes the resulting QUA program and returns the results, organized by bus.
```

src/qililab/platform/platform.py
```diff
@@ -574,13 +574,34 @@ def execute_qprogram(  # pylint: disable=too-many-locals
...
 
         Args:
-            qprogram (QProgram): The QProgram to execute.
+            qprogram (QProgram): The :class:`.QProgram` to execute.
+            bus_mapping (dict[str, str], optional): A dictionary mapping the buses in the :class:`.QProgram` (keys )to the buses in the platform (values).
+                It is useful for mapping specific experiments to generic :class:`.QProgram` 's. Defaults to None.
```
fedonman, line 597: It's the other way around.

```suggestion
                It is useful for mapping a generic :class:`.QProgram` to a specific experiment. Defaults to None.
```

src/qililab/platform/platform.py
```diff
@@ -574,13 +574,34 @@ def execute_qprogram(  # pylint: disable=too-many-locals
...
+            calibration (Calibration, optional): :class:`.Calibration` file containing information of the experiment in the YAML format. Defaults to None.
```
fedonman, line 598:
```suggestion
            calibration (Calibration, optional): :class:`.Calibration` instance containing information of previously calibrated values like waveforms, weights and crosstalk matrix. Defaults to None.
```

fedonman (commented):
Awesome job! :100: There are lots more to be added (Ramiro wants a detailed documentation of the execution flow), but it's a very good start. Added some comments.

src/qililab/platform/platform.py
GuillermoAbadLopez: What do you think about the new change:
```
- For ``Qblox`` ... and executes the sequences (desynching buses if necessary), and acquires the results.
```

GuillermoAbadLopez: I can also just delete de parenthesis if you prefer 🙌

fedonman: If we don't mention the syncing part before execution, then we shouldn't mention the desyncing afterwards, since it is confusing. So without the parentheses is better I think.

fedonman (approved):
Answered in the last unresolved discussion. Other than that, it's good to go! Nice work :100:

PR #774, opened 2024-08-07
[QHC-631] Implement Crosstalk Matrix Features

fedonman:


docs/releases/changelog-dev.md
```diff
@@ -2,6 +2,31 @@
...
 
+- Added `CrosstalkMatrix` class to represent and manipulate a crosstalk matrix, where each index corresponds to a bus. The class includes methods for initializing the matrix, getting and setting crosstalk values, and generating string representations of the matrix.
+
+  Example:
+
+  ```Python
+  # Create an empty crosstalk matrix
+  crosstalk_matrix = CrosstalkMatrix()
+
+  # Add crosstalk values
```
jjmartinezQT, line 13:
```suggestion
  # Add crosstalk values, where the keys are in matrix shape [row][column]
```

src/qililab/qprogram/crosstalk_matrix.py
```diff
@@ -0,0 +1,96 @@
...
+            for bus2 in sorted_buses:
+                if bus1 == bus2:
+                    row.append("\\".rjust(col_width))
+                else:
+                    row.append(f"{self.matrix.get(bus1, {}).get(bus2, 1.0):{col_width}.1f}")
+            rows.append(" ".join(row))
+        return header + "\n".join(rows)
+
+    @classmethod
+    def from_buses(cls, buses: Collection[str]) -> "CrosstalkMatrix":
```
jjmartinezQT, line 84: The only thing I am missing in this class is a method that returns a matrix in numpy way. But I am not 100% sure that is how @visagim  will prefer it, to perform the linear algebra matrix multiplication, or retrieving it indexed by buses works better for him. I do not mind merging it like this and plan a new task to include that new functionality if Victor tell us he needs it.

jjmartinezQT (changes requested):

src/qililab/qprogram/crosstalk_matrix.py
fedonman: Yes, I didn't implement this method, since I am not aware of the way that buses will be mapped to the matrix indexes (e.g. `drive_q0_bus` gets assigned to index 0 etc.) I would prefer to merge it as it is, and discuss this mapping in the next sprint with Victor.

jjmartinezQT: Sounds good to me.

jjmartinezQT (approved):
Good work.

PR #775, opened 2024-08-12
[QHC-670]Add xtalk compensation matrix to annealing program workflow

visagim:


visagim:
Ready for review while I fix code quality issues

src/qililab/analog/annealing_program.py
```diff
@@ -41,11 +42,15 @@ class AnnealingProgram:
...
     """
 
-    def __init__(self, platform: Any, annealing_program: list[dict[str, dict[str, float]]]):
+    def __init__(
+        self,
+        flux_to_bus_topology: list[Runcard.FluxControlTopology],
+        annealing_program: list[dict[str, dict[str, float]]],
+    ):
         """Init method"""
-        self._platform = platform
```
jjmartinezQT, line 46: we really need docstrings here for the flux topology.

src/qililab/analog/annealing_program.py
```diff
@@ -57,37 +62,62 @@ def transpile(self, transpiler: Callable):
...
-                annealing_step[circuit_element]["phix"] = phix
-                annealing_step[circuit_element]["phiz"] = phiz
+                transpiled_step[f"phix_{self._chip_element_to_short(chip_element)}"] = phix
+                transpiled_step[f"phiz_{self._chip_element_to_short(chip_element)}"] = phiz
 
-    def get_waveforms(self) -> dict[str, tuple[Bus, ArbitraryWave]]:
+            self._transpiled_program.append(transpiled_step.copy())
+
+    def _chip_element_to_short(self, chip_element: str) -> str:
+        """Parse names from algorithm notation (e.g. qubit_0) to runcard notation (e.g. q0)
```
jjmartinezQT, line 76: Is this not enforcing a runcard notation? What if the runcard element is defined as "q_0"? If we are enforcing a notation this should be documented and properly explained and circulated in the appropiate channels. Changelog is failing to reflect this.

src/qililab/analog/annealing_program.py
```diff
@@ -57,37 +62,62 @@ def transpile(self, transpiler: Callable):
...
+                        f"More than one flux pointing at bus {flux_bus.bus} in the runcard flux to bus topology"
+                    )
+                bus_to_flux_map[flux_bus.bus] = flux_bus.flux
+        flux_to_bus_map = {v: k for k, v in bus_to_flux_map.items()}
+
+        # Initialize annealing waveforms
+        annealing_waveforms = {bus: [] for bus in bus_to_flux_map}  # type: ignore[var-annotated]
+        # get xtalk matrix
+        if correct_xtalk:
+            # TODO: where do we load the CrosstalkMatrix from?
```
jjmartinezQT, line 112: Crosstalk matrix loads from calibration file. Is that what you are asking? If not please clarify what that todo really mean.

jjmartinezQT (changes requested):

src/qililab/analog/annealing_program.py
jjmartinezQT: And also merge latest from main as there was a bug related to flux_topology and runcard serialization that Guille did a hotfix for.

src/qililab/qprogram/crosstalk_matrix.py
```diff
@@ -14,9 +14,51 @@
 
 from typing import Collection
 
+import numpy as np
+
 from qililab.yaml import yaml
 
 
+class FluxVector:
```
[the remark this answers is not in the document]
visagim: Pretty much, for the time being this is only used for the xtalk. But it's better to have an object for the `__matmul__` method rather than just a typed dictionary, this was something we agreed with @fedonman (correct me if wrong)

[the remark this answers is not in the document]
jjmartinezQT: Yeah, I agree with the objectification.

src/qililab/analog/annealing_program.py
visagim: This is already in the changelog for the annealing program pr   [#767](https://github.com/qilimanjaro-tech/qililab/pull/767), but if that's not clear we can perhaps modify it / emphasize it on the docstring. We could also look into other ways to enforce some notation / alternatives to this from the code side. This will most likely be dealt with at https://linear.app/qilimanjaro/issue/QHC-694/objectify-analog-program-execution-workflow

src/qililab/analog/annealing_program.py
visagim: > Is that what you are asking?

Yes. I don't know the workflow to load the calibration matrix and how to include it here. If loaded from platform, maybe we can just add the xtalk matrix from argument when we call `get_waveforms`  instead of the boolean `correct_xtalk` (or pass a `Calibration`  object, i dont know)

jjmartinezQT: passing a calibration object I think should be the way, it is mandatory for the measurement anyway at the moment, and it should remain mandatory. So when we receive an execute_annealing, we are getting the calibration file as parameter, that should help you with this doubt, no?

src/qililab/analog/annealing_program.py
jjmartinezQT: We should circulate it more, no need for now for extra ways to enforce, but we already introduced some bug in main because not everyone was aware of the flux topology being in the runcard and failed to serialize.

src/qililab/platform/platform.py
```diff
@@ -640,14 +641,16 @@ def execute_anneal_program(
             averages (int, optional): Amount of times to run and average the program over. Defaults to 1.
         """
         if calibration.has_waveform(bus=readout_bus, name=measurement_name):
-            annealing_program = AnnealingProgram(self, annealing_program_dict)
+            annealing_program = AnnealingProgram(
```
jjmartinezQT, line 644: here at this point you already have the calibration class created to use as a parameter to the class AnnealingProgram.

src/qililab/analog/annealing_program.py
jjmartinezQT: Check my comment in platform to see that you already have this object loaded previously for you.

src/qililab/platform/platform.py
visagim: It is true that we require a `Calibration` for `execute_anneal_program` but so far this was not a requirement for `AnnealingProgram` itself (just considering modularity). I think it'd be better to require a `CrosstalkMatrix` from `get_waveforms`  and for `execute_anneal_program`  to load that xtalk matrix from the calibration file (just like we do for calibrated measurements)

visagim: That way if someone wants to run methods from `AnnealProgram` as standalone they can just create their own xtalk matrix without needing to load a calibration file

jjmartinezQT: Whatever you think is best, but in the execute_anneal you have the parameter already if you need to use it, how you propagate it until the point of usage from there is to you to propose. But as it is now, you can already use a calibration object at that point.

jjmartinezQT: What I want to say is, do not leave a todo hanging in there if you already have a way to solve it. This should be a complete PR as much as possible. If in the future we re-visit decisions is ok, but do not leave a PR uncompleted to have yet another one to finish it.

visagim: It was left as it is because at the time we did not have a method to load the calibration matrix from a calibration. Since now we have it at https://github.com/qilimanjaro-tech/qililab/pull/784 i will merge those files from there

jjmartinezQT: Sounds good.

jjmartinezQT: resolving this now.

jjmartinezQT (approved):

PR #779, opened 2024-08-19
Qhc 684 implement distortions in qprogram

jordivallsq:
Created a section on platform execute_qprogram_qblox to define the waveform distortions based on pulse_distortions


jordivallsq:
Requesting permission from @jjmartinezQT to launch a new realease from qpysequence and fix the test error

fedonman (changes requested):
Awesome job! I like the simplicity. We should add tests to verify that the waveforms are modified indeed based on the distortion provided.

jordivallsq:
> Awesome job! I like the simplicity. We should add tests to verify that the waveforms are modified indeed based on the distortion provided. E.g. play the same waveform to two buses, one without distortions and the other with, and check that the first one the pulse is umodified and the second one is.

I added the tests including some qpysequence changes to make it more consistent

src/qililab/platform/platform.py
```diff
@@ -860,7 +860,13 @@ def _execute_qprogram_with_qblox(  # pylint: disable=too-many-locals
             markers=markers,
         )
         buses = {bus_alias: self._get_bus_by_alias(alias=bus_alias) for bus_alias in sequences}
-
+        for bus_alias, bus in buses.items():
+            if bus.distortions:
+                for distrortion in bus.distortions:
```
fedonman, line 865: minor typo
```suggestion
                for distortion in bus.distortions:
```

src/qililab/platform/platform.py
```diff
@@ -860,7 +860,13 @@ def _execute_qprogram_with_qblox(  # pylint: disable=too-many-locals
...
+                    for waveforms in sequences[bus_alias]._waveforms._waveforms:  # pylint: disable=protected-access
+                        sequences[bus_alias]._waveforms.modify(  # pylint: disable=protected-access
+                            waveforms.name, distrortion.apply(waveforms.data)
+                        )
```
fedonman, line 869: minor typo
```suggestion
                    for waveform in sequences[bus_alias]._waveforms._waveforms:  # pylint: disable=protected-access
                        sequences[bus_alias]._waveforms.modify(  # pylint: disable=protected-access
                            waveform.name, distrortion.apply(waveform.data)
                        )
```

tests/platform/test_platform.py
```diff
@@ -655,6 +655,33 @@ def test_execute_qprogram_with_qblox(self, platform: Platform):
...
+        for bus_alias, bus in buses.items():
+            if bus.distortions:
+                for distrortion in bus.distortions:
+                    for waveforms in sequences[bus_alias]._waveforms._waveforms:  # pylint: disable=protected-access
+                        sequences[bus_alias]._waveforms.modify(  # pylint: disable=protected-access
+                            waveforms.name, distrortion.apply(waveforms.data)
+                        )
+        assert test_waveforms_q0.to_dict() == sequences["drive_line_q0_bus"]._waveforms.to_dict()
+        assert test_waveforms_q1.to_dict() == sequences["drive_line_q1_bus"]._waveforms.to_dict()
+
```
fedonman, line 684: You don't call `platform.execute_qprogram()` method at any point. We should always call the method-under-test in our tests, and not replicate its contents, because if the method changes the test will continue to pass.

Check the `test_execute_qprogram_with_qblox()` and `test_execute_qprogram_quantum_machines()`. I can also help you with implementation if you get stuck since these `execute` methods are a bit complex.

fedonman (changes requested):
Great work :100: I requested changes mainly because of the test.

tests/platform/test_platform.py
fedonman: You can also add the asserts in the `test_execute_qprogram_with_qblox()` directly.

jordivallsq: I implemented the changes in a different function than `test_execute_qprogram_with_qblox()` with just a play

fedonman (approved):
:100: :100:

PR #782, opened 2024-08-21
[QHC-662] Implement the Experiment class

fedonman:


src/qililab/qprogram/experiment.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def execute_qprogram(self, qprogram: QProgram):
+        """Execute a quantum program within the experiment.
+
+        Appends an ExecuteQProgram operation to the active block of the experiment.
+
+        Args:
+            qprogram (QProgram): The quantum program to be executed.
+        """
+        operation = ExecuteQProgram(qprogram=qprogram)
```
jordivallsq, line 50: Regarding this `ExecuteQProgram`, shouldn't we add all the other variables introduced in platform.execute_qprogram? e.g. bus mapping or debug.

If we do not, why do we use the same variables for `SetParameter`?

jordivallsq (approved):
Great work!!! Tested it in Hardware and it works. I left a couple of comments but couldn't find anything big.

Looking forward to see the execute_experiment

PR #783, opened 2024-08-21
added upsampling_mode pulsed for flux lines(amplified output mode in controller config)

jordivallsq:
Following QM indications, we should add the "upsampling_mode": "pulsed" from https://docs.quantum-machines.co/1.2.0/docs/Guides/opx1000_fems/?h=upsampling#sampling-rate to improve the step response while upsampling


jjmartinezQT (approved):
Good to go.

PR #787, opened 2024-08-27
Solved nshots bug by fixing qua for loop implementation

jordivallsq:
In this PR the step/2 represented in the qua for loop is removed as it is redundant and causes errors.
This addition to the stop contition causes crashes for experiments with nshots and can cause to run loops with variable values above the stop limit.
Note that qua._for() current condition encompasses [start, stop] while np.arange is [start, stop) (there is no need to add a fraction of step to include the stop value).


visagim (approved):
nice

jordivallsq:
HW gave the approval that the bug is fixed with this implementation

PR #788, opened 2024-08-29
Qhc 723 temporal solution to bypass dataloss error

jordivallsq:
Confirmation from HW that this fix solves temporarily the issues with the data loss error.

This changes are temporal as QM fix their code, they are also integrated with the close qprogram fix when an error is thrown

### DIFF
--- a/src/qililab/platform/platform.py
+++ b/src/qililab/platform/platform.py
@@ -879,6 +879,7 @@ class Platform:  # pylint: disable = too-many-public-methods, too-many-instance-
 
             except StreamProcessingDataLossError as dataloss:
                 time_interval = datetime.datetime.now() - start_time
+                start_time = datetime.datetime.now()
                 warnings.warn(
                     f"Warning: {dataloss} raised, retrying experiment ({iteration+1}/{dataloss_tries} available tries) after {time_interval.seconds} s"
                 )
@@ -886,12 +887,13 @@ class Platform:  # pylint: disable = too-many-public-methods, too-many-instance-
                 if iteration + 1 != dataloss_tries:
                     time.sleep(1 * dataloss_tries)
                     continue
-                else:
-                    cluster.turn_off()
-                    raise dataloss
+                cluster.turn_off()
+                raise dataloss
             except Exception as e:
                 cluster.turn_off()
                 raise e
+
+        return results
 
     def execute(
         self,
--- a/src/qililab/platform/platform.py
+++ b/src/qililab/platform/platform.py
@@ -17,6 +17,7 @@
 import ast
 import io
 import re
+import warnings
 from copy import deepcopy
 from dataclasses import asdict
 from queue import Queue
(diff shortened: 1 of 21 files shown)


src/qililab/platform/platform.py
```diff
@@ -837,38 +844,56 @@ def _execute_qprogram_with_quantum_machines(  # pylint: disable=too-many-locals,
         threshold_rotations: dict[str, float | None] = {},
         thresholds: dict[str, float | None] = {},
         calibration: Calibration | None = None,
+        dataloss_tries: int = 3,
```
DZPM, line 847: why a different dataloss_tries (3) than before (2)?

src/qililab/platform/platform.py
```diff
@@ -837,38 +844,56 @@ def _execute_qprogram_with_quantum_machines(  # pylint: disable=too-many-locals,
...
+
+            except StreamProcessingDataLossError as dataloss:
+                time_interval = datetime.datetime.now() - start_time
+                start_time = datetime.datetime.now()
+                warnings.warn(
+                    f"Warning: {dataloss} raised, retrying experiment ({iteration+1}/{dataloss_tries} available tries) after {time_interval.seconds} s"
                 )
-                measurement_result.set_classification_threshold(thresholds.get(measurement.bus, None))
-                results.append_result(bus=measurement.bus, result=measurement_result)
+                print(traceback.format_exc())
```
DZPM, line 886: Why `print()` ?

As a good practice, I prefer a warning, as warnings logs can be easily disabled.

src/qililab/platform/platform.py
```diff
@@ -837,38 +844,56 @@ def _execute_qprogram_with_quantum_machines(  # pylint: disable=too-many-locals,
...
+                if iteration + 1 != dataloss_tries:
+                    time.sleep(1 * dataloss_tries)
```
DZPM, line 888: The way it's implemented, this `sleep(dataloss_tries)` timeout will be part of the `time_interval`.

Not wrong, just take in mind that the real `time_interval` would be a bit shorter.

DZPM (approved):
Please add at least a test case to validate the `except StreamProcessingDataLossError` block. A mock that rises an error should be enough.

src/qililab/platform/platform.py
jordivallsq: This is thee number of loops suggested by HW as too many may be too long on queue (they also use 5 though)

src/qililab/platform/platform.py
jordivallsq: great point, fixed!

src/qililab/platform/platform.py
jordivallsq: Fixed, I moved the time_start after the wait

DZPM (approved):
LGTM

DZPM:
Jordi asked me to force merge

PR #789, opened 2024-09-02
Fixed float numbers for Voltage ad float scalar

jordivallsq:


visagim (commented):
confirm that this change in `stop`  does not enter in conflict with `calculate_iterations` at https://github.com/qilimanjaro-tech/qililab/blob/92c8a4e38d06f0545582998a887de112118ecdc6/src/qililab/qprogram/quantum_machines_compiler.py#L543-L556

visagim:
> confirm that this change in `stop` does not enter in conflict with `calculate_iterations` at
> 
> https://github.com/qilimanjaro-tech/qililab/blob/92c8a4e38d06f0545582998a887de112118ecdc6/src/qililab/qprogram/quantum_machines_compiler.py#L543-L556

I dont think it does but just to be sure

jordivallsq:
It does not, I checked this already in debug mode

src/qililab/qprogram/quantum_machines_compiler.py
```diff
@@ -259,6 +259,11 @@ def _handle_for_loop(self, element: ForLoop):
         if element.variable.domain is Domain.Time:
             start = max(start, self.MINIMUM_TIME)
 
+        if element.variable.domain is Domain.Voltage or (
+            element.variable.domain is Domain.Scalar and isinstance(element.variable, FloatVariable)
+        ):
+            stop += step / 2
```
fedonman, line 265: Do you want to check if the variable is of `float` type? Then you can do it directly with either `isinstance(element.variable, FloatVariable)` or `isinstance(element.variable, float)`. No need to check the domains.

jordivallsq: True I made everything more complicated, I started with Voltage and realized that float scalar should be included. Fixed it!

jordivallsq: @fedonman As you can see in the tests, the Frequencies also are included in the condition FloatVariable and so the stop is also modified even though we don't want to include it inside the condition.

This test error only appears because frequency is considered a Float:
image

fedonman (approved):
:100:

PR #790, opened 2024-09-02
[QHC-713] Implement ExperimentExecutor

fedonman:
- Introduced the `ExperimentExecutor` class to manage and execute quantum experiments within the Qililab framework. This class provides a streamlined way to handle the setup, execution, and results retrieval of experiments.

  Temporary Constraints:

  - The experiment must contain only one `QProgram`.
  - The `QProgram` must contain a single measure operation.
  - Parallel loops are not supported.

- Introduced the `platform.execute_experiment()` method for executing experiments. This method simplifies the interaction with the ExperimentExecutor by allowing users to run experiments with a single call.

  
- Variables in `QProgram` and `Experiment` now require a label.

  ```Python
  qp = QProgram()
  gain = qp.variable(label="gain", domain=Domain.Voltage)
  ```


DZPM (approved):
LGTM

src/qililab/qprogram/experiment_executor.py
```diff
@@ -0,0 +1,258 @@
...
+            # Track the index for this loop
+            self.loop_indices[loop_label] = 0
+
+            return loop_task_id
+
+        stored_operations.append(create_progress_bar)
+
+        def advance_progress_bar(variable_value: int | float) -> None:
+            loop_task_id = self.task_ids[block.uuid]
+            progress.update(loop_task_id, description=f"Looping over {loop_label}: {variable_value}")
```
jjmartinezQT, line 120: Could we fix the number of decimals here? Fabio got OCD seeing the video, me a little bit less but also. :D

jjmartinezQT (changes requested):
Silly comment about decimals. This is just amazing, awesome work.

jjmartinezQT (approved):
Amazing job.

PR #793, opened 2024-09-05
Implemented delay qblox qprogram

jordivallsq:


src/qililab/qprogram/qblox_compiler.py
```diff
@@ -130,12 +133,13 @@ def __init__(self) -> None:
         self._buses: dict[str, BusCompilationInfo]
         self._sync_counter: int
 
-    def compile(
+    def compile(  # noqa: C901
```
jjmartinezQT, line 136: Pylint skip lines are ok if you do not manage to get around it, but please include the text instead of the code so a developer looking at the code knows immediately what it is about ;)

jjmartinezQT (changes requested):
Good work, minor change required.

jjmartinezQT (approved):
thanks!

jjmartinezQT (approved):

src/qililab/qprogram/qblox_compiler.py
jordivallsq: Flake8 does not allow to ignore the error with the sae texts as pylint but I changed the complexity with `noqa: max-complexity=25`

jjmartinezQT: I see, thought it was pylint.

PR #794, opened 2024-09-06
Update crosstalk_matrix.py

jjmartinezQT:
change in string representation for xtalk matrix


jordivallsq:
Have we validated that QM allows to add the same element into the configuration of the xtalk? I could not find any instance of this happening in their documentation https://docs.quantum-machines.co/1.2.0/docs/Guides/features/?h=crossta#crosstalk-matrix

jjmartinezQT:
> Have we validated that QM allows to add the same element into the configuration of the xtalk? I could not find any instance of this happening in their documentation https://docs.quantum-machines.co/1.2.0/docs/Guides/features/?h=crossta#crosstalk-matrix

We do apply this crosstalk matrix in software, inverting it and using it to multiply the array of fluxes after transpilation.

fedonman (approved):
:100:

PR #796, opened 2024-09-09
[QHC-734] Change experiment's results h5 file to support multiple measurements

fedonman:


src/qililab/result/experiment_results_writer.py
```diff
@@ -0,0 +1,250 @@
...
+    def platform(self, platform: str):
+        """Sets the YAML representation of the platform.
+
+        Args:
+            platform (str): The YAML string representing the platform.
+        """
+        path = ExperimentResults.PLATFORM_PATH
+        if path in self._file:
+            del self._file[path]
+        self._file[path] = platform
```
DZPM, line 214: I don't understand why are you deleting `self._file[path]`, isn't it overwritten in the next line?

I guess an h5py file does not get overwritten?

fedonman: If I don't delete it, I get the following error:

image

src/qililab/qprogram/decorators.py
```diff
@@ -18,9 +18,7 @@ def wrapper(*args, **kwargs):
             if parameter not in kwargs and len(args) == 1:
                 kwargs[parameter] = None
             # Get the argument by name
-            param_value = (
-                kwargs.get(parameter) if parameter in kwargs else args[1]
-            )  # assuming 1 is the default position
+            param_value = kwargs.get(parameter) if parameter in kwargs else None
```
visagim, line 21: why are we changing this from args[1] to None?

src/qililab/yaml.py
```diff
@@ -30,9 +31,64 @@ def deque_constructor(constructor, node):
     return deque(constructor.construct_sequence(node))
 
 
+# Lambda representer: convert lambda to a string
+def lambda_representer(representer, data):
```
visagim, line 35: should we add typings?

src/qililab/result/experiment_results.py
```diff
@@ -0,0 +1,273 @@
...
+                    f"{ExperimentResults.MEASUREMENTS_PATH}/{measurement_name}/results"
+                ]
+                # Store the dimensions
+                self.dimensions[(qprogram_name, measurement_name)] = qprogram_data[
+                    f"{ExperimentResults.MEASUREMENTS_PATH}/{measurement_name}/results"
+                ].dims
+
+        return self
+
+    def __exit__(self, *args):
```
visagim, line 70: *args is unused?

src/qililab/result/experiment_results.py
```diff
@@ -0,0 +1,273 @@
...
+        """Exit the context manager and close the HDF5 file."""
+        if self._file is not None:
+            self._file.close()
+
+    def get(self, qprogram: int | str = 0, measurement: int | str = 0):
```
visagim, line 75: missing return in function definition

src/qililab/qprogram/experiment_executor.py
```diff
@@ -1,188 +1,366 @@
...
+                    nested_operations = self._prepare_operations(element, progress)
+                    elements_operations.extend(nested_operations)
+
+            return elements_operations
+
+        def store_results(qprogram_results: QProgramResults, qprogram_index: int):
+            """Store the result in the correct location within the ExperimentResultsWriter."""
+            # Determine the index in the ExperimentResultsWriter based on current loop indices
+            for measurement_index, measurement_result in enumerate(qprogram_results.timeline):
+                indices = tuple([qprogram_index, measurement_index]) + tuple(index for _, index in loop_indices.items())
```
visagim, line 346: sourcery suggests 
```suggestion
                for measurement_index, measurement_result in enumerate(qprogram_results.timeline):
                indices = (qprogram_index, measurement_index) + tuple(
                    index for _, index in loop_indices.items()
                )
```

visagim (changes requested):
Some minor comments only actually. Great job! :tada: 
Also please make sure you test this on hardware :pray:

src/qililab/qprogram/decorators.py
fedonman: Because before we always tried to get the first parameter (args[1]) which is wrong. Now, if `parameter` is not there, the `isinstance(None, Variable)` will not raise the `ValueError` falsely.

src/qililab/yaml.py
fedonman: Not needed. These functions are not called by humans, but by YAML library. And all their parameters' types are `Any` which is the implicit value if you don't provide a specific type.

src/qililab/result/experiment_results.py
fedonman: I changed it to `*_` so it is discarded.

src/qililab/result/experiment_results.py
fedonman: Nice catch. Writing the return definition, I saw that the dictionary was too complicated, so I created a dataclass `DimensionInfo` that stores the information more organized.

src/qililab/result/experiment_results.py
```diff
@@ -0,0 +1,273 @@
...
+
+            fig, ax1 = plt.subplots()
+            ax1.set_title(self.path)
+            ax1.set_xlabel(x_labels[0])
+            ax1.set_ylabel(y_labels[0])
+
+            # Force scientific notation
+            ax1.ticklabel_format(axis="both", style="sci", scilimits=(-3, 3))
+
+            mesh = ax1.pcolormesh(x_edges, y_edges, s21.T, cmap="viridis", shading="auto")
```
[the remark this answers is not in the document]
fedonman: True! There was (and still is) an ongoing discussion about a "global" qilimanjaro styling in plots. We had a small discussion with Ramiro that we might even allow users to pass a dictionary with styling preferences and we take that into account for plotting. Experimentalists will help a lot in these plot methods. But all of these will be done in a separate PR, the point of this PR is to create a "prototype" version so experimentalists can, well, experiment.

src/qililab/qprogram/experiment_executor.py
fedonman: nice catch!

visagim:
let me know once code coverage is fixed and i can approve!

fedonman:
> let me know once code coverage is fixed and i can approve!

@visagim codecov fails for the usual `TYPE_CHECKING` bug, tests cover all changes

visagim (approved):

DZPM (approved):

PR #797, opened 2024-09-10
Nshots for execute anneal program.

jjmartinezQT:
This PR introduces the change suggested by hardware to include running n_shots in an annealing execution in qililab.


visagim (approved):
all good

PR #798, opened 2024-09-13
fixed bug IF

jordivallsq:
Fixed bug for set parameter IF in QM.
I didn't add changelog for the bug as inside the changelog there is already a change inside bug fixes explaining this issue (PR 764)


fedonman (approved):
nice catch! :100:

PR #799, opened 2024-09-16
declares yaml as unsafe to allow serialization of zero

jjmartinezQT:


fedonman (approved):
obama-awards-obama-a-medal

PR #805, opened 2024-09-17
[QHC-752] Change xtalk matrix from collection to list

visagim:


jordivallsq (approved):
Good job!

Also I see that this change does not affect the tests. I will test the Xtalk in a QM anyway and I will let you know if something happens.

PR #806, opened 2024-09-17
[QHC-753] Add padding to anneal waveforms

visagim:


src/qililab/analog/annealing_program.py
```diff
@@ -108,9 +110,13 @@ def get_waveforms(self, crosstalk_matrix: CrosstalkMatrix | None = None) -> dict
                 bus_to_flux_map[flux_bus.bus] = flux_bus.flux
         flux_to_bus_map = {v: k for k, v in bus_to_flux_map.items()}
 
+        # add padding to waveforms if duration is not multiple of minimum clock time
+        padded_ns = 0
+        if len(self._transpiled_program) % minimum_clock_time != 0:
+            padded_ns = 4 - len(self._transpiled_program) % 4 if len(self._transpiled_program) % 4 != 0 else 0
```
DZPM, line 116: Question: this `4` is a padding that's required by a specific device manufacturer, right? Could another device manufacturer could have a different padding?

If so, maybe it could make sense to put it in a constant (`MANDATORY_PADDING`?), so the code is easier to understand/maintain in the future.

src/qililab/analog/annealing_program.py
```diff
@@ -87,7 +87,9 @@ def _chip_element_to_short(self, chip_element: str) -> str:
             else f"{chip_element[0]}{split_element[1]}_{split_element[2]}"
         )
 
-    def get_waveforms(self, crosstalk_matrix: CrosstalkMatrix | None = None) -> dict[str, ArbitraryWave]:
+    def get_waveforms(
+        self, crosstalk_matrix: CrosstalkMatrix | None = None, minimum_clock_time: int = 1
+    ) -> dict[str, ArbitraryWave]:
         """Returns a dictionary containing (bus, waveform) for each flux control from the transpiled fluxes. `AnnealingProgram.transpile` should be run first. The waveform is an arbitrary waveform obtained from the transpiled fluxes.
 
         Args:
```
jordivallsq, line 95: Inside args please give a little bit of context about the minimum_clock_time (and why is it defaulted to 1)

tests/analog/test_annealing_program.py
```diff
@@ -118,7 +118,7 @@ def test_transpile(self, annealing_program, transpiled_program_dictionary):
 
     def test_get_waveforms(self, annealing_program_transpiled):
         """Test get waveforms method works as intended"""
-        anneal_waveforms = annealing_program_transpiled.get_waveforms()
+        anneal_waveforms = annealing_program_transpiled.get_waveforms(minimum_clock_time=1)
```
DZPM, line 121: Could you add some more tests with different `minimum_clock_time` values?

Testing only the default value could be risky

DZPM (changes requested):
Please add tests for different `minimum_clock_time` values

tests/platform/test_platform.py
```diff
@@ -149,10 +149,10 @@ def get_anneal_qprogram(runcard, flux_to_bus_topology):  # pylint: disable=too-m
     platform.flux_to_bus_topology = flux_to_bus_topology
     anneal_waveforms = {
         next(element.bus for element in platform.flux_to_bus_topology if element.flux == "phix_q0"): Arbitrary(
-            np.array([1])
+            np.array([0.0, 0.0, 0.0, 1.0])
         ),
         next(element.bus for element in platform.flux_to_bus_topology if element.flux == "phiz_q0"): Arbitrary(
-            np.array([2])
+            np.array([0.0, 0.0, 0.0, 2.0])
```
jordivallsq, line 155: Can we test more complex arbitray shapes just in case? If it makes sense of course

src/qililab/analog/annealing_program.py
visagim: yup, well spotted

tests/platform/test_platform.py
visagim: i dont think it makes a lot of sense at this step, since we already have somewhat more complicated waveforms at the `get_waveform`test. Even those are not quite complicated, but making more complex arrays won't make a difference since we're not operating with them that much (i..e. there's not gonna be any difference between a size 3 array with 2 0s and 1 random float and a 10k long array with floats, and the later would make reviewing/updating the tests more difficult in the future)

PR #812, opened 2024-09-26
Fixed missing path to set/get IF

jordivallsq:


fedonman:
Gathered everything at https://github.com/qilimanjaro-tech/qililab/pull/818

PR #813, opened 2024-09-26
[QHC-765] Use ruff for linting and formatting

fedonman:


src/qililab/__init__.py
```diff
@@ -26,14 +26,52 @@
...
+]
+
+
 with contextlib.suppress(NameError, ImportError):
     # Since Ipython magic methods can only be imported from inside a Jupyter Notebook,
     # here we first check that `get_ipython` exists (which means we are inside a Jupyter Notebook)
-    get_ipython()  # type: ignore  # noqa: F405 # pylint: disable=undefined-variable
-    from .slurm import submit_job  # pragma: no cover
+    get_ipython()  # type: ignore  # noqa: F821
+    from .slurm import submit_job as submit_job  # pragma: no cover
```
DZPM, line 77: Isn't submit_job the same as submit_job?

src/qililab/slurm.py
```diff
@@ -89,7 +89,7 @@ def submit_job(line: str, cell: str, local_ns: dict) -> None:
     # are imported inside the SLURM job)
     notebook_code = "\n".join(local_ns["In"]).split("\n")
     import_lines = [line for line in notebook_code if line.startswith(("import ", "from ")) and "slurm" not in line]
-    executable_code = "\n".join(import_lines + [cell])
+    executable_code = "\n".join([*import_lines, cell])
```
DZPM, line 92: It's less legible. Is the old way invalid?

tests/waveforms/test_arbitrary.py
```diff
@@ -1,5 +1,3 @@
-import re
```
DZPM, line 1: Surprising an unused import didn't ring any bell

DZPM:
First: Good job!! :goat:

Second: this PR is huge, will have a big impact on other PRs. Let's make sure everyone rebases once this is merged

src/qililab/slurm.py
fedonman: This rule is described here: https://docs.astral.sh/ruff/rules/collection-literal-concatenation/

It claims it is more readable and efficient and follows PEP 448: https://peps.python.org/pep-0448/

tests/waveforms/test_arbitrary.py
fedonman: Yes, because `pre-commit` checks only files in the commit. I now run `ruff` in the whole qililab codebase

src/qililab/__init__.py
fedonman: Yes. This is the rule about unused imports: https://docs.astral.sh/ruff/rules/unused-import/

If you check any `__init__.py` file you will see I have added the `__all__` variable as suggested. But for this specific case I can't, since it is imported conditionally. So I chose the other option of re-importing as stated in the rule.

fedonman:
@DZPM Finalized `pyproject.toml`. I believe this is the one we should use as global qili config. Added many more rules, and ignore rules. Implemented some in qililab, others are commented out, so we can implement them one-by-one.

DZPM (approved):

PR #814, opened 2024-10-02
[QHC-768] Add `GetParameter` operation to Experiment

fedonman:
- A new `GetParameter` operation has been added to the Experiment class, accessible via the `.get_parameter()` method. This allows users to dynamically retrieve parameters during experiment execution, which is particularly useful when some variables do not have a value at the time of experiment definition but are provided later through external operations. The operation returns a `Variable` that can be used seamlessly within `SetParameter` and `ExecuteQProgram`.

  Example:

  ```
  experiment = Experiment()

  # Get a parameter's value
  amplitude = experiment.get_parameter(bus="drive_q0", parameter=Parameter.AMPLITUDE)

  # The returned value is of type `Variable`. It's value will be resolved during execution.
  # It can be used as usual in operations.

  # Use the variable in a SetOperation.
  experiment.set_parameter(bus="drive_q1", parameter=Parameter.AMPLITUDE, value=amplitude)

  # Use the variable in an ExecuteQProgram with the lambda syntax.
  def get_qprogram(amplitude: float, duration: int):
    square_wf = Square(amplitude=amplitude, duration=duration)

    qp = QProgram()
    qp.play(bus="readout", waveform=square_wf)
    return qp

  experiment.execute_qprogram(lambda: amplitude=amplitude: get_qprogram(amplitude, 2000))
  ```


visagim (approved):
Not much to comment since most of the changes are removing `dataclass` and rest of changes seem fine to me. Test in on hardware and if it works it's good to go. Great job!

PR #817, opened 2024-10-07
[QHC-735] Split execution of qprogram to compile/execute

fedonman:
The execution of `QProgram` has been split into two distinct steps: **compilation** and **execution**.

  1. **Compilation**: Users can now compile a `QProgram` by calling:

  ```python
  platform.compile_qprogram(
    qprogram: QProgram,
    bus_mapping: dict[str, str] | None = None,
    calibration: Calibration | None = None
  )
  ```

  This method can be executed without being connected to any instruments. It returns either a `QbloxCompilationOutput` or a `QuantumMachinesCompilationOutput`, depending on the platform setup.

  2. **Execution**: Once the compilation is complete, users can execute the resulting output by calling:

  ```python
  platform.execute_compilation_output(
    output: QbloxCompilationOutput | QuantumMachinesCompilationOutput,
    debug: bool = False
  )
  ```

  If desired, both steps can still be combined into a single call using the existing method:

  ```python
  platform.execute_qprogram(
    qprogram: QProgram,
    bus_mapping: dict[str, str] | None = None,
    debug: bool = False
  )
  ```


jjmartinezQT (approved):
Very good job, I have not seen any weak point in this splitting so far. We need to circulate this amongst experimentalists as they have been asking for this for some time now.

PR #822, opened 2024-10-21
Small bug I found, in the `_order_result`

GuillermoAbadLopez:
If `qubit` is not in `qubits_m`, then`qubits_m[qubit]` is going to explode


GuillermoAbadLopez:
It does not explode if you assign it...

PR #825, opened 2024-10-30
[QHC-783] Changes needed for the monthly OKR

fedonman:


jjmartinezQT (approved):
had reviewed already but not yet approved, good work.

PR #826, opened 2024-11-05
added timeout and modified set and get for qm

jordivallsq:
I also added a bugfix after the major runcard changes involving set and get paremeters for QM


visagim (approved):
Great job! We'll follow through changing "identifier" to "channel" as discussed in [slack](https://qilimanjaro.slack.com/archives/C02RB1ZGU3G/p1730900170589679) on a separate pr

PR #828, opened 2024-11-07
modify magic method + unit tests

ziiiki:
https://linear.app/qilimanjaro/issue/CLO-849/install-slurm-in-sauron-and-give-support-1q-and-2q-in-qinu

### DIFF
--- a/src/qililab/slurm.py
+++ b/src/qililab/slurm.py
@@ -15,7 +15,7 @@
 import ast
 import os
 from types import ModuleType
-
+import subprocess
 from IPython.core.magic import needs_local_scope, register_cell_magic
 from IPython.core.magic_arguments import argument, magic_arguments, parse_argstring
 from submitit import AutoExecutor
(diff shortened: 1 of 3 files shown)


src/qililab/slurm.py
```diff
@@ -85,13 +86,17 @@ def submit_job(line: str, cell: str, local_ns: dict) -> None:
...
+    gres = args.gres
     job_name = args.name
     time_limit = int(args.time)
     folder_path = args.logs
     execution_env = args.execution_environment
     begin_time = args.begin
     low_priority = args.low_priority
 
+    if gres is None:
+        raise ValueError("GRES needs to be provided!")
```
DZPM, line 98: Would be nice to return an error with a list of valid options.
Do we have the list of GRES chips?

If we don't have the list, at least an example? Like `--gres aQPU1`
Otherwise, some clue on when to find it?

ziiiki: yes, this was to be included in the documentation! But I will return the values in the error, good idea.

ziiiki: see the update code: 
image

ziiiki: see the updated code: 
image

PR #832, opened 2024-11-19
Temporal fix for QM timeout error

jordivallsq:
I created a try and except similar to the dataloss error to restart the measurement in case of random timeout issue

### DIFF
--- a/src/qililab/platform/platform.py
+++ b/src/qililab/platform/platform.py
@@ -913,7 +913,7 @@ class Platform:
         debug: bool = False,
     ):
         qua, configuration, measurements = output.qua, output.configuration, output.measurements
-        for iteration in np.arange(timeout_tries):  # TODO: This is a temporal fix as QM fixes the dataloss error
+        for iteration in np.arange(timeout_tries):  # TODO: This is a temporal fix as QM fixes the timeout error
             try:
                 cluster.append_configuration(configuration=configuration)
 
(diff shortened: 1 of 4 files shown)


src/qililab/platform/platform.py
```diff
@@ -898,29 +906,51 @@ def _execute_qblox_compilation_output(self, output: QbloxCompilationOutput, debu
...
-        results = QProgramResults()
-        # Doing manual classification of results as QM does not return thresholded values like Qblox
-        for measurement in measurements:
-            measurement_result = QuantumMachinesMeasurementResult(
-                measurement.bus,
-                *[acquisitions[handle] for handle in measurement.result_handles],
-            )
-            measurement_result.set_classification_threshold(measurement.threshold)
-            results.append_result(bus=measurement.bus, result=measurement_result)
+        for iteration in np.arange(timeout_tries):  # TODO: This is a temporal fix as QM fixes the dataloss error
```
jjmartinezQT, line 916: the comment indicates this is for dataloss error but it is not, this is for the infinite hanging up, no?

jjmartinezQT (changes requested):
small thing about a comment, rest looks ok to me.

src/qililab/platform/platform.py
jordivallsq: yes, the name was from a previous version, now it says timeout instead

PR #833, opened 2024-11-20
[QHC-801] Add arbitrary waveforms methods to QDAC2

visagim:
Still some minor todos left but code is functional already

### DIFF
--- a/src/qililab/instruments/qdevil/qdevil_qdac2.py
+++ b/src/qililab/instruments/qdevil/qdevil_qdac2.py
@@ -20,6 +20,7 @@ from qililab.instruments import InstrumentFactory, ParameterNotFound, check_devi
 from qililab.instruments.voltage_source import VoltageSource
 from qililab.typings import ChannelID, InstrumentName, Parameter, ParameterValue
 from qililab.typings import QDevilQDac2 as QDevilQDac2Driver
+from qililab.waveforms import Waveform
 
 
 @InstrumentFactory.register
@@ -39,9 +40,11 @@ class QDevilQDac2(VoltageSource):
         """Contains the settings of a specific signal generator."""
 
         low_pass_filter: list[str]
+        mode: str = "offset"
 
     settings: QDevilQDac2Settings
     device: QDevilQDac2Driver
+    _cache: dict[int | str, bool] = {}  # noqa: RUF012
 
     @property
     def low_pass_filter(self):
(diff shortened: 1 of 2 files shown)


jjmartinezQT:
I have reviewed it and works fine. Victor has helped me understand it better. I will document this a bit more in a PoC for active reset and show it to hardware to agree on next steps. Could you please @jordivallsq check the code and play with it if you still want to give it a go about the traces? I think is fine but worth adding an extra pair of eyes. 

Below screenshot of two different waveforms played in two different channels, sync by default without any modification to Victor's notebook, after Victor helped me understanding this. 

tek00000

src/qililab/instruments/qdevil/qdevil_qdac2.py
```diff
@@ -102,6 +105,58 @@ def set_parameter(self, parameter: Parameter, value: ParameterValue, channel_id:
...
+
+        Raises:
+            ValueError: if a waveform is already allocated
+        """
+        values = list(waveform.envelope())  # TODO: does np array work?
+        if channel_id in self._cache:
+            raise ValueError(
+                f"Device {self.name} already has a waveform allocated to channel {channel_id}. Clear the cache before allocating a new waveform"
+            )
+        trace = self.device.allocate_trace(channel_id, len(values))
```
jjmartinezQT, line 132: The QDAC manual says that uploading waveforms over LAN connection can sometimes cause errors due to 
unestability of the network, so they recommend to check after uploading if there are any errors, and if there are, re-upload the waveform again. I think we should check for that to make this as robust as possible.

"When uploading vast amount of binary data over LAN, for example multiple large (4-6 mio points)
arbitrary waveform traces), the instrument may occasionally report an error and the upload will fail. It
happens if there are some instabilities in the local network, but other times this is not the cause. So
always check for errors after uploading a trace, and in case of errors, re-upload the trace."

jjmartinezQT, line 132: This is a change we need now, traces length must be even values. We need to check for that or otherwise raise an warning for the moment, but maybe an error is better as it does duplicate the last point in the current firmware but they might change that in future firmware updates:

"The maximum length of a single trace is 6
million points (6,291,456). The length of traces should be an even number.
Note that with the current firmware version (13-1.57), no error is produced when uploading traces with
uneven number of points. Instead, the last point is duplicated."

Also we need to check the range is between [-1, 1]:
"Trace values should be in the interval [-1;1]. Their scaling into actual volts is handled by the AWG
generator of the relevant channels. "

jjmartinezQT (changes requested):

src/qililab/instruments/qdevil/qdevil_qdac2.py
jjmartinezQT: thanks for the fix!

jjmartinezQT (approved):

PR #835, opened 2024-11-21
added channel_id to QRM-RF

jordivallsq:


jjmartinezQT (approved):
well spotted.

src/qililab/platform/platform.py
```diff
@@ -893,9 +893,10 @@ def _execute_qblox_compilation_output(self, output: QbloxCompilationOutput, debu
         results = QProgramResults()
         for bus_alias, bus in buses.items():
             if bus.has_adc():
-                bus_results = bus.acquire_qprogram_results(acquisitions=acquisitions[bus_alias])
-                for bus_result in bus_results:
-                    results.append_result(bus=bus_alias, result=bus_result)
+                for channel in buses[bus.alias].channels:
```
jjmartinezQT, line 896: we need to modify the unittest for this to pass.

jjmartinezQT (changes requested):

tests/platform/test_platform.py
```diff
@@ -693,7 +693,7 @@ def test_execute_qprogram_with_qblox(self, platform: Platform):
 
         # assert run executed all three times (6 because there are 2 buses)
```
jjmartinezQT, line 694: comment says 6 because there are two buses, now with the channel id it went to 9, why? can we specify it in the test?

jjmartinezQT (changes requested):

tests/platform/test_platform.py
jordivallsq: I am checking right now why, I wanted to see if there was anything else in code coverage to handle

src/qililab/platform/platform.py
jordivallsq: The unitests were left intact and all passed, the problem was a missing condition to exclude all non qblox modules

tests/platform/test_platform.py
jordivallsq: Answered above

jjmartinezQT (approved):
good work

PR #839, opened 2024-11-26
Adding Block handlers to qblox and quantum machines compilers.

jjmartinezQT:
This PR brings the empty handlers needed to blocks in qprogram compilers so the new preparation and measurement.

### DIFF
--- a/src/qililab/qprogram/quantum_machines_compiler.py
+++ b/src/qililab/qprogram/quantum_machines_compiler.py
@@ -174,11 +174,9 @@ class QuantumMachinesCompiler:
                     raise NotImplementedError(
                         f"{element.__class__} operation is currently not supported in Quantum Machines."
                     )
-                if isinstance(element, (InfiniteLoop, ForLoop, Loop, Average, Parallel)):
+                if isinstance(element, (InfiniteLoop, ForLoop, Loop, Average, Parallel, Block)):
                     with handler(element):
                         traverse(element)
-                elif isinstance(element, Block):
-                    traverse(element)
                 else:
                     handler(element)
             self._qprogram_block_stack.pop()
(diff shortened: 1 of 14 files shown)


tests/qprogram/test_qblox_compiler.py
```diff
@@ -354,6 +362,19 @@ def test_play_named_operation_and_bus_mapping(self, play_named_operation: QProgr
...
+        measurement_block = calibration.get_block("measurement")
+        measure_program = QProgram()
+        measure_program.add_block(measurement_block)
+        
+        with (
+            patch.object(QbloxCompiler, "_handle_block") as handle_block,
+        ):
+            compiler.compile(measure_program)
+
+            assert handle_block.call_count == 1
```
fedonman, line 377: I don't trust this test tbh, because it just checks that the handler is called, not that correct Q1ASM is produced. Can you add a test for the Q1ASM? Check the process with the rest of the tests for the other handlers. Essentially you can take another qprogram fixture, wrap some parts with `with qp.block():` and check that the Q1ASM does not change. Or assert the Q1ASM for this QProgram directly.

src/qililab/qprogram/quantum_machines_compiler.py
```diff
@@ -137,6 +137,7 @@ def __init__(self) -> None:
             ResetPhase: self._handle_reset_phase,
             Wait: self._handle_wait,
             Sync: self._handle_sync,
+            Block: self._handle_block,
```
fedonman, line 140: We should add a test for this as well: Wrap a fixture with the `with qp.block:` and assert the QUA remains unchanged.

jjmartinezQT: yes, I was just adding tests but have no finished yet, also the unittest for QM has not been done yet

jjmartinezQT: Tests done @fedonman

fedonman (approved):
great work 💯

PR #841, opened 2024-11-28
Discovered bugg

GuillermoAbadLopez:


jjmartinezQT (changes requested):
This is not a bug, please check latest Qililab changelog.

PR #842, opened 2024-12-02
Live plotting

jordivallsq:
Added live plotinside experiment class


src/qililab/qprogram/experiment_executor.py
```diff
@@ -553,11 +553,17 @@ def execute(self) -> str:
         # Create file path to store results
         results_path = self._create_results_path(executed_at=executed_at)
 
+        # TODO: Find an alternative way of getting the experiment path at the beginning of the experiment execution
+        if self.live_plot:
+            print(results_path)
```
jjmartinezQT, line 558: This print statement should be deleted?

src/qililab/qprogram/experiment_executor.py
```diff
@@ -104,9 +104,10 @@ class ExperimentExecutor:
         - The results will be saved in a timestamped directory within the `base_data_path`.
     """
 
-    def __init__(self, platform: "Platform", experiment: Experiment):
+    def __init__(self, platform: "Platform", experiment: Experiment, live_plot: bool = False):
```
jjmartinezQT, line 107: Do we need this flag? should not be by default the experimental option? Have we checked with the experimentalists on the behaviour of this future when it comes to turning it on and off, if needed to be turn on and off like this flag does?

jjmartinezQT (changes requested):
We need to produce an example of usage so Oscar can review this functionality and give feedback before merging into main.

src/qililab/qprogram/experiment_executor.py
jordivallsq: This print statement is the temporary "solution" to get the actual data required for the live plotting. The live plot currently works in a separate kernel (but I can also implement it to print at the same cell as it is being executed.

jjmartinezQT: A print statement cannot be a solution. In any case, it should be a warning or something else but not a print.

src/qililab/qprogram/experiment_executor.py
jordivallsq: That is a very good point, indeed the desired behaviour is to have liveplot, I don't think it will hurt to have an option to disable it but definitely the default option should be True

jjmartinezQT: Yes, let's confirm with Oscar.

src/qililab/qprogram/experiment_executor.py
DZPM: Agreed with Juanjo, a `print()` here is not a good practice.

As the method returns the `results_path`, maybe let the caller print the variable themselves?

src/qililab/qprogram/experiment_executor.py
fedonman: I think we should always allow the users to not use live plotting, and just run an experiment.

src/qililab/result/experiment_results.py
```diff
@@ -107,6 +116,25 @@ def get(self, qprogram: int | str = 0, measurement: int | str = 0) -> tuple[np.n
...
+            np.ndarray: The data array.
+        """
+        if isinstance(qprogram, int):
+            qprogram = f"QProgram_{qprogram}"
+        if isinstance(measurement, int):
+            measurement = f"Measurement_{measurement}"
+
+        data = self.data[qprogram, measurement][()]
+
+        return data
```
fedonman, line 136: This is duplicate code. There is the `get` method that returns data and dimensions. If you need only the data, one can do `data, _ = results.get()`.

jordivallsq: You are right, also right now is not even used, this is a leftover from some tests I did before. Thanks for pointing this out!

src/qililab/result/experiment_live_plot.py
```diff
@@ -0,0 +1,179 @@
...
+                    colorbar_y=1 - (n * (1 / len(dims_dict)) + (1 / (2 * len(dims_dict)))),
+                    colorbar_len=1 / len(dims_dict),
+                )
+                self._live_plot_fig.update_yaxes(title_text=y_labels[0], row=n + 1, col=1)
+
+            else:
+                raise NotImplementedError("3D and higher dimension plots are not supported yet.")
+
+        if self._slurm_execution:
+            server = Flask("Live Plot Server")
```
fedonman, line 113: Do we need a Flask server here? In the documentation tutorial, they don't use it: [https://dash.plotly.com/tutorial](https://dash.plotly.com/tutorial)

ChatGPT agrees too:

image

Let's remove it, since it adds an unnecessary dependency in our side, and let Dash create the Flask server with the version it prefers.

requirements.txt
```diff
@@ -16,3 +16,5 @@ ruamel.yaml==0.18.6
 rich==13.8.0
 submitit==1.5.0
 dill==0.3.9
+dash==2.18.2
+flask>=3.0
```
fedonman, line 20: Let's remove this, see comment below.

src/qililab/result/experiment_live_plot.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        self.path = path
+        self._slurm_execution = slurm_execution
+
+        self.live_plot_dict: dict[tuple[str, str], int] = {}
+
+        self._live_plot_fig: go.Figure | go.FigureWidget
+        self._dash_app: Dash
+
+    def _live_plot_figures(self, dims_dict: dict[tuple[str, str], list]):
```
fedonman, line 56: This should be public method since it is called externally.

src/qililab/result/experiment_live_plot.py
```diff
@@ -0,0 +1,179 @@
...
+        else:
+            warnings.filterwarnings("ignore")
+            display(self._live_plot_fig)
+
+    def stop_execution(self):
+        if self._slurm_execution:
+            self.server.shutdown()
+            self.server_thread.join()
+
+    def _live_plot(self, data: np.ndarray, qprogram_name: str, measurement_name: str):
```
fedonman, line 148: This should be public as well since it is called externally.

fedonman (commented):
Really nice implementation! 💯 💯 Left some comments on first glance, and will do a recheck later.

src/qililab/result/experiment_live_plot.py
jordivallsq: done!

src/qililab/result/experiment_live_plot.py
jordivallsq: Done!

requirements.txt
jordivallsq: Removed

src/qililab/result/experiment_live_plot.py
jordivallsq: Initially I was not using Flask as you see in the example. Flask was used to see pas the tests in conjunction with theads to close the server but it did not work. So I will delete the irrellevant part of the code.

jjmartinezQT:
I added unit tests to comply with code coverage. There is usually always a way to mock almost everything in the code for code coverage to pass. I used chat gpt to help and took less than an hour. Let's make sure we cover as much as possible always in code coverage.

jjmartinezQT (approved):
gw

jjmartinezQT:
Let's put this one up to date with base branch and merge it @jordivallsq , so users start using it and giving feedback.

PR #843, opened 2024-12-02
Checking maximum attenuation in RF modules

jjmartinezQT:
This PR brings a check for the maximum attenuation in the RF modules of the Qblox Cluster. If the attenuation in the runcard is set to a higher value than the maximum allowed, it will raise an Exception so experimentalists are aware of the instrumentation limitations in the rack.

### DIFF
--- a/src/qililab/instruments/qblox/qblox_qrm_rf.py
+++ b/src/qililab/instruments/qblox/qblox_qrm_rf.py
@@ -87,6 +87,14 @@ class QbloxQRMRF(QbloxQRM):
         if parameter == Parameter.LO_FREQUENCY:
             parameter = Parameter.OUT0_IN0_LO_FREQ
 
+        if parameter == Parameter.OUT0_ATT:
+            max_att = self.device._get_max_out_att_0()
+            if value > max_att:
+                raise Exception(
+                    f"`{Parameter.OUT0_ATT}` for this module cannot be higher than {max_att}dB.\n"
+                    "Please specify an attenuation level, multiple of 2, below this value."
+                )
+
         if parameter in self.parameters:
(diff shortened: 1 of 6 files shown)


src/qililab/instruments/qblox/qblox_qrm_rf.py
```diff
@@ -87,6 +87,14 @@ def set_parameter(self, parameter: Parameter, value: ParameterValue, channel_id:
         if parameter == Parameter.LO_FREQUENCY:
             parameter = Parameter.OUT0_IN0_LO_FREQ
 
+        if parameter == Parameter.OUT0_ATT:
+            max_att = self.device._get_max_out_att_0()
+            if value > max_att:
+                raise Exception(
```
DZPM, line 93: DRY: instead of repeating the same Exception in three places, we could implement a custom one.

In example, `AttenuationLevelException()` or `MaxAttenuationException()`

fedonman: That's a good point (outside of this PR too). We have a module `exceptions` for this reason, but we haven't used it at all. There are plenty of places in our code that we should raise custom domain-specific errors.

jjmartinezQT: Yeah, saw that module and I am creating a custom exception there for wrong set parameters values, a bit generic but good enough I think to begin with.

jjmartinezQT: I was creating a custom exception but it goes too deep into changes to be distracted now from this quick PR plus calibrating Contralto. We should tackle custom exceptions in a PR for all instruments and parameters.

PR #845, opened 2024-12-03
Module of spi rack is now not a hidden variable

jordivallsq:
For hardware is usefull to access the qcodes parameter of the SPI module (for now they are using this for Xtalk). And so it has to be a public function.


src/qililab/instrument_controllers/qblox/qblox_spi_rack_controller.py
```diff
@@ -75,7 +75,7 @@ def _check_supported_modules(self):
                     + f"and {InstrumentTypeName.QBLOX_S4G}."
                 )
 
-    def _module(self, module_id: int):
+    def module(self, module_id: int):
```
fedonman, line 78: Just a comment: Your code is perfectly fine and approved. But I think that method `module()` is wrong at its core. The `modules` attribute is `modules: Sequence[QbloxD5a | QbloxS4g]`, where both `QbloxD5a` and `QbloxS4g` are our instrument classes. So if you call a `get_module()` method you would expect to get these classes, not the QCoDeS hidden classes. All communication with QCoDeS (or other) drivers should be only with the `device` attribute that all instruments and instrument controllers have.

So in general, I would advise of not using this method to get the QCoDeS instrument. Instead get the qililab instrument and use `.device.`

fedonman (approved):
Approved 💯 , but left a general comment.

src/qililab/instrument_controllers/qblox/qblox_spi_rack_controller.py
fedonman: Or better change this method to return our instruments (which is the correct behaviour) and the access the QCoDeS with `device` if you need that. E.g. `spi_rack.get_module(module_id=2).device`

jordivallsq: The thing, and reason why I wanted to change this method is that with Oscar we tried to use device to interact with the Xtalk (Right now with QCodes parameters) for the SPI. We tried to use device but we realize that for this to work we needed a layer above for it to work efficiently. I agree with your point of view but I think that until we handle the Xtalk we need to use module.
I can set a TODO to make it private again when we have handled the xtalk

jordivallsq, line 78:
```suggestion
    # TODO: once the Xtalk is handled by qililab module has to be a private method
    def module(self, module_id: int):
```

jordivallsq: Added a simple TODO for after the Xtalk is implemented

PR #852, opened 2024-12-09
setting to zero only dacs from runcard

jjmartinezQT:
This PR fixes the bug where all dac channels are set to zero in the current and voltage sources when turning the instrument off.


fedonman (commented):
Should we do the same in the `reset()` method?

jjmartinezQT:
> Should we do the same in the `reset()` method?

I guess so, first I was thinking if reset should remain all instrument proper reset, but I cannot see any advantage on it so will do, thanks for spotting.

fedonman (approved):
90oysg

PR #853, opened 2024-12-13
Qhc 736 bug wrong array output for qbloxresult from qprogram

jjmartinezQT:
This PR reshapes QBlox acquisitions so they are standard with other vendors acquisitions when having nested loops, i.e.:

outer_loop = 51
inner_loop = 11

old_acquisitions.shape(561, 2)
new_acquisitions.shape(51, 11, 2)


src/qililab/result/qprogram/measurement_result.py
```diff
@@ -26,8 +26,9 @@ class MeasurementResult(ABC):
 
     name: ResultName
 
-    def __init__(self, bus: str):
+    def __init__(self, bus: str, shape=None):
```
fedonman, line 29: add typings for `shape`.

src/qililab/result/qprogram/qblox_measurement_result.py
```diff
@@ -39,9 +39,10 @@ class QbloxMeasurementResult(MeasurementResult):
 
     name = ResultName.QBLOX_QPROGRAM_MEASUREMENT
 
-    def __init__(self, bus: str, raw_measurement_data: dict):
+    def __init__(self, bus: str, raw_measurement_data: dict, shape=None):
```
fedonman, line 42: same here.

src/qililab/result/qprogram/qblox_measurement_result.py
```diff
@@ -52,7 +53,8 @@ def array(self) -> np.ndarray:
         """
         path0 = self.raw_measurement_data["bins"]["integration"]["path0"]
         path1 = self.raw_measurement_data["bins"]["integration"]["path1"]
-        return np.array([path0, path1])
+
+        return np.array([path0, path1]) if self.shape is None else np.array([path0, path1]).reshape(*self.shape)
```
fedonman, line 57: If you reshape the whole array there should be a `2` in front for the I/Q: `reshape(2, *self.shape)`.

fedonman: Also I believe is more clear to do it in two steps. This will also trigger the codecov that tests are missing.

```suggestion
        array = np.array([path0, path1])
        if self.shape:
               array = array.reshape(2, *self.shape)
        return array
```

src/qililab/result/qprogram/qblox_measurement_result.py
```diff
@@ -61,4 +63,4 @@ def threshold(self) -> np.ndarray:
         Returns:
             np.ndarray: The thresholded data.
         """
-        return np.array(self.raw_measurement_data["bins"]["threshold"])
+        return np.array(self.raw_measurement_data["bins"]["threshold"]) if self.shape is None else np.array(self.raw_measurement_data["bins"]["threshold"]).reshape(*self.shape)
```
fedonman, line 66: same here for the multiline comment.

tests/result/qprogram/test_qblox_measurement_result.py
```diff
@@ -37,6 +37,7 @@ def test_array_property(self, raw_measurement_data: dict, qblox_measurement_resu
         path0 = raw_measurement_data["bins"]["integration"]["path0"]
         path1 = raw_measurement_data["bins"]["integration"]["path1"]
         expected_array = np.array([path0, path1])
+        print(f"expected_array shape {expected_array.shape}")
```
fedonman, line 40: print statement

tests/instruments/qblox/test_qblox_qrm.py
```diff
@@ -394,8 +394,8 @@ def test_acquire_qprogram_results(self, qrm: QbloxQRM):
         qrm.upload_qpysequence(qpysequence=sequence, channel_id=0)
 
         qp_acqusitions = {
-            "acquisition_0": AcquisitionData(bus="readout_q0", save_adc=False),
-            "acquisition_1": AcquisitionData(bus="readout_q0", save_adc=True)
+            "acquisition_0": AcquisitionData(bus="readout_q0", save_adc=False, shape=(-1,)),
+            "acquisition_1": AcquisitionData(bus="readout_q0", save_adc=True, shape=(-1,))
```
fedonman, line 398: We should add tests that cover no shape, 1D shape and multiloop shape and check both `array()` and `threshold` methods.

fedonman (changes requested):
Left some comments.

src/qililab/result/qprogram/measurement_result.py
jjmartinezQT: Addressed.

jjmartinezQT:
> Another remark: Take note that QuantumMachines in case there is 1 loop let's say 10 iterations, returns an array with shape (2, 1, 10) and not (2, 10)

Then we should adopt this format right? So both results have the same one.

fedonman:
> > Another remark: Take note that QuantumMachines in case there is 1 loop let's say 10 iterations, returns an array with shape (2, 1, 10) and not (2, 10)
> 
> Then we should adopt this format right? So both results have the same one.

Yes, we should have the same shapes for both vendors.

tests/instruments/qblox/test_qblox_qrm.py
jjmartinezQT: added some tests to cover that.

jjmartinezQT:
> > > Another remark: Take note that QuantumMachines in case there is 1 loop let's say 10 iterations, returns an array with shape (2, 1, 10) and not (2, 10)
> > 
> > 
> > Then we should adopt this format right? So both results have the same one.
> 
> Yes, we should have the same shapes for both vendors.

Cool, I will re-adapt QBlox then, thanks.

fedonman:
> > > > Another remark: Take note that QuantumMachines in case there is 1 loop let's say 10 iterations, returns an array with shape (2, 1, 10) and not (2, 10)
> > > 
> > > 
> > > Then we should adopt this format right? So both results have the same one.
> > 
> > 
> > Yes, we should have the same shapes for both vendors.
> 
> Cool, I will re-adapt QBlox then, thanks.

I double checked and is the same as you have it now. sorry for the confusion

fedonman:
> > > > > Another remark: Take note that QuantumMachines in case there is 1 loop let's say 10 iterations, returns an array with shape (2, 1, 10) and not (2, 10)
> > > > 
> > > > 
> > > > Then we should adopt this format right? So both results have the same one.
> > > 
> > > 
> > > Yes, we should have the same shapes for both vendors.
> > 
> > 
> > Cool, I will re-adapt QBlox then, thanks.
> 
> I double checked and is the same as you have it now. sorry for the confusion

So (2, N) for one loop, (2, M, N) for nested loop etc.

jjmartinezQT:
> > > > > > Another remark: Take note that QuantumMachines in case there is 1 loop let's say 10 iterations, returns an array with shape (2, 1, 10) and not (2, 10)
> > > > > 
> > > > > 
> > > > > Then we should adopt this format right? So both results have the same one.
> > > > 
> > > > 
> > > > Yes, we should have the same shapes for both vendors.
> > > 
> > > 
> > > Cool, I will re-adapt QBlox then, thanks.
> > 
> > 
> > I double checked and is the same as you have it now. sorry for the confusion
> 
> So (2, N) for one loop, (2, M, N) for nested loop etc.

Music to my ears then! Will fix the unittest failing then.

fedonman (approved):
Awesome 💯 💯 

770eb057449582c44672b2d2d1cc39b6

PR #856, opened 2024-12-17
Routing fixes

GuillermoAbadLopez:
Changes regarding the `digital_compilation_settings` in the `Transpiler` instantiation

### DIFF
--- a/src/qililab/digital/circuit_transpiler.py
+++ b/src/qililab/digital/circuit_transpiler.py
@@ -44,10 +44,10 @@ class CircuitTranspiler:
 
     def __init__(self, settings: DigitalCompilationSettings | Any):  # type: ignore # ignore typing to avoid importing platform and causing circular imports
         if isinstance(settings, DigitalCompilationSettings):
-            self.digital_compilation_settings = settings
+            self.settings = settings
         else:
             try:
-                self.digital_compilation_settings = settings.digital_compilation_settings
+                self.settings = settings.digital_compilation_settings
             except AttributeError as e:
                 raise ValueError("`setting`s must be a `DigitalCompilationSettings` or `Platform`.") from e
 
@@ -214,7 +214,7 @@ class CircuitTranspiler:
             ValueError: If StarConnectivity Placer and Router are used with non-star topologies.
         """
         # Get the chip's connectivity
-        topology = nx.Graph(coupling_map if coupling_map is not None else self.digital_compilation_settings.topology)
+        topology = nx.Graph(coupling_map if coupling_map is not None else self.settings.topology)
 
         circuit_router = CircuitRouter(topology, placer, router)
 
@@ -274,7 +274,7 @@ class CircuitTranspiler:
         Returns:
             Circuit: Circuit with optimized transpiled gates.
         """
-        optimizer = CircuitOptimizer(self.digital_compilation_settings)
+        optimizer = CircuitOptimizer(self.settings)
 
         output_circuit = Circuit(circuit.nqubits)
         output_circuit.add(optimizer.optimize_transpilation(circuit))
@@ -304,5 +304,5 @@ class CircuitTranspiler:
         Returns:
             list[PulseSequences]: List of :class:`PulseSequences` classes.
         """
-        circuit_to_pulses = CircuitToPulses(self.digital_compilation_settings)
+        circuit_to_pulses = CircuitToPulses(self.settings)
         return [circuit_to_pulses.run(circuit) for circuit in circuits]
--- a/src/qililab/digital/circuit_transpiler.py
+++ b/src/qililab/digital/circuit_transpiler.py
@@ -48,12 +48,11 @@ class CircuitTranspiler:
 
     def __init__(self, settings: Union[DigitalCompilationSettings, "Platform"]):
         if isinstance(settings, DigitalCompilationSettings):
-            self.settings = settings
+            self.settings: DigitalCompilationSettings = settings
+        elif hasattr(settings, "digital_compilation_settings") and settings.digital_compilation_settings is not None:
+            self.settings = settings.digital_compilation_settings
         else:
-            try:
-                self.settings = settings.digital_compilation_settings
-            except AttributeError as e:
-                raise ValueError("`setting`s must be a `DigitalCompilationSettings` or `Platform`.") from e
+            raise ValueError("`setting`s must be a `DigitalCompilationSettings` or a `Platform` with them defined.")
 
     def transpile_circuits(
         self,
--- a/src/qililab/digital/circuit_transpiler.py
+++ b/src/qililab/digital/circuit_transpiler.py
@@ -14,6 +14,8 @@
 
 """Circuit Transpiler class"""
 
+from __future__ import annotations
+
 from typing import TYPE_CHECKING, Union
 
 import networkx as nx
@@ -47,7 +49,7 @@ class CircuitTranspiler:
             It can be both directly the `DigitalCompilationSettings` object or a `Platform` object with the settings defined in it.
     """
 
-    def __init__(self, settings: Union[DigitalCompilationSettings, "Platform"]):
+    def __init__(self, settings: Union[DigitalCompilationSettings, Platform]):
         if isinstance(settings, DigitalCompilationSettings):
             self.settings: DigitalCompilationSettings = settings
         elif hasattr(settings, "digital_compilation_settings") and settings.digital_compilation_settings is not None:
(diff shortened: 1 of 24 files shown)


src/qililab/digital/circuit_transpiler.py
```diff
@@ -37,11 +39,17 @@ class CircuitTranspiler:
...
-    def __init__(self, digital_compilation_settings: DigitalCompilationSettings):  # type: ignore # ignore typing to avoid importing platform and causing circular imports
-        self.digital_compilation_settings = digital_compilation_settings
+    def __init__(self, settings: DigitalCompilationSettings | Any):  # type: ignore # ignore typing to avoid importing platform and causing circular imports
+        if isinstance(settings, DigitalCompilationSettings):
+            self.digital_compilation_settings = settings
+        else:
+            try:
+                self.digital_compilation_settings = settings.digital_compilation_settings
+            except AttributeError as e:
+                raise ValueError("`setting`s must be a `DigitalCompilationSettings` or `Platform`.") from e
```
fedonman, line 52: Why do you allow passing settings as `Platform`? `DigitalCompilationSettings` has everything you need, no?

GuillermoAbadLopez: Yess!

But I think, for users that want to build the `Transpiler` to do routing/transpiling manually, its going to be much easier if they can just instantiate it with the platform they are used to.

Since they will never construct only a `DigitalCompilationSettings`, but always the `Platform`, and then they would need to pass the `platform.digital_compilation_settings`... So I accepted both, and select it for them :)

GuillermoAbadLopez: **Two more things:**
1) At the start, I was just correcting, only having the `Platform` -> The `DigitalCompilationSettings`, and then I decided to keep both. But if we don't want that, at least, we need to change `Platform` -> `DigitalCompilationSettings` here.
2) And the `Any`, comes from `Platform` causing circular import error :)

fedonman: The general principle is that classes should only accept parameters in their `__init__` method that are essential for their operation. For instance, if someone wants to use the `Transpiler` independently—such as to compile a circuit and inspect the result—they shouldn’t be forced to build the entire `Platform` object. Instead, they should only need the relevant settings. If a `Platform` object is already available, its `platform.digital_compilation_settings` can simply be passed as the parameter, as you correctly mentioned.

The circular import issue you encountered is also a sign that the class hierarchy needs improvement. When `Platform` imports `Transpiler` and `Transpiler` imports `Platform`, it creates a problematic dependency cycle. To work around this, you're resorting to using `Any`, which is also discouraged and should only be used in exceptional cases (e.g., when dealing with truly dynamic types that we don't know in compile time).

While renaming the parameter to `settings` is perfectly fine, please revert the rest of the changes to the previous state. 🙏

fedonman: You could also rename the class attribute to just `settings`.

GuillermoAbadLopez: > they shouldn’t be forced to build the entire Platform object.
I mean, but they aren't, because I was accepting both `Platform` and the `DigitalCompilationSettings`, so they are not forced :)

> you're resorting to using `Any`, which is also discouraged and should only be used in exceptional cases 
That I agree with, it was an ugly quick solution, which I shouldn't have done

GuillermoAbadLopez: > You could also rename the class attribute to just settings.

Like the idea, doing it right now 👌

GuillermoAbadLopez: @fedonman Comming back to the discussion about accepting only `DigitalCompilationSettings` or both that and `Platfom`, the difference for the user in the instantiation of a `Transpiler` will be this:

With both (right now):
```python
transpiler = Transpiler(platform)
routed_c, layout = transpiler.route_circuit(c, ...)
```
vs, what you are asking:
```python
transpiler = Transpiler(platform.digital_compilation_settings)
routed_c, layout = transpiler.route_circuit(c, ...)
```
if you think it's okey anyway, I'll change it tomorrow 👍

src/qililab/digital/circuit_transpiler.py
```diff
@@ -89,20 +101,20 @@ def transpile_circuits(
             platform = build_platform(runcard="<path_to_runcard>")
 
             # Create transpiler:
-            transpiler = CircuitTranspiler(platform)
+            transpiler = CircuitTranspiler(platform)  # Or CircuitTranspiler(platform.digital_compilation_settings)
```
fedonman, line 104: change this

src/qililab/digital/circuit_transpiler.py
```diff
@@ -37,20 +43,26 @@ class CircuitTranspiler:
...
+        else:
+            raise ValueError("`setting`s must be a `DigitalCompilationSettings` or a `Platform` with them defined.")
 
     def transpile_circuits(
         self,
         circuits: list[Circuit],
-        placer: Placer | type[Placer] | tuple[type[Placer], dict] | None = None,
-        router: Router | type[Router] | tuple[type[Router], dict] | None = None,
+        placer: Union["Placer", type["Placer"], tuple[type["Placer"], dict], None] = None,
+        router: Union["Router", type["Router"], tuple[type["Router"], dict], None] = None,
```
fedonman, line 62: In general we prefer the `|` notation. If you get conflicts because of the strings add `from __future__ import annotations` as the first import and you don't need the strings any more 😄

src/qililab/digital/circuit_transpiler.py
```diff
@@ -14,20 +14,26 @@
...
 from qililab.settings.digital.digital_compilation_settings import DigitalCompilationSettings
 
 from .gate_decompositions import translate_gates
 
+if TYPE_CHECKING:
+    from qibo.transpiler.placer import Placer
+    from qibo.transpiler.router import Router
+
+    from qililab.platform.platform import Platform
+    from qililab.pulse.pulse_schedule import PulseSchedule
```
fedonman, line 35: nice!

the-mandalorian-this-is-the-way

src/qililab/digital/circuit_transpiler.py
```diff
@@ -142,8 +154,8 @@ def transpile_circuits(
     def route_circuit(
         self,
         circuit: Circuit,
-        placer: Placer | type[Placer] | tuple[type[Placer], dict] | None = None,
-        router: Router | type[Router] | tuple[type[Router], dict] | None = None,
+        placer: Union["Placer", type["Placer"], tuple[type["Placer"], dict], None] = None,
+        router: Union["Router", type["Router"], tuple[type["Router"], dict], None] = None,
```
fedonman, line 158: same here, the previous syntax will work with the `from __future__ import annotations`.

src/qililab/digital/circuit_transpiler.py
```diff
@@ -266,13 +278,13 @@ def optimize_transpilation(self, circuit: Circuit) -> Circuit:
...
         """
-        optimizer = CircuitOptimizer(self.digital_compilation_settings)
+        optimizer = CircuitOptimizer(self.settings)
 
         output_circuit = Circuit(circuit.nqubits)
         output_circuit.add(optimizer.optimize_transpilation(circuit))
         return output_circuit
 
-    def circuit_to_pulses(self, circuits: list[Circuit]) -> list[PulseSchedule]:
+    def circuit_to_pulses(self, circuits: list[Circuit]) -> list["PulseSchedule"]:
```
fedonman, line 287: same here, you won't need the string.

fedonman (commented):
Minor changes, and is good to go 🚀

src/qililab/digital/circuit_transpiler.py
fedonman: Although I think you shouldn't even need the annotations import in this case. Since it is external class (usually you need it for self-referencing).

GuillermoAbadLopez: Ohh nice to know! I'll change it right away 👌 

I thought the `from __future__ import annotations` only worked for classes defined in the same file, great!

src/qililab/digital/circuit_transpiler.py
GuillermoAbadLopez: okei, I'll leave only the first, and with the docstring it should be understandable 👌

src/qililab/digital/circuit_transpiler.py
fedonman: Yes, please change back to `DigitalCompilationSettings` only.

GuillermoAbadLopez: Done! 🔥

fedonman (approved):
italy-approved

PR #861, opened 2024-12-20
[QHC-831] Implement Square Waveform Optmization

fedonman:
We have introduced an optimization in the QbloxCompiler that significantly reduces memory usage when compiling square waveforms. The compiler now uses a heuristic algorithm that segments long waveforms into smaller chunks and loops over them. This optimization follows a two-pass search:

  1. **First Pass**: The compiler tries to find a chunk duration that divides the total waveform length evenly (i.e., remainder = 0).
  1. **Second Pass**: If no exact divisor is found, it looks for a chunk duration that leaves a remainder of at least 4 ns. This leftover chunk is large enough to be stored or handled separately.

  Each chunk duration is restricted to the range (\[100, 500\]) ns, ensuring that chunks are neither too small (leading to excessive repetitions) nor too large (risking out-of-memory issues). If no duration within (\[100, 500\]) ns meets these remainder constraints, the compiler defaults to using the original waveform in its entirety.

### DIFF
--- a/src/qililab/qprogram/qblox_compiler.py
+++ b/src/qililab/qprogram/qblox_compiler.py
@@ -717,27 +717,27 @@ class QbloxCompiler:
 
     @staticmethod
     def calculate_square_waveform_optimization_values(duration):
-        def remainder_conditions(pd):
-            rem = duration % pd
-            return rem, (pd >= 4 and (rem == 0 or rem >= 4))
+        def remainder_conditions(chunk_duration):
+            remainder = duration % chunk_duration
+            return remainder, (chunk_duration >= 4 and (remainder == 0 or remainder >= 4))
 
-        def find_piece_duration(condition_func):
-            for pd in range(100, 501):
-                if pd <= duration:
-                    rem, valid = remainder_conditions(pd)
-                    if valid and condition_func(rem):
-                        return pd
+        def find_chunk_duration(condition_func):
+            for chunk_duration in range(100, 501):
+                if chunk_duration <= duration:
+                    remainder, valid = remainder_conditions(chunk_duration)
+                    if valid and condition_func(remainder):
+                        return chunk_duration
             return None
 
         # First try for remainder == 0
-        final_pd = find_piece_duration(lambda rem: rem == 0)
-        if final_pd is not None:
-            return final_pd, duration // final_pd, duration % final_pd
+        final_chunk_duration = find_chunk_duration(lambda rem: rem == 0)
+        if final_chunk_duration is not None:
+            return final_chunk_duration, duration // final_chunk_duration, duration % final_chunk_duration
 
         # If not found, try for remainder ≥ 4
-        final_pd = find_piece_duration(lambda rem: rem >= 4)
-        if final_pd is not None:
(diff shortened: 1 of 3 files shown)


jjmartinezQT:
> _ No description provided. _

A description would be good :D altho this one I know

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -591,19 +595,59 @@ def _handle_acquire(self, element: Acquire):
     def _handle_play(self, element: Play):
         waveform_I, waveform_Q = element.get_waveforms()
         waveform_variables = element.get_waveform_variables()
-        if not waveform_variables:
-            index_I, index_Q, duration = self._append_to_waveforms_of_bus(
+        if waveform_variables:
+            logger.error("Variables in waveforms are not supported in Qblox.")
```
jjmartinezQT, line 599: nice one

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -669,3 +713,30 @@ def _hash_waveform(waveform: Waveform):
...
+        if final_pd is not None:
+            return final_pd, duration // final_pd, duration % final_pd
+
+        # If not found, try for remainder ≥ 4
+        final_pd = find_piece_duration(lambda rem: rem >= 4)
+        if final_pd is not None:
+            return final_pd, duration // final_pd, duration % final_pd
+
+        # If no suitable piece_duration found, fallback to entire duration
+        return duration, 1, 0
```
jjmartinezQT, line 742: Do you think is worthy raising a warning here telling the user it tried to optimise for values that do not allow optimisation?

jjmartinezQT (approved):
Great improve. I approve it because it is up to the user to use this functionality and by default everything else will work as before, so we can allow this one to be merged before testing in hardware I think, but please test it as soon as we are back from xmas and fridges are cold.

fedonman:
> > _ No description provided. _
> 
> A description would be good :D altho this one I know

Yeah I will write the changelog and make some minor changes on Monday.

jjmartinezQT:
> > > _ No description provided. _
> > 
> > 
> > A description would be good :D altho this one I know
> 
> Yeah I will write the changelog and make some minor changes on Monday.

Cool, thanks!

src/qililab/qprogram/qblox_compiler.py
fedonman: not really, this is just a fallback e.g. for small waveforms like 40ns that will stay the same.

jjmartinezQT: Cool, good on my side once changelog and codecoverage have been sorted. good job.

jjmartinezQT (approved):
good one, make sure this does not fall forgotten in getting spread because of xmas :D We can start including this kind of changes in the tech newsletter

fedonman:
> good one, make sure this does not fall forgotten in getting spread because of xmas :D We can start including this kind of changes in the tech newsletter

Added changelog. For some reason the fallback return is not counted in codecov although I have added cases I have checked they do fallback. I will ignore it and merge it.

PR #862, opened 2024-12-23
[QHC-844] Improving digital Transpilation

GuillermoAbadLopez:
## What this PR does:
(each of the next points is pretty big deal, so check them all please 🙌🙃)
- [x] Move `optimize` flag, for actual optional optimizations (& Improve `optimize` word use in methods names)
- [x] Make `Transpilation`/`execute`/`compile` only for single circuits (unify code standard across `qililab`)
- [x] Make `Transpilation` efficient, by not constructing the `Circuit` class so many times, between methods
- [x] Pass a transpilation `kwargs` instead of so many args in `platform`/`qililab`'s `execute(...)`
- [x] Improve documentation on transpilation, simplifying it in `execute()`'s, and moving it to a new section.
- [x] Make the deletion of gates with no amplitude (fill delete_gates_with_no_amplitude(...))
- [x] Make bunching of consecutive Drag Gates if phis equal (fill bunch_drag_gates(... only_same_phi=True))
- [x] Cover/change test for all of this.TODOs in this PR:

---
### Solve more buggs about final layout/compute it yourself?
- [x] Check if Update qibo version solves layout problems
- [x] Update code to qibo version update
- [x] Solve bugs that made reported layout not works always
- [x] Extracted GateHandler methods from `CircuitToPulses`, `CircuitOptimizer` & `CircuitRouter` into a private class `_GateHandler` in `native_gates`.
- [x] Solve errors for StarConnectivity (The initial_layout wrong, final_layout dunno) -> Not undoing wire_names

### Improve readability of layout:
- [x] Make more clear what are keys/values in the dict
- [x] State clearly the final ordering of qubits ([@Ameer](https://qilimanjaro.slack.com/team/U04NW0FKTA8))
- [x] Maybe log both init_layout (placement) & final_layout (placement+routing)

---
### Future PRs TODOs:v
2) And https://github.com/qilimanjaro-tech/qililab/pull/868 (for different phis, if we end-up doing)


GuillermoAbadLopez:
### The documentation after this PR looks like:
Screenshot 2024-12-30 at 12 05 04
Screenshot 2024-12-30 at 12 05 17

src/qililab/digital/circuit_router.py
```diff
@@ -70,6 +70,7 @@ def __init__(
         # 3) Layout stage, where the initial_layout will be created.
 
     def route(self, circuit: Circuit, iterations: int = 10) -> tuple[Circuit, dict[str, int]]:
+        # Docstring related to the public method: :meth:`.CircuitTranspiler.route_circuit()`. Change it there too.
```
fedonman, line 73: Is this comment needed here?

src/qililab/digital/circuit_to_pulses.py
```diff
@@ -44,7 +43,8 @@ def __init__(self, settings: DigitalCompilationSettings):
         self.settings: DigitalCompilationSettings = settings
         """Object containing the digital compilations settings and the info on chip's physical qubits."""
 
-    def run(self, circuit: Circuit) -> PulseSchedule:
+    def run(self, gate_list: list[gates.Gate]) -> PulseSchedule:
+        # Docstring related to the public method: :meth:`.CircuitTranspiler.gates_to_pulses()`. Change it there too.
```
fedonman, line 47: Is this comment needed here?

src/qililab/digital/circuit_transpiler.py
```diff
@@ -52,31 +52,41 @@ def __init__(self, settings: DigitalCompilationSettings):
...
-        circuits: list[Circuit],
+        circuit: Circuit,
         routing: bool = False,
         placer: Placer | type[Placer] | tuple[type[Placer], dict] | None = None,
         router: Router | type[Router] | tuple[type[Router], dict] | None = None,
         routing_iterations: int = 10,
-        optimize: bool = True,
-    ) -> tuple[list[PulseSchedule], list[dict]]:
+        optimize: bool = False,
+    ) -> tuple[PulseSchedule, dict]:
```
fedonman, line 66: Add typing details for the returned `dict`, like `dict[str, int]`

docs/fundamentals/transpilation.rst
```diff
@@ -1,2 +1,103 @@
+.. _transpilation:
+
 Transpilation
 =============
+
```
fedonman, line 5: Nice!!

fedonman (changes requested):
Nice job! Left some comments.

src/qililab/execute_circuit.py
```diff
@@ -25,61 +23,47 @@
...
 def execute(
-    program: Circuit | list[Circuit],
-    runcard: str | dict,
-    nshots: int = 1,
-    routing: bool = False,
-    placer: Placer | type[Placer] | tuple[type[Placer], dict] | None = None,
-    router: Router | type[Router] | tuple[type[Router], dict] | None = None,
-    routing_iterations: int = 10,
-    optimize: bool = True,
+    program: Circuit | list[Circuit], runcard: str | dict, nshots: int = 1, transpile_config: dict = {}
```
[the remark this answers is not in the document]
GuillermoAbadLopez: > Also, default parameter values should not be mutable: https://docs.astral.sh/ruff/rules/mutable-argument-default/

Damn, true... I guess I'll have to set it back to `None` then... started like that, but that made the code much uglier, with more checks, that complicated it... 😅. But definitely can go back to it, makes sense 👌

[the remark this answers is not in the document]
GuillermoAbadLopez: > I don't like replacing the parameters with a generic dictionary. Now, it is not clear what the dictionary should contain, what are the default values, there is no autocomplete etc.
> 
> In general, having all options as separate parameters is preferred.
> 
> If we really do need to use a single parameter, we should use a NamedDict or a new class.

Yeah I didn't like it either, but it was so many parameters, and might keep growing..., all of them related to the same thing, that in the end discussing it with @visagim we thought, it didn't make sense to keep the params separated 😓 

I also tried first to do a default class, but what I didn;t like about it, is the user having to import and initialize a random class, from a file in `qililab`, dunno how intuitive that would be 😅. But its true that at least you would then have initialization. 

Dunno I'm not 100% convinced what is better, I can try to generate the three examples:
- Keep params separated
- Ganeric kwargs
- Specific qililab dataclass

and ask again in a Code Discussion/daily, where everybody can see it and decide 👍 :

src/qililab/digital/circuit_router.py
GuillermoAbadLopez: Yeah, good point... I actually have a problem here, which I wanted to ask about:

If you see, Inside the `Transpiler` we have basically methods duplicated from the `Optimizer`..., since we removed the handling of multiple circuits at the same time.. One option would be to remove such duplication, but at the same time that duplications serves as a kind of interface to add more functionality, and shows the user which functions can be called from the transpiler directly, so its not super bad...

Of course this makes that the docstrings are duplicated, and *that* is more problematic, since we wouldn't; want our devs, only updating the `Optimizer` docstrings, but not the `Transpiler`'s which are more public. So the options we have are:
- Delete the duplication in some way (I think this will complicate how we show public method to the user...)
- Add this silly comment for devs to notice, that they need to change the public documentation always
- Maybe just delete the docstring from the not public methods? So there is no mistake by the devs? And they know to go and search elsewhere, which docstring to change 🤔...

GuillermoAbadLopez: I'll think more about it the next days, and ask the group with a clearer head...

But such problem make me notice there is a structural code problem from base with having this duplication, which maybe should be addressed more deeply... But 😬 ...

src/qililab/digital/circuit_to_pulses.py
GuillermoAbadLopez: Same answer as above, closing this one, to discuss it in the previous one 👌

src/qililab/platform/platform.py
```diff
@@ -986,45 +984,33 @@ def execute(
         repetition_duration: int = 200_000,
         num_bins: int = 1,
         queue: Queue | None = None,
-        routing: bool = False,
-        placer: Placer | type[Placer] | tuple[type[Placer], dict] | None = None,
-        router: Router | type[Router] | tuple[type[Router], dict] | None = None,
-        routing_iterations: int = 10,
-        optimize: bool = True,
+        transpile_config: dict = {},
```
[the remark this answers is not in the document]
GuillermoAbadLopez: Lets discuss it in the above thread, closing this one 👌

src/qililab/digital/circuit_transpiler.py
GuillermoAbadLopez: True, I'll add it 🙌

src/qililab/digital/circuit_router.py
GuillermoAbadLopez: @fedonman, I ended up deciding to delete a bunch of duplicated docstrings, so we don't have confusion on where to update them! 

And deleted the comments you pointed out, instead now I added inside the duplicated docstring itself, check if you agree with the implementation @fedonman 🙌

src/qililab/execute_circuit.py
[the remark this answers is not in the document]
GuillermoAbadLopez: ## UPDATE:
Okey, i ended up creating a `TypedDict` that can be imported for suggestions/autocompletion 🙌 

Also changed the default value to non-mutable: None 🚀

GuillermoAbadLopez:
### After the last changes, the documentation for the `execute`'s & `compile` ends up like:
(it only changed the bottom part, the docstring of the `transpile_config` attribute)
Screenshot 2025-01-13 at 15 02 57

src/qililab/digital/circuit_transpiler.py
```diff
@@ -29,13 +28,29 @@
...
+
+    from qibo import Circuit, gates
     from qibo.transpiler.placer import Placer
     from qibo.transpiler.router import Router
 
     from qililab.pulse.pulse_schedule import PulseSchedule
     from qililab.settings.digital.digital_compilation_settings import DigitalCompilationSettings
 
 
+class DigitalTranspileConfig(TypedDict, total=False):
```
fedonman, line 41: I looked a bit further into this and `TypedDict` is not the way to go. Mainly because the user loses track of the default values. E.g. If I don't provide `routing` is the default True or False? What number is the default `routing_iterations`? I cannot know except if I follow the method callings and find out that the `CircuitTranspiler.transpile_circuit()` uses `routing=False` as default, or `routing_iterations=10`.

Best solution is to make this a `dataclass` with default values, and use the same object in all methods. That means don't do `transpiler.transpile_circuit(program, **(transpile_config or {}))` in line 1182 of Platform, just pass the object: `transpiler.transpile_circuit(program, transpile_config)`

fedonman, line 41: Also, could you rename the class to `DigitalTranspilationConfig` and the variable names to `transpilation_config`. Sorry I am a grammar nazi on naming things 😅 It also aligns with the `DigitalCompilationConfig` we already have.

src/qililab/digital/circuit_transpiler.py
```diff
@@ -106,62 +131,66 @@ def transpile_circuits(
...
+            placer (Placer | type[Placer] | tuple[type[Placer], dict], optional): ``Placer`` instance, or subclass ``type[Placer]`` to
+                use, with optionally, its kwargs dict (other than connectivity), both in a tuple. Defaults to ``ReverseTraversal``.
+            router (Router | type[Router] | tuple[type[Router], dict], optional): ``Router`` instance, or subclass ``type[Router]`` to
+                use, with optionally, its kwargs dict (other than connectivity), both in a tuple. Defaults to ``Sabre``.
             routing_iterations (int, optional): Number of times to repeat the routing pipeline, to get the best stochastic result. Defaults to 10.
             optimize (bool, optional): whether to optimize the circuit and/or transpilation. Defaults to True.
 
         Returns:
-            tuple[list[PulseSchedule],list[dict[str, int]]]: list of pulse schedules and list of the final layouts of the qubits, in each circuit {"qI": J}.
+            tuple[PulseSchedule, dict[str, int]]: Pulse schedule and final layouts of the qubits, in the circuit {"qI": J}.
```
fedonman, line 157: Could you explain more what is the returned `dict`? I can't understand the `{"qI": J}` example.

fedonman (changes requested):
Great work overall 💯 💯  Left some minor comments.

docs/fundamentals/transpilation.rst
```diff
@@ -1,2 +1,104 @@
...
+5. \*\*)Optimizing the resulting Drag gates, by combining multiple pulses into a single one. Refer to the :meth:`.CircuitTranspiler.optimize_transpiled_gates()` method for more information.
+
+6. Pulse Schedule Conversion: Converts the native gates into a pulse schedule using calibrated settings from the runcard. Refer to the :meth:`.CircuitTranspiler.gates_to_pulses()` method for more information.
+
+.. note::
+
+    \*) If ``routing=False`` in ``transpile_config`` (default behavior), step 1. is skipped.
+
+    \*\*) If ``optimize=False`` in ``transpile_config`` (default behavior), steps 2. and 5. are skipped.
```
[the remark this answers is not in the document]
GuillermoAbadLopez: Okey, yes I understand this might not be the best, lemme change it!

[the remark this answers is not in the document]
GuillermoAbadLopez: Changed all of them, and I think now they are much better, let me know what you think 👍 

Screenshot 2025-01-22 at 10 45 50
Screenshot 2025-01-22 at 10 45 30
Screenshot 2025-01-22 at 10 45 41
Screenshot 2025-01-22 at 10 45 41
Screenshot 2025-01-22 at 10 48 44

src/qililab/digital/circuit_to_pulses.py
```diff
@@ -44,34 +43,21 @@ def __init__(self, settings: DigitalCompilationSettings):
         self.settings: DigitalCompilationSettings = settings
         """Object containing the digital compilations settings and the info on chip's physical qubits."""
 
-    def run(self, circuit: Circuit) -> PulseSchedule:
-        """Translates a circuit into a  pulse sequences.
+    def run(self, gate_list: list[gates.Gate]) -> PulseSchedule:
```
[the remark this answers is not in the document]
GuillermoAbadLopez: Right, seems reasonable, I'll change it, but I'll call it `circ_gates` instead, cause gates is an import from qibo, and also this way the name has even more info 👌

src/qililab/digital/circuit_transpiler.py
GuillermoAbadLopez: Yeah. actually i have a full [other PR for this](https://github.com/qilimanjaro-tech/qililab/pull/869) (the next one I'm gonna tackle)

Do you mind if we leave it like this for now, and I address it immediately there? We can even wait until all PRs are approved separately, and merge them together into `main`, so there is no moment, were this is public 👌

src/qililab/digital/circuit_to_pulses.py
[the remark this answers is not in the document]
fedonman: `circ_gates` is an abbreviation though, that we don't want either. `circuit_gates` is the full name.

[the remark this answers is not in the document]
fedonman: However, the logic goes like this: Your own library is more important than the imported library. So if you want to use `gates` and Qibo is already using `gates`, you just rename Qibo's to something else e.g. `from qibo import gates as qibo_gates`. Or use the full path: `import  qibo` and then `qibo.gates.X`.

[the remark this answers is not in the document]
GuillermoAbadLopez: > However, the logic goes like this: Your own library is more important than the imported library. So if you want to use gates and Qibo is already using gates, you just rename Qibo's to something else e.g. from qibo import gates as qibo_gates. Or use the full path: import  qibo and then qibo.gates.X.

Yeah, I actually was gonna do that first, but then I thought that in this case it might be better to have `circ_gates` actually, since all these are lists of gates from the same circuit! So I think it really is more useful here 👌 

I'll change the abbreviation to the full `circuit_gates` 🔥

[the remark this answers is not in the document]
GuillermoAbadLopez: Changed!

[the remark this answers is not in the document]
GuillermoAbadLopez: Also did [this other PR,](https://github.com/qilimanjaro-tech/qililab/pull/881) to `main` changing some extra ones I found, not related to this PR 👌

src/qililab/digital/circuit_transpiler.py
GuillermoAbadLopez: hahaa dun't worry, I think that a great idea for all the company overall TOC :) 🙌

GuillermoAbadLopez: Done! 🔥

GuillermoAbadLopez: just changed to what you said 👍

fedonman (approved):
Great work! 💯 💯 💯 

nodding-yes

fedonman:
@GuillermoAbadLopez Please refrain from adding new commits to a PR once it is approved, because it needs re-approving and is difficult to see which changes are made. Always create a new PR for these cases! 🙏

GuillermoAbadLopez:
Also I will need to add some more changes, since there is a problem with the ordering of the returned results.

it will be more complicated than I though, due to multiple measurements, and measurements not needed to be in the end of a circuit... (But I'll create a new PR pointing here for that 👍 )

fedonman:
@GuillermoAbadLopez Merge this PR, so other PRs point to main please 🙏

GuillermoAbadLopez:
@fedonman can't do it like that, since the commits are already in this branch! 

We would need to:
1) review that PR, before mergin this to `main`
2) Merge the other one, which contains the commits approved alerady, but then that would mess the information/historial the PRs have...
3) Undo the commits here, and merge this one, and create a new one with the new commits... which is what would be more serious, but what would take more time..

I would rather do 1) if we can, we will lose much less time, and I don't think those changes are sucha. big deal, to make us lose time over them

GuillermoAbadLopez:
Also. these changes actually were necessary to be in this PR i believe, since they arise from testing this one in HW, without them this code would be buggy, so I don't think it's a good idea to merge it, and have another PR to solve it, if we can solve it before putting in `main` 👌  (of course I could have done a PR that pointed here, but i didn't 😅 )

So now that we have a first thing that works with these last changes, I won't expand this branch any further (we merge it), I'll create new PR for next improvements 👍

GuillermoAbadLopez:
Also @fedonman . these changes actually were necessary to be in this PR i believe, since they arise from testing this one in HW, without them this code would be buggy, so I don't think it's a good idea to merge it, and have another PR to solve it, if we can solve it before putting in main 👌 (of course I could have done a PR that pointed here, but i didn't 😅 )

So now that we have a first thing that works with these last changes, I won't expand this branch any further (we merge it), I'll create new PR for next improvements 👍

PR #863, opened 2024-12-28
[QHC-845] Make optimization of `Drag` gates

GuillermoAbadLopez:
### TODOs in this PR:
- [x] Make the deletion of gates with no amplitude (fill `delete_gates_with_no_amplitude(...)`)
- [x] Make bunching of consecutive Drag Gates if phis equal (fill `bunch_drag_gates(... only_same_phi=True)`)
- [x] Make tests to cover this, and manually check it works!

---
### And in the [next PR](https://github.com/qilimanjaro-tech/qililab/pull/868), we will follow up on Drag Bunching, for different phis.


src/qililab/digital/circuit_optimizer.py
```diff
@@ -148,37 +148,124 @@ def optimize_transpiled_gates(cls, gate_list: list[gates.Gate]) -> list[gates.Ga
...
+        # TODO: ADD BUNCHING DRAG GATES FOR DIFFERENT PHI's!
+        # gate_list = cls.bunch_drag_gates(gate_list, only_same_phi=False)
+        gate_list = cls.normalize_angles_of_drags(gate_list)
         gate_list = cls.delete_gates_with_no_amplitude(gate_list)
         return gate_list
 
-    @staticmethod
-    def bunch_drag_gates(gate_list: list[gates.Gate]) -> list[gates.Gate]:
+    @classmethod
+    def bunch_drag_gates(cls, gate_list: list[gates.Gate], only_same_phi: bool = False) -> list[gates.Gate]:
```
fedonman, line 159: Why is this a `classmethod` and not a `staticmethod` like `merge_consecutive_drags`? I would prefer it for consistency.

src/qililab/digital/circuit_optimizer.py
```diff
@@ -148,37 +148,124 @@ def optimize_transpiled_gates(cls, gate_list: list[gates.Gate]) -> list[gates.Ga
...
         """Bunches consecutive Drag gates together into a single one.
 
         Args:
             gate_list (list[gates.Gate]): list of gates of the transpiled circuit, to bunch drag gates.
+            only_same_phi (bool, optional): If True, only Drag gates with the same phi are bunched. Defaults to False.
```
fedonman, line 164: Why do we need this flag? We have only implemented the case of same phi.

src/qililab/digital/circuit_optimizer.py
```diff
@@ -148,37 +148,124 @@ def optimize_transpiled_gates(cls, gate_list: list[gates.Gate]) -> list[gates.Ga
...
+        if drag1.qubits != drag2.qubits:
+            raise ValueError("Cannot merge Drag gates acting on different qubits.")
+
+        # For the same phi, we just need to add the theta parameters:
+        if drag1.parameters[1] == drag2.parameters[1]:
+            return Drag(drag1.qubits[0], drag1.parameters[0] + drag2.parameters[0], drag1.parameters[1])
+
+        # If we are merging only same phi, we don't merge gates with phi_1 != phi_2:
+        if only_same_phi:
+            return None
```
fedonman, line 223: Continuing previous comment, this could be removed.

src/qililab/digital/circuit_optimizer.py
```diff
@@ -148,37 +148,124 @@ def optimize_transpiled_gates(cls, gate_list: list[gates.Gate]) -> list[gates.Ga
...
+            only_same_phi (bool): If True, only Drag gates with the same phi are merged.
+
+        Returns:
+            Drag | None: Merged Drag gate. None, if not possible to merge.
+        """
+        if drag1.qubits != drag2.qubits:
+            raise ValueError("Cannot merge Drag gates acting on different qubits.")
+
+        # For the same phi, we just need to add the theta parameters:
+        if drag1.parameters[1] == drag2.parameters[1]:
```
fedonman, line 218: maybe you need to use the `np.isclose` here too?

src/qililab/digital/circuit_optimizer.py
```diff
@@ -148,37 +148,124 @@ def optimize_transpiled_gates(cls, gate_list: list[gates.Gate]) -> list[gates.Ga
...
         Returns:
             list[gates.Gate]: list of gates of the transpiled circuit, with gates without amplitude deleted."""
-        # Add deletion of Drag gates without amplitude here:
-        # gate_list =
+        for gate in gate_list:
+            if isinstance(gate, Drag) and np.isclose(gate.parameters[0], 0):
+                gate_list.remove(gate)
+        return gate_list
+
+    @classmethod
```
fedonman, line 242: This could be a `staticmethod` as well.

src/qililab/digital/circuit_optimizer.py
fedonman: Or the other way around, make the other a `classmethod`.

fedonman (commented):
Awesome work and nice tests! Left some comments.

src/qililab/digital/circuit_optimizer.py
GuillermoAbadLopez: The way I think about it, is the preference is always, in this order:
- staticmethod
- classmethod
- instancemethod (default)

And you have to go down, as you need it, from resources

In this case `merge_consecutive_drags()` can be static, so I made it that, and `bunch_drag_gates()`, can't you need to go to `classmethod`, so I made it that.

src/qililab/digital/circuit_optimizer.py
GuillermoAbadLopez: Leaving all the discussion in the comment above

src/qililab/digital/circuit_optimizer.py
GuillermoAbadLopez: Not really, cause it uses `cls._normalize_angle()` , no? 😬

src/qililab/digital/circuit_optimizer.py
GuillermoAbadLopez: Applied!

src/qililab/digital/circuit_optimizer.py
fedonman: Even if you implement the optimization for different phis, there is no point in having this flag. Just apply the correct optimization.

src/qililab/digital/circuit_optimizer.py
fedonman: Check other comment above.

src/qililab/digital/circuit_optimizer.py
GuillermoAbadLopez: Damn, true... nice to know

I'll change it later, thanks! 👌

GuillermoAbadLopez: Buff this means, there are lots of method in `qililab`, that are class methods, when they could be static in reality 😬

src/qililab/digital/circuit_optimizer.py
GuillermoAbadLopez: Keeping the discussion above 👍

src/qililab/digital/circuit_optimizer.py
GuillermoAbadLopez: The reason of the flag, was to prioritize first a sweep only merging pairs with same phi's, since the math was trivial.
And then doing another sweep to merge everything, which might require very complicated logic.

(Because if we do different phi's together, when the first difference come, it will result in a weird mixed phi's, and that will propagate, and then none of the later, will match, not using the trivial maths anymore... (which in reality since we use RX and RY's mostly, should be used a lot..)

But maybe it doesn't even make sense this, since now I saw that this introduced RZ, which fucks everything up, and we need to rethink how the full implementation works, since we need to loop with the virtualZ stuff again, etc...

---
### In summary: I'll delete it yeah, since the next implementation for diff phi's seems that it will be much different than the previous, but it did have a logic in the past that made the flag useful :)

src/qililab/digital/circuit_optimizer.py
```diff
@@ -162,23 +163,104 @@ def bunch_drag_gates(gate_list: list[gates.Gate]) -> list[gates.Gate]:
...
+                new_drag: Drag | None = CircuitOptimizer.merge_consecutive_drags(drag1, drag2)
+
+                # If successful bunching, substitute old gates with new one:
+                if new_drag is not None:
+                    gate_list[idx1], drag1 = new_drag, new_drag
+                    gate_list[idx2 + idx1 + 1] = None
+
+                # If not successful bunching, we can't merge this gate, go next one:
+                if new_drag is None:
+                    break
```
fedonman, line 194: It's always more clear to have the breaking condition first, and otherwise continue with the execution flow.

```suggestion
                # If not successful bunching, we can't merge this gate, go next one:
                if new_drag is None:
                    break
                 
                gate_list[idx1], drag1 = new_drag, new_drag
                gate_list[idx2 + idx1 + 1] = None
```

fedonman (commented):
Awesome :100: Minor suggestion and is good to go.

src/qililab/digital/circuit_optimizer.py
GuillermoAbadLopez: Nice improvement indeed, thanks!

src/qililab/digital/circuit_optimizer.py
```diff
@@ -162,23 +163,103 @@ def bunch_drag_gates(gate_list: list[gates.Gate]) -> list[gates.Gate]:
...
+
+        # For the same phi, we just need to add the theta parameters:
+        if np.isclose(drag1.parameters[1], drag2.parameters[1]):
+            return Drag(drag1.qubits[0], drag1.parameters[0] + drag2.parameters[0], drag1.parameters[1])
+
+        # TODO: ADD BUNCHING DRAG GATES FOR GENERAL DIFFERENT PHI's!
+        return None  # This should return the merged Drag gate, for different phi's!
 
     @staticmethod
     def delete_gates_with_no_amplitude(gate_list: list[gates.Gate]) -> list[gates.Gate]:
```
visagim, line 220: not sure we want to do this? Technically if a user wants to send a gate with 0 amplitude, for whatever reason, there's no reason to prevent it

src/qililab/digital/circuit_optimizer.py
```diff
@@ -162,23 +163,103 @@ def bunch_drag_gates(gate_list: list[gates.Gate]) -> list[gates.Gate]:
...
-        # Add deletion of Drag gates without amplitude here:
-        # gate_list =
+        for gate in gate_list:
+            if isinstance(gate, Drag) and np.isclose(gate.parameters[0], 0):
+                gate_list.remove(gate)
         return gate_list
 
+    @staticmethod
+    def normalize_angles_of_drags(gate_list: list[gates.Gate]) -> list[gates.Gate]:
+        """Normalizes angles in the gates of the circuit.
```
visagim, line 235: this is already done in the transpiler (circuit to pulses) - you can check what's going on there because i think it wasnt trivial.

visagim:
I don't have a lot of time to do a very comprehensive review (and @fedonman is also reviewing so it should be fine) but i added some comments anyways

src/qililab/digital/circuit_optimizer.py
GuillermoAbadLopez: This is case the Optimization has substracted thetas, and a 0 has remained :).

For example a Rx(theta)Rx(-theta)

src/qililab/digital/circuit_optimizer.py
GuillermoAbadLopez: I could also have imported it, or moved the function to common ground and import it in both places, but I don't think its worth it, for such a simple function, and this way also gives us more flexibility to modify it differently in the future, in each place if we want 🤔

visagim: why not just leave it in circuit to pulses? Or remove it from circuit to pulses and add it here. Ofc the compromise would be to more it out of the circuit to pulses class (maybe in the same file but as an outside method) and import it from there

visagim: but do whatever otherwise, i guess this is still WIP and we will remove redundant code in the future once we have transpilation from circuit to qprogram and no longer need the circuit to pulses class.

GuillermoAbadLopez: And it's like a 5lines code, didn't wanna mess with it, just in case, since the winning is not that huge, but I could invest a bit of time and delete one, but not sure if it's worth it 😅

Also because as you said it will be removed :)

GuillermoAbadLopez: @visagim  I just moved the method `normalize_angle()` into `native_gates`, together with the Drag definition, from where it's called in both places, in this commit: https://github.com/qilimanjaro-tech/qililab/pull/862/commits/5edbf707a6fa8563152d97722e5f9415637e5819, in [the big PR](https://github.com/qilimanjaro-tech/qililab/pull/862) containing now all the changes.

Its still called both in the `ToPulses` and the `Optimizer`, but at least now, there is no code duplication 👌 

I'll test with Adri these next days, and maybe I can remove the call from on of those places, I'll check it 👍

PR #865, opened 2025-01-08
[QHC-840] Raise an error if qdac input values are outside of the range

flavie-lebars:


src/qililab/instruments/qdevil/qdevil_qdac2.py
```diff
@@ -88,6 +88,8 @@ def set_parameter(self, parameter: Parameter, value: ParameterValue, channel_id:
             self.settings.ramping_enabled[index] = ramping_enabled
             if self.is_device_active():
                 if ramping_enabled:
+                    if self.ramp_rate[index] < 0.01 or self.ramp_rate[index] > 2**7:
```
fedonman, line 91: Minor suggestion: When we have magic numbers (like 0.01, and 2e7 in this case) which usually come from hardware constraints, it is preferred to define them as private class constants.

This has two benefits: each value is associated with an explanatory name, e.g. `_MINIMUM_RAMPING_RATE`, and if for some reason the value changes, we only have to change it in one place. It eliminates the risks of typos when using the same value in several places too.

As a reference, this is the definitions for Qblox modules:

```
class QbloxModule(Instrument):
    """Qblox Module class.

    Args:
        device (QcmQrm): Instance of the Qblox QcmQrm class used to connect to the instrument.
        settings (QbloxPulsarSettings): Settings of the instrument.
    """

    _MAX_BINS: int = 131072
    _NUM_MAX_SEQUENCERS: int = 6
    _NUM_MAX_AWG_OUT_CHANNELS: int = 4
    _MIN_WAIT_TIME: int = 4  # in ns
```

fedonman (commented):
Great job and great tests! Left a minor comment, and is ready to go.

tests/instruments/qdevil/test_qdevil_qdac2.py
```diff
@@ -6,7 +7,7 @@
 from qililab.instruments.qdevil.qdevil_qdac2 import QDevilQDac2
 from qililab.typings.enums import Parameter
 from qililab.waveforms import Square
-import re
+from qcodes_contrib_drivers.drivers.QDevil.QDAC2 import QDac2Channel
```
jjmartinezQT, line 10: Now that we do not need the channel this import can be removed I think.

tests/instruments/qdevil/test_qdevil_qdac2.py
```diff
@@ -26,9 +27,36 @@ def fixture_qdac() -> QDevilQDac2:
...
+            "voltage": [0.5],
+            "span": ["low"],
+            "ramping_enabled": [True],
+            "ramp_rate": [0.001],
+            #"ramping_rate": [0.001, 2e7],
+            "dacs": [10],
+            "low_pass_filter": ["dc"],
+        }
+    )
+    #channel = MagicMockQDac2Channel(QDevilQDac2, "channel", 10)
```
jjmartinezQT, line 48: Let's remove this two comments as we are not longer using the channel.

jjmartinezQT (changes requested):

src/qililab/instruments/qdevil/qdevil_qdac2.py
```diff
@@ -35,6 +35,8 @@ class QDevilQDac2(VoltageSource):
         settings (QDevilQDac2Settings): Settings of the instrument.
     """
 
+    _MIN_QDAC: float = 0.01
+    _MAX_QDAC: float = 2e7
```
fedonman, line 39:
```suggestion
    _MIN_RAMPING_RATE: float = 0.01
    _MAX_RAMPING_RATE: float = 2e7
```

Could you rename those two?

fedonman (approved):
Great work!

italy-approved

jjmartinezQT (approved):

PR #871, opened 2025-01-14
[QHC-851] Change `@classmethods` to `@staticmethods`

GuillermoAbadLopez:
### Following the discussions in: https://github.com/qilimanjaro-tech/qililab/pull/863#discussion_r1914690500. I left only the `register` and `get` methods from `factories` as `@classmethods`, all the rest in qililab, are passed to `@staticmethods`:

Screenshot 2025-01-14 at 15 26 22

### All I did was:
- A cmd+F, to change `classmethod` for `staticmethods` and changed all the `cls` by the corresponding classes, so they are not needed (Only left factories with it, since those really use class variables, and change the status of the class :fire:)
- The rest of changes are autoformater changes 👌 

---

### Update:
- Undid the changes for the `from_dict` methods, since those needed for possible inheritances
- Also removed a lot of duplicated code, that I did long ago, for the serializations of `PulseShapes` and `PulseDistortions`

### DIFF
--- a/src/qililab/pulse/pulse_distortion/pulse_distortion.py
+++ b/src/qililab/pulse/pulse_distortion/pulse_distortion.py
@@ -15,6 +15,7 @@
 """PulseDistortion abstract base class."""
 
 from abc import abstractmethod
+from copy import deepcopy
 from dataclasses import dataclass
 
 import numpy as np
@@ -115,27 +116,38 @@ class PulseDistortion(FactoryElement):
             np.ndarray: Distorted pulse envelope.
         """
 
-    @staticmethod
-    def from_dict(dictionary: dict) -> "PulseDistortion":
-        """Loads PulseDistortion object from dictionary.
+    @classmethod
+    def from_dict(cls, dictionary: dict) -> "PulseDistortion":
+        """Loads a `PulseDistortion` object from a dictionary representation.
 
         Args:
-            dictionary (dict): Dictionary representation of the PulseDistortion object. It must include the name of the
-            correction.
+            dictionary (dict): Dictionary representation of the `PulseDistortion` object, containing all its attributes.
+                If called directly from the Abstract parent class: `PulseDistortion`, it must include the child class name to instantiate.
 
         Returns:
-            PulseDistortion: Loaded class.
+            PulseDistortion: Loaded child class, including the name of the pulse distortion and its attributes set.
         """
-        distortion_class = Factory.get(name=dictionary["name"])
-        return distortion_class.from_dict(dictionary)
+        # For calls from parent PulseShape:
+        if cls is PulseDistortion:
+            shape_class = Factory.get(name=dictionary["name"])
+            return shape_class.from_dict(dictionary)
 
-    @abstractmethod
+        # For calls from childs directly:
+        local_dictionary = deepcopy(dictionary)
+        name = local_dictionary.pop("name", None)
+        if name not in [cls.name, None]:
+            raise ValueError(f"Class: {cls.name} to instantiate, does not match the given dict name {name}")
+        return cls(**local_dictionary)
+
     def to_dict(self) -> dict:
-        """Returns dictionary of PulseDistortion.
+        """Returns dictionary representation of `PulseDistortion`. This includes the class name an all its attributes.
 
         Returns:
-            dict: Dictionary describing the pulse distortion.
+            dict: Dictionary describing the `PulseDistortion`, including the name and its attributes.
         """
+        return {"name": self.name.value} | {
+            attribute: getattr(self, attribute) for attribute in self.__dataclass_fields__.keys()
+        }
 
     def normalize_envelope(self, envelope: np.ndarray, corr_envelope: np.ndarray) -> np.ndarray:
         """Normalizes the envelope depending on the `norm_factor` and `auto_norm` attributes.
--- a/src/qililab/pulse/pulse_distortion/pulse_distortion.py
+++ b/src/qililab/pulse/pulse_distortion/pulse_distortion.py
@@ -17,11 +17,14 @@
 from abc import abstractmethod
 from copy import deepcopy
 from dataclasses import dataclass
+from typing import Type, TypeVar
 
 import numpy as np
 
 from qililab.typings.factory_element import FactoryElement
 from qililab.utils import Factory
+
+T = TypeVar("T", bound="PulseDistortion")
 
 
 @dataclass(frozen=True, eq=True, kw_only=True)
@@ -117,7 +120,7 @@ class PulseDistortion(FactoryElement):
         """
 
     @classmethod
-    def from_dict(cls, dictionary: dict) -> "PulseDistortion":
+    def from_dict(cls: Type[T], dictionary: dict) -> T:
         """Loads a `PulseDistortion` object from a dictionary representation.
 
         Args:
--- a/src/qililab/pulse/pulse_distortion/pulse_distortion.py
+++ b/src/qililab/pulse/pulse_distortion/pulse_distortion.py
@@ -135,8 +135,8 @@ class PulseDistortion(FactoryElement):
         # For calls from childs directly:
         local_dictionary = deepcopy(dictionary)
         name = local_dictionary.pop("name", None)
-        if name not in [cls.name, None]:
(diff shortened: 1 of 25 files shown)


src/qililab/qprogram/calibration.py
```diff
@@ -170,8 +170,8 @@ def save_to(self, file: str):
 
         return file
 
-    @classmethod
-    def load_from(cls, file: str):
+    @staticmethod
```
fedonman, line 173: Yeah, I forgot to mention this. Methods that return instances of the class should still remain `@classmethod`. This is because, you can use inheritance (and generics) to define the method in the parent class and also work in child classes. Whereas with `@staticmethod` you have to define the method to all children.

We still don't use much of those, but with the PR on pydantic, generics are heavily introduced in the codebase, so I don't want to change it back.

Check this example to understand:

```
from __future__ import annotations

from typing import Any, Dict, Type, TypeVar

T = TypeVar("T", bound="Parent")


class Parent:
    def __init__(self, name: str, age: int) -> None:
        self.name = name
        self.age = age

    @classmethod
    def from_dict(cls: Type[T], data: Dict[str, Any]) -> T:
        """
        Build an instance of `cls` (which can be `Parent` or a subclass).
        This allows Child classes to inherit the same method if
        their constructor's parameter names match dictionary keys.
        """
        # Directly unpack the dictionary into __init__ using **
        return cls(**data)

    def __repr__(self) -> str:
        return f"Parent(name={self.name!r}, age={self.age})"


class Child(Parent):
    def __init__(self, name: str, age: int, grade: str) -> None:
        """
        Child extends Parent by adding a 'grade' parameter.
        """
        super().__init__(name, age)
        self.grade = grade

    def __repr__(self) -> str:
        return f"Child(name={self.name!r}, age={self.age}, grade={self.grade!r})"


if __name__ == "__main__":
    # Dictionaries that match the constructor signatures
    parent_data = {"name": "Alice", "age": 40}
    child_data = {"name": "Bob", "age": 10, "grade": "5th"}

    # Calling Parent.from_dict will produce a Parent instance
    p = Parent.from_dict(parent_data)
    # Calling Child.from_dict will produce a Child instance
    c = Child.from_dict(child_data)

    print(p)  # Parent(name='Alice', age=40)
    print(c)  # Child(name='Bob', age=10, grade='5th')
```

src/qililab/qprogram/crosstalk_matrix.py
```diff
@@ -43,8 +43,8 @@ def to_dict(self):
         """
         return self.vector
 
-    @classmethod
-    def from_dict(cls, flux_dict: dict[str, float]) -> "FluxVector":
+    @staticmethod
```
fedonman, line 46: here too

src/qililab/qprogram/crosstalk_matrix.py
```diff
@@ -155,8 +155,8 @@ def __str__(self) -> str:
             rows.append(" ".join(row))
         return header + "\n".join(rows)
 
-    @classmethod
-    def from_array(cls, buses: list[str], matrix_array: np.ndarray) -> "CrosstalkMatrix":
+    @staticmethod
```
fedonman, line 158: here too

src/qililab/qprogram/crosstalk_matrix.py
```diff
@@ -179,8 +179,8 @@ def from_array(cls, buses: list[str], matrix_array: np.ndarray) -> "CrosstalkMat
                 instance.matrix[bus1][bus2] = float(matrix_array[i, j])
         return instance
 
-    @classmethod
-    def from_buses(cls, buses: dict[str, dict[str, float]]) -> "CrosstalkMatrix":
+    @staticmethod
```
fedonman, line 182: here too.

src/qililab/qprogram/variable.py
```diff
@@ -30,17 +30,17 @@ class Domain(Enum):
...
 
-    @classmethod
-    def from_yaml(cls, _, node):
+    @staticmethod
+    def from_yaml(_, node):
         """Method to be called automatically during YAML deserialization."""
         _, value = node.value.split("-")
         value = int(value)
-        return cls(value)
+        return Domain(value)
```
fedonman, line 43: These must be class methods by `ruamel.yaml` documentation. Probably because of the same reason as mentioned above: they can serialize/deserialize child classes automatically.

src/qililab/typings/enums.py
```diff
@@ -355,16 +355,16 @@ class Parameter(str, Enum):
...
         return representer.represent_scalar("!Parameter", f"{node.name}-{node.value}")
 
-    @classmethod
-    def from_yaml(cls, _, node):
+    @staticmethod
+    def from_yaml(_, node):
         """Method to be called automatically during YAML deserialization."""
         _, value = node.value.split("-")
-        return cls(value)
+        return Parameter(value)
```
fedonman, line 367: Here too.

fedonman (changes requested):
Added comments for the methods that create instances of the same class.

src/qililab/qprogram/variable.py
GuillermoAbadLopez: Undone!

src/qililab/qprogram/variable.py
GuillermoAbadLopez, line 1: This one is okey, like this right? @fedonman

src/qililab/qprogram/variable.py
```diff
@@ -30,8 +30,8 @@ class Domain(Enum):
     Phase = 3
     Voltage = 4
 
-    @classmethod
-    def to_yaml(cls, representer, node):
+    @staticmethod
```
GuillermoAbadLopez, line 33: This one is okey, like this right? @fedonman

src/qililab/typings/enums.py
```diff
@@ -355,8 +355,8 @@ class Parameter(str, Enum):
     T_PHI = "t_phi"
     GATE_OPTIONS = "options"
 
-    @classmethod
-    def to_yaml(cls, representer, node):
+    @staticmethod
+    def to_yaml(representer, node):
```
GuillermoAbadLopez, line 359: Same here @fedonman, right?

src/qililab/qprogram/variable.py
fedonman: Both `from_yaml` and `to_yaml` need to be classmethods

GuillermoAbadLopez: Cool, changed!

src/qililab/typings/enums.py
GuillermoAbadLopez: Solved!

src/qililab/pulse/pulse_shape/pulse_shape.py
```diff
@@ -51,21 +53,34 @@ def envelope(self, duration: int, amplitude: float, resolution: float) -> np.nda
 
     @classmethod
     def from_dict(cls, dictionary: dict) -> "PulseShape":
```
fedonman, line 55: Try adding the generic type, as I showed you in the example, so we can have strong typing as before. Now, IDE and static analyzers will think that all `load_dict` methods return a `PulseShape`. (instead of the specific child class).

src/qililab/pulse/pulse_distortion/pulse_distortion.py
```diff
@@ -115,25 +118,36 @@ def apply(self, envelope: np.ndarray) -> np.ndarray:
 
     @classmethod
     def from_dict(cls, dictionary: dict) -> "PulseDistortion":
```
fedonman, line 120: Here too. Try adding the generic types to the method signature.

fedonman (commented):
Just one thing more and is good to go!

src/qililab/pulse/pulse_shape/pulse_shape.py
GuillermoAbadLopez: Wow, didn't know you could do that, how cool!

Committed 👌

src/qililab/pulse/pulse_distortion/pulse_distortion.py
GuillermoAbadLopez: Commited!

PR #872, opened 2025-01-15
Qhc 848 support iq modulation for r&s sgs100 a

flavie-lebars:


src/qililab/instruments/rohde_schwarz/sgs100a.py
```diff
@@ -106,6 +116,12 @@ def set_parameter(self, parameter: Parameter, value: ParameterValue, channel_id:
                 else:
                     self.turn_off()
             return
+        if parameter == Parameter.IQ_STATE:
```
fedonman, line 119: This enables hardware modulation, right? If yes, we have a `Parameter.HARDWARE_MODULATION` that we use in other instruments as well.

Let's verify this with measurements team.

src/qililab/instruments/rohde_schwarz/sgs100a.py
```diff
@@ -78,6 +79,15 @@ def rf_on(self):
         """
         return self.settings.rf_on
 
+    @property
+    def IQ_state(self):
```
fedonman, line 83: With the same concept, this could be `modulation` or `hardware_modulation` or similar.

fedonman (commented):
Everything looks good for me! Great work :100: 

Made a suggestion on using an existing parameter.

fedonman (approved):
Forgot to re-approve this. Great work! :100: 

approved

PR #874, opened 2025-01-16
Apply square waveforms optimization by default

fedonman:


jjmartinezQT (approved):
great work, make sure we communicate in aQPU forum so everyone is aware.

PR #880, opened 2025-01-22
New stream processing qm

jordivallsq:
Now the averages can be introduces inside another loop and the results will make sense


tests/qprogram/test_quantum_machines_compiler.py
```diff
@@ -182,7 +182,8 @@ def fixture_measure_operation_with_average() -> QProgram:
     weights = IQPair(I=Square(1.0, duration=200), Q=Square(1.0, duration=200))
     qp = QProgram()
     with qp.average(shots=1000):
-        qp.measure(bus="readout", waveform=drag_wf, weights=weights)
+        with qp.average(shots=1000):
+            qp.measure(bus="readout", waveform=drag_wf, weights=weights)
```
fedonman, line 186: I think, we shouldn't allow this! There should be only one `average` operation per `measure` operation. 

But, even if we don't enforce this rule now, tests should reflect the real case scenarios, e.g. 1 for-loop -> 1 average, 2 for-loops -> 1 average etc. So let's make new tests for these cases.

fedonman: And I would leave this fixture as it is, since it tests a specific case. (average -> measure with no for-loops)

src/qililab/qprogram/quantum_machines_compiler.py
```diff
@@ -69,7 +69,7 @@ def __init__(
         self.stream_Q: qua_dsl._ResultSource = stream_Q
         self.stream_raw_adc: qua_dsl._ResultSource | None = stream_raw_adc
         self.loops_iterations: list[int] = []
-        self.average: bool = False
+        self.average_list: list[bool] = []
```
fedonman, line 72: Not very important, but you could combine the two lists into one, if you make a dataclass that holds information for the loops. For example

```
LoopInfo:
   iterations: int
   is_average: bool
```

This will make the code cleaner in the handle methods below, since you would only need to read/update one list.

fedonman (commented):
Great and super fast implementation! 💯 Left some comments, the most important is about testing.

src/qililab/qprogram/quantum_machines_compiler.py
jordivallsq: Great point should I create a whole dataclass or simply a list of tuples? The dataclass is more elegant but it will only be used in a couple of functions within the same class.

fedonman: Dataclass!

jordivallsq: Done!

tests/qprogram/test_quantum_machines_compiler.py
jordivallsq: Done

src/qililab/qprogram/quantum_machines_compiler.py
```diff
@@ -68,8 +77,7 @@ def __init__(
         self.stream_I: qua_dsl._ResultSource = stream_I
         self.stream_Q: qua_dsl._ResultSource = stream_Q
         self.stream_raw_adc: qua_dsl._ResultSource | None = stream_raw_adc
-        self.loops_iterations: list[int] = []
-        self.average: bool = False
+        self.loop_info: list[LoopInfo] = []
```
fedonman, line 80: Could you rename this to `loops`? Just so the list does not have the same name with the single element of the list which is confusing.

fedonman (approved):
Awesome! 💯 Just a minor naming suggestion. Otherwise, it's good to go!

youre-approved-shirley-auto-group

src/qililab/qprogram/quantum_machines_compiler.py
jordivallsq: Done!

PR #881, opened 2025-01-22
Removing `_list` in `list` variables for DRY principle

GuillermoAbadLopez:


fedonman (approved):
💯 💯 

youre-approved-shirley-auto-group

GuillermoAbadLopez:
@fedonman I added some tests, but the coverage didn't increase, so I can't merge it...
(code-cov is not displaying correctly for `qililab`, so I can't check which lines are missing)

Also I don't think its really necessary in such PR that only changes variable name... Do you have power to force merge it? 🙃

PR #886, opened 2025-02-04
Qhc 855 sequences pulses plot

flavie-lebars:


src/qililab/instruments/rohde_schwarz/sgs100a.py
```diff
@@ -106,6 +116,12 @@ def set_parameter(self, parameter: Parameter, value: ParameterValue, channel_id:
                 else:
                     self.turn_off()
             return
+        if parameter == Parameter.HARDWARE_MODULATION:
+            value = bool(value)
+            self.settings.hardware_modulation = value
+            if self.is_device_active():
+                self.device.IQ_state(value)
```
jordivallsq, line 123: I would remove some lines and make use of the function hardware_modulation that you have created such as the other parameters above.
```suggestion
            self.settings.hardware_modulation = bool(value)
            if self.is_device_active():
                self.device.IQ_state(self.hardware_modulation)
```

src/qililab/instruments/rohde_schwarz/sgs100a.py
```diff
@@ -48,6 +48,7 @@ class SGS100ASettings(Instrument.InstrumentSettings):
         power: float
         frequency: float
         rf_on: bool
+        hardware_modulation: bool
```
jordivallsq, line 51: Is this change related to the plot of pulses? If its not I suggest to create another PR involving this topic, add me as a reviewer as I am already reviewing this with the floowing recomendations

docs/releases/changelog-dev.md
```diff
@@ -13,6 +13,12 @@
 - Raises an error when the inputed value for the QDAC is outside of the bounds provided by QM. Done in 3 ways, runcard, set_parameter RAMPING_ENABLED and set_parameter RAMPING_RATE.
 [#865](https://github.com/qilimanjaro-tech/qililab/pull/865)
 
+- Code to enable the IQ mixer feature already part of qcode in qililab for the rohde and schwarz SGS100a
```
jordivallsq, line 16: Why is the code for the IQ mixer feature inside here if in principle is comming from a different PR?

jordivallsq (commented):
I have some comments regarding the IQ mixer of the R&S. These are not related to this PR but I can aprove it once the changes have been done.

jordivallsq (changes requested):
there are some changes that need to be done as after some tests with the 2Tone the drive and readout did not fully sinchronized as expected

PR #888, opened 2025-02-11
Qhc 850 debug the enabling of the rf on the rhode

flavie-lebars:


jordivallsq (approved):
Great work! Make sure to test this change with hardware and we are good to go

PR #890, opened 2025-02-12
reset qpysequence while Qblox disconnects

jordivallsq:
Fixed small bug while the Qblox didn't returned anything on the second consecutive experiment run because the cache was not restarted but the Q1ASMs disappear while the Qblox is disconnected


jjmartinezQT:
Good catch, please include that one line in the unittests so coverage covers it.

jjmartinezQT (dismissed):
Lacks coverage in code coverage for that line. Once added, good to go on my side.

jordivallsq:
I have added the missing tests

jjmartinezQT (approved):
good job!

PR #893, opened 2025-02-12
More improvements on transpilation

GuillermoAbadLopez:


GuillermoAbadLopez:
IDEA:
Try tracking the gate with an atribute that keeps the order inside the circuit

src/qililab/digital/circuit_router.py
```diff
@@ -158,10 +159,19 @@ def _apply_initial_remap(transpiled_circ: Circuit) -> Circuit:
...
-            new_queue.append(gate)
+            gate_info = deepcopy(gate.raw)
+            gate_info["_control_qubits"] = tuple(wire_names.index(qubit) for qubit in gate_info["_control_qubits"])
+            gate_info["_target_qubits"] = tuple(wire_names.index(qubit) for qubit in gate_info["_target_qubits"])
+            qubits = list(gate_info["_control_qubits"] + gate_info["_target_qubits"])
+            if len(qubits) != len(gate_info["init_args"]):
+                raise ValueError(
+                    f"Gate {gate_info['_class']} has a different number of qubits than the number of init_args, which shouldn't happen with qibo gates."
+                )
+            gate_info["init_args"] = qubits
```
GuillermoAbadLopez, line 172: Here we are also assuming that the `init_args` are the qubits ordered... So that doesn't get solved, although the code is indeed cleaner, and we have an extra check for the `len` of `init_args`, which will make it more robust

PR #896, opened 2025-02-20
allows the user to only put 1 offset, qililab sets the second as 0 an…

flavie-lebars:
…d a warning is given to the user

### DIFF
--- a/src/qililab/qprogram/qblox_compiler.py
+++ b/src/qililab/qprogram/qblox_compiler.py
@@ -420,8 +420,8 @@ class QbloxCompiler:
             else convert(element.offset_path0)
         )
         if element.offset_path1 is None:
-            offset_1 = 0
-            logger.warning("Qblox requires an offset for the two paths, the offset of the second path has been set to 0.")
+            offset_1 = element.offset_path0
+            logger.warning("Qblox requires an offset for the two paths, the offset of the second path has been set to the same as the first path.")
         else:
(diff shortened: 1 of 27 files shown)


src/qililab/qprogram/qblox_compiler.py
```diff
@@ -420,8 +420,10 @@ def _handle_set_offset(self, element: SetOffset):
             else convert(element.offset_path0)
         )
         if element.offset_path1 is None:
-            raise ValueError("No offset has been given for path 1 inside set_offset.")
-        offset_1 = (
+            offset_1 = 0
```
jordivallsq, line 423: I would set `offset_1 = element.offset_path0 ` as it should be more logical for the user. For example in IQ pairs it is more logical for both offsets to be the same rather than one being set "arbitrarily" to zero, after all 0 is actually an arbitrary number from the point of view of the system. All this with the necessary warning of course.

jordivallsq, line 423: after each set offset we need to do an update parameter in case we do a set offset just before a wait and we need this offset changed (usually the case)

flavie-lebars: I will do it in another PR so that I check prior if other parameters, such as the set markers need this fix as well

flavie-lebars: thanks, this has been changed

PR #898, opened 2025-02-27
Qhc 883 add upd param to QBlox latched parameters

flavie-lebars:


fedonman:
Sorry to comment this *after* you have edited all tests. :sweat_smile: I don't think we should add an `upd_param` before **every** `wait` instruction. Instead, we should add one only **if needed**. The algorithm can be pretty simple:
- Store a flag `upd_param_instruction_pending` for each bus, initialized as `False`.
- Whenever a latched command is handled, like `SetFrequency` operation, set the flag to `True`.
- If a real-time operation is handled, like `Play`, set the flag to `False` (since parameters are updated automatically)
- If a `Wait` operation is handled and the flag is `True`, only then add an `upd_param` instruction.

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -125,6 +125,9 @@ def __init__(self) -> None:
         # Delay. Defaults 0 delay and is updated if delays parameter is provided within the runcard.
         self.delay = 0
 
+        # Latched Paramter flag
+        self.upd_param_instruction_pending: bool
```
jordivallsq, line 129: You should set here the default as false, this way you don't need to add line 244

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -388,6 +392,7 @@ def _handle_set_frequency(self, element: SetFrequency):
         self._buses[element.bus].qpy_block_stack[-1].append_component(
             component=QPyInstructions.SetFreq(frequency=frequency)
         )
+        self._buses[element.bus].upd_param_instruction_pending = True
```
jordivallsq, line 395: I thought that set phase, frequency and gain didn't need the update parameter, we should simply delete this from the handlers

jordivallsq (commented):
Good job, I left a couple of comments

src/qililab/qprogram/qblox_compiler.py
flavie-lebars: thanks!

src/qililab/qprogram/qblox_compiler.py
flavie-lebars: @fedonman do you think I should keep the flag only for offset or have it for all latched parameters?

fedonman: All latched parameters since their behaviour is the same: set_mrk, set_freq, reset_ph, set_ph, set_awg_gain and set_awg_offs

<https://docs.qblox.com/en/main/cluster/q1_sequence_processor.html|https://docs.qblox.com/en/main/cluster/q1_sequence_processor.html>

flavie-lebars: perfect thanks!

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -464,14 +473,35 @@ def _handle_add_waits(self, bus: str, duration: int):
...
+                    )
+                elif duration == 4:  # replace the wait with an update parameter of 4 ns
+                    self._buses[bus].qpy_block_stack[-1].append_component(component=QPyInstructions.UpdParam(4))
+
+                elif 4 < duration < 8:  # increase the min wait time of the update parameter and remove the wait
+                    self._buses[bus].qpy_block_stack[-1].append_component(component=QPyInstructions.UpdParam(duration))
+
+                self._buses[bus].upd_param_instruction_pending = False
+            else:
+                self._buses[bus].qpy_block_stack[-1].append_component(component=QPyInstructions.Wait(wait_time=duration % INST_MAX_WAIT))
```
fedonman, line 504: Is there a way to simplify this a bit? Seems a little confusing with the nested-ifs.

tests/qprogram/test_qblox_compiler.py
```diff
@@ -1491,3 +1513,58 @@ def test_calculate_iterations(self, start, stop, step, expected_result):
...
+                            upd_param        4              
+                            play             2, 3, 5        
+                            set_awg_offs     32767, 0       
+                            upd_param        6              
+                            set_mrk          0              
+                            upd_param        4              
+                            stop                            
+        """
+        
+        assert is_q1asm_equal(sequences["drive"], drive_str)
```
fedonman, line 1570: :100: :100:

fedonman (commented):
Looks good to me! I have a small comment on the `_handle_waits`. Other than that, it's good to go.

fedonman (approved):
In any case, I am approving this since the implementation is working fine! If we can address the simplification of the code would be nice, but not mandatory right now. Nice job!

rosycheeks-approved

fedonman (approved):
:100: :100:

PR #899, opened 2025-03-05
Qhc 864 xtalk implementation with experiment class

jordivallsq:
Implemented the software loop implementation of the crosstalk


docs/releases/changelog-dev.md
```diff
@@ -20,6 +20,52 @@
...
+    with experiment.for_loop(variable=flux_z, start=0, stop=0.4, step=0.01):
+        experiment.set_parameter(alias="flux_x1", parameter=ql.Parameter.FLUX, value=flux_x)
+        experiment.set_parameter(alias="flux_z1", parameter=ql.Parameter.FLUX, value=flux_z)
+        experiment.execute_qprogram(qp)
+```
+
+With `Calibration` the code does not need to call for `experiment.crosstalk`:
+
+```
+experiment = ql.Experiment(label="liveplot_test", calibration=calibration)
```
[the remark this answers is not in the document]
fedonman: The warning should be handled during execution too. (inside `platform.set_parameter` or experiment executor's handlers).

Otherwise, I would get a warning every-time I create an Experiment. E.g. I want to create an experiment, and send it through qiboconnection (and use whatever calibration is stored in the system).

Experiment / QProgram and their operations are just placeholders for information, there no logic run up until their execution.

Also, the same warning should be thrown when you do `platform.set_parameter` directly.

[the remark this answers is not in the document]
jordivallsq: Exactly, the Experiment class is just a placeholder that is why I had to create additional lines of code to represent those warnings. BTW those warnings are necessary for measurement as some experiments whould burn the chip for high currents and was a request from measurements.
Inside `platform.set_parameter` there are also the warnings that you mention.
regarning getting a warning every time you create an experiment, this will only happen when you use the FLUX parameter and you are not using the crosstalk class, which is not a good idea and also not common.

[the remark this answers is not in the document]
fedonman: So you agree that everything must be moved to the `execution` of the experiment, aka `ExperimentExecutor`. :smiley: 

Imagine creating a portfolio of experiments, as we have in `qilitools` for QProgram. If we enforce to pass a calibration file, this is not possible. `Calibration`, `Bus Mapping` and anything that has to do with the actual execution, should be passed and checked during execution.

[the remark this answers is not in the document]
fedonman: One should be able to reuse the same experiment and do `platform.execute_experiment(flux_vs_flux, calibration=calibrationA)` and then ``platform.execute_experiment(flux_vs_flux, calibration=calibrationB)`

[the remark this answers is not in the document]
fedonman: If the experiment needs a calibration and the user doesn't provide it, or provides a "bad" one, of course we need the checks to disallow execution. But this will be done inside `platform.execute_experiment` -> `ExperimentExecutor`, not when the user creates the "description" of the experiment.

src/qililab/qprogram/operations/set_parameter.py
```diff
@@ -21,10 +22,18 @@
...
 class SetParameter(Operation):
     def __init__(
-        self, alias: str, parameter: Parameter, value: int | float | bool, channel_id: int | None = None
+        self,
+        alias: str,
+        parameter: Parameter,
+        value: int | float | bool,
+        channel_id: int | None = None,
+        crosstalk: CrosstalkMatrix | None = None,
+        flux_list: list[str] | None = None,
```
fedonman, line 31: Why are these added here? They are not used anywhere.

```suggestion
```

src/qililab/qprogram/experiment.py
```diff
@@ -90,15 +105,33 @@ def set_parameter(self, alias: str, parameter: Parameter, value: int | float | i
...
+            if alias not in self.crosstalk_data.matrix.keys():  # type: ignore[union-attr]
+                for key in list(self.crosstalk_data.matrix.keys()):  # type: ignore[union-attr]
+                    self.crosstalk_data[alias] = {key: 0.0}  # type: ignore[index]
+                    self.crosstalk_data[key] = {alias: 0.0}  # type: ignore[index]
+                self.crosstalk_data[alias] = {alias: 1.0}  # type: ignore[index]
+                warnings.warn(
+                    f"{alias} not inside crosstalk matrix, adding it with identity values\n{self.crosstalk_data}"
+                )
+            if isinstance(value, Variable):
+                self.value_flux_list[value.label] = alias
```
fedonman, line 124: This is exactly the same code with `Platform._process_crosstalk`. We should remove it from here and keep only the code in `Platform`, which is executed in runtime.

src/qililab/qprogram/structured_program.py
```diff
@@ -99,6 +102,15 @@ def variables(self) -> list[Variable]:
         """
         return list(self._variables)
 
+    @property
+    def get_crosstalk(self) -> CrosstalkMatrix | None:
```
fedonman, line 106: If it's a property then `crosstalk` is a better name. `get_crosstalk` implies a method.
```suggestion
    def crosstalk(self) -> CrosstalkMatrix | None:
```

src/qililab/qprogram/structured_program.py
```diff
@@ -250,10 +262,22 @@ def _float_variable(label: str, domain: Domain = Domain.Scalar) -> FloatVariable
 
         if domain == Domain.Time:
             return _int_variable(label, domain)
-        if domain in [Domain.Frequency, Domain.Phase, Domain.Voltage]:
+        if domain in [Domain.Frequency, Domain.Phase, Domain.Voltage, Domain.Flux]:
             return _float_variable(label, domain)
         raise NotImplementedError
 
+    def crosstalk(self, crosstalk: CrosstalkMatrix | None = None, calibration: Calibration | None = None):
```
fedonman, line 269: And this is better named `set_crosstalk`.

```suggestion
    def set_crosstalk(self, crosstalk: CrosstalkMatrix | None = None, calibration: Calibration | None = None):
```

src/qililab/qprogram/structured_program.py
```diff
@@ -250,10 +262,22 @@ def _float_variable(label: str, domain: Domain = Domain.Scalar) -> FloatVariable
...
+        """Declare Crosstalk matrix and offsets
+
+        Args:
+            crosstalk (CrosstalkMatrix): Crosstalk class including all necessary information
+        """
+        if calibration:
+            crosstalk = calibration.crosstalk_matrix
+        if not crosstalk:
+            raise ValueError("crosstalk must have either Calibration with crosstalk or CrosstalkMatrix")
+        self._crosstalk = crosstalk
```
fedonman, line 279: Why do we allow `Calibration` as a parameter in a method that sets the crosstalk? I think it is clearer to allow only `CrosstalkMatrix` and if the user has a `Calibration` instead just access it by `calibration.crosstalk_matrix`

src/qililab/qprogram/structured_program.py
```diff
@@ -46,6 +48,7 @@ def __init__(self) -> None:
         self._block_stack: deque[Block] = deque([self._body])
         self._variables: dict[Variable, VariableInfo] = {}
         self._buses: set[str] = set()
+        self._crosstalk: CrosstalkMatrix | None = None
```
fedonman, line 51: Here you have an attribute `_crosstalk: CrosstalkMatrix` and in Experiment which inherits from StructuredProgram you have an attribute `crosstalk_data: CrosstalkMatrix`. Why do we need both? If it is duplicate we should remove one of the two.

src/qililab/platform/platform.py
```diff
@@ -495,9 +512,111 @@ def set_parameter(
...
+        Args:
+            crosstalk (CrosstalkMatrix | None): _description_
+            calibration (Calibration | None): _description_
+        """
+        if isinstance(crosstalk, CrosstalkMatrix):
+            self.crosstalk = crosstalk
+        elif isinstance(crosstalk, Calibration):
+            self.crosstalk = crosstalk.crosstalk_matrix
+        else:
+            raise ValueError("add_crosstalk must have either Calibration with crosstalk or CrosstalkMatrix")
```
fedonman, line 603: As with the previous comment, it is clearer to rename this to `set_crosstalk` for consistency, and remove the `Calibration` as parameter.

fedonman (changes requested):
Great work Jordi! I have several comments, not on the implementation per se, but rather on simplifying the code structure and not duplciating things.

Also, I have an idea that can eliminate the need for keeping a CrosstalkMatrix inside Experiment altogether: Create a new Operation, `SetCrosstalkOperation` and use that when calling `experiment.set_crosstalk()`. Then in the ExperimentExecutor call the `platform.set_crosstalk()` in the same way we call the `platform.set_parameter` from a `SetParameterOperation`. This way all the logic happens during execution. We can talk about it.

fedonman:
> Great work Jordi! I have several comments, not on the implementation per se, but rather on simplifying the code structure and not duplciating things.
> 
> Also, I have an idea that can eliminate the need for keeping a CrosstalkMatrix inside Experiment altogether: Create a new Operation, `SetCrosstalkOperation` and use that when calling `experiment.set_crosstalk()`. Then in the ExperimentExecutor call the `platform.set_crosstalk()` in the same way we call the `platform.set_parameter` from a `SetParameterOperation`. This way all the logic happens during execution. We can talk about it.

Continuing from this. We should define a new operation that means set the crosstalk matrix for the experiment.

```
@yaml.register_class
class SetCrosstalk(Operation):
    def __init__(
        self,
        crosstalk: CrosstalkMatrix
    ) -> None:
        super().__init__()
        self.crosstalk: CrosstalkMatrix crosstalk
```

and then keep your `experiment.set_crosstalk()` method, that will instead add the operation to the experiment body, like all methods do.

```
def set_crosstalk(crosstalk: CrosstalkMatrix):
   operation = SetCrosstalk(crosstalk=crosstalk)
   self._active_block.append(operation)
```

Then, within `ExperimentExecutor` if a `SetCrosstalk` is encountered, create the proper lambda function that will call the `platform.set_crosstalk()`:

```
if isinstance(element, SetCrosstalk):
     elements_operations.append(
            lambda operation=element: self.platform.set_crosstalk(crosstalk=crosstalk)
      )
```

This way you treat `crosstalk` as any other element of Experiment. The `set_crosstalk()` method becomes optional (you can set it or you may not) and there is no need to store any information in `StructuredProgram` or `Experiment`. Everything is handled by `ExperimentExecutor` and the current state of the platform.

src/qililab/qprogram/structured_program.py
jordivallsq: I removed the other one

src/qililab/platform/platform.py
jordivallsq: I removed the calibration and changed the name

src/qililab/qprogram/experiment.py
jordivallsq: I removed this code, this was nnecessary when we applied the warniings in the Experiment class bit now everything is easier

docs/releases/changelog-dev.md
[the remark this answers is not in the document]
jordivallsq: I changed this part into your suggestion

src/qililab/qprogram/experiment.py
```diff
@@ -90,15 +98,15 @@ def set_parameter(self, alias: str, parameter: Parameter, value: int | float | i
             parameter (Parameter): The parameter to set.
             value (int | float): The value to set for the parameter.
         """
-        operation = SetParameter(alias=alias, parameter=parameter, value=value, channel_id=channel_id)
+        operation = SetParameter(alias=alias, parameter=parameter, value=value, channel_id=channel_id)  # type: ignore[arg-type]
```
fedonman, line 101: Shouldn't need `#type: ignore` here.

src/qililab/qprogram/structured_program.py
```diff
@@ -99,6 +101,15 @@ def variables(self) -> list[Variable]:
...
 
+    @property
+    def crosstalk(self) -> CrosstalkMatrix | None:
+        """Get the crosstalk.
+
+        Returns:
+            CrosstalkMatrix | None: Crosstalk matrix and metadata
+        """
+        return self._crosstalk
+
```
fedonman, line 112: This will also not be needed.

src/qililab/qprogram/structured_program.py
```diff
@@ -46,6 +47,7 @@ def __init__(self) -> None:
         self._block_stack: deque[Block] = deque([self._body])
         self._variables: dict[Variable, VariableInfo] = {}
         self._buses: set[str] = set()
+        self._crosstalk: CrosstalkMatrix | None = None
```
fedonman, line 50: This will also not be needed.

src/qililab/qprogram/experiment_executor.py
```diff
@@ -104,10 +106,16 @@ class ExperimentExecutor:
...
-    def __init__(self, platform: "Platform", experiment: Experiment):
+    def __init__(self, platform: "Platform", experiment: Experiment, calibration: Calibration | None):
         self.platform = platform
         self.experiment = experiment
 
+        self.crosstalk: CrosstalkMatrix | None
+        if calibration and calibration.crosstalk_matrix:
+            self.crosstalk = calibration.crosstalk_matrix
+        else:
+            self.crosstalk = self.experiment.crosstalk
```
fedonman, line 117: This will become simpler.

src/qililab/platform/platform.py
```diff
@@ -330,6 +332,15 @@ def __init__(self, runcard: Runcard):
         self.experiment_results_path_format: str = "{date}/{time}/{label}.h5"
         """Format of the experiment results path."""
 
+        self.crosstalk: CrosstalkMatrix | None = None
```
fedonman, line 335: This is the source of truth! No need to re-store crosstalk in `structured_program` or somewhere else.

If users want to assign a new crosstalke **during** executing, they will use the new `SetCrosstalk` operation.

src/qililab/qprogram/experiment_executor.py
fedonman, line 117: Platform already contains a `crosstalk`. No need to re-assign to an ExperimentExecutor's attribute.

src/qililab/qprogram/experiment_executor.py
```diff
@@ -338,17 +351,20 @@ def process_elements(elements: list[Block | Operation]) -> list[Callable]:
...
-                                lambda operation=element,
-                                value=current_value_of_variable[element.value.uuid]: self.platform.set_parameter(
+                                lambda operation=element, value=current_value_of_variable[
+                                    element.value.uuid
+                                ]: self.platform.set_parameter(
                                     alias=operation.alias,
                                     parameter=operation.parameter,
                                     value=value,
                                     channel_id=operation.channel_id,
+                                    crosstalk=self.crosstalk,
```
fedonman, line 367: It will not be needed to pass `crosstalk`. Platform knows it.

src/qililab/qprogram/experiment_executor.py
```diff
@@ -359,6 +375,7 @@ def process_elements(elements: list[Block | Operation]) -> list[Callable]:
                                 parameter=operation.parameter,
                                 value=operation.value,
                                 channel_id=operation.channel_id,
+                                crosstalk=self.crosstalk,
```
fedonman, line 378: It will not be needed to pass `crosstalk`. Platform knows it.

src/qililab/result/experiment_results.py
```diff
@@ -157,6 +158,15 @@ def platform(self) -> str:
...
         return self._file[ExperimentResults.PLATFORM_PATH][()].decode("utf-8")
 
+    @property
+    def crosstalk(self) -> str:
+        """Gets the YAML representation of the platform.
+
+        Returns:
+            str: The YAML string of the platform.
+        """
+        return self._file[ExperimentResults.CROSSTALK][()].decode("utf-8")
```
fedonman, line 168: Probably, won't need this, since `crosstalk` a) should be contained in the platform serialization and b) can be changed multiple times during execution (this will be serialized in the `experiment`).

fedonman (changes requested):
I am requesting changes to re-trigger pullpo, and do a sum-up. If you follow the workshop on the `ExperimentsExecutor`, and previous reviewing on the new `SetCrosstalk` operation, the details of the rest of the implementation should become apparent. I commented some of them for clarity.

> Continuing from this. We should define a new operation that means set the crosstalk matrix for the experiment.
> 
> ```
> @yaml.register_class
> class SetCrosstalk(Operation):
>     def __init__(
>         self,
>         crosstalk: CrosstalkMatrix
>     ) -> None:
>         super().__init__()
>         self.crosstalk: CrosstalkMatrix = crosstalk
> ```
> 
> and then keep your `experiment.set_crosstalk()` method, that will instead add the operation to the experiment body, like all methods do.
> 
> ```
> def set_crosstalk(crosstalk: CrosstalkMatrix):
>    operation = SetCrosstalk(crosstalk=crosstalk)
>    self._active_block.append(operation)
> ```
> 
> Then, within `ExperimentExecutor` if a `SetCrosstalk` is encountered, create the proper lambda function that will call the `platform.set_crosstalk()`:
> 
> ```
> if isinstance(element, SetCrosstalk):
>      elements_operations.append(
>             lambda operation=element: self.platform.set_crosstalk(crosstalk=crosstalk)
>       )
> ```
> 
> This way you treat `crosstalk` as any other element of Experiment. The `set_crosstalk()` method becomes optional (you can set it or you may not) and there is no need to store any information in `StructuredProgram` or `Experiment`. Everything is handled by `ExperimentExecutor` and the current state of the platform.

src/qililab/qprogram/experiment.py
jordivallsq: This was because I added Variable as a possible type of value forgetting bool (the previous set parameter the possibe types of value were wrong (line 83 of the old code).

src/qililab/qprogram/structured_program.py
jordivallsq: Also removed set_crosstalk from here and defined it in Experiment class

src/qililab/qprogram/structured_program.py
jordivallsq: everything removed from the structured program

src/qililab/qprogram/experiment_executor.py
jordivallsq: changed `self.crosstalk` by `self.platform.crosstalk`

src/qililab/qprogram/experiment_executor.py
jordivallsq: removed on the set_parameters

src/qililab/result/experiment_results.py
jordivallsq: Got rid of saving the crosstalk as it is present in platform and the experiment as a set

jordivallsq:
All requested changes done

src/qililab/qprogram/experiment_executor.py
```diff
@@ -104,10 +106,13 @@ class ExperimentExecutor:
...
     """
 
-    def __init__(self, platform: "Platform", experiment: Experiment):
+    def __init__(self, platform: "Platform", experiment: Experiment, calibration: Calibration | None):
         self.platform = platform
         self.experiment = experiment
 
+        if calibration and calibration.crosstalk_matrix:
+            self.platform.crosstalk = calibration.crosstalk_matrix
+
```
fedonman, line 115: Last thing, let's remove this for now, since platform already has the crosstalk_matrix, and if users want to change now have the `experiment.set_crosstalk()` method.

In general, i want us to have the `Calibration` passed to the `execute()` method, but we have to think how to propagate it to  the whole experiment, like the `execute_qprograms` and so on, so let's do it in another PR after some discussion.

docs/releases/changelog-dev.md
```diff
@@ -2,8 +2,82 @@
...
+  `qprogram.draw(self, averages_displayed=False)`
+  ```python
+  qp = QProgram()
+  qprogram.draw()
+  ```
+    
+Both methods compile the qprogram internally to generate the sequencer and call `QbloxDraw.draw(self, sequencer, runcard_data=None, averages_displayed=False) -> dict`.
+Aditionally both 
+
+  [#901](https://github.com/qilimanjaro-tech/qililab/pull/901)
```
fedonman, line 74: Should this be here? It's from Flavie's PR.

docs/releases/changelog-dev.md
```diff
@@ -2,8 +2,82 @@
...
+experiment.crosstalk(calibration=calibration)
+with experiment.for_loop(variable=flux_x, start=0, stop=0.4, step=0.01):
+    with experiment.for_loop(variable=flux_z, start=0, stop=0.4, step=0.01):
+        experiment.set_parameter(alias="flux_x1", parameter=ql.Parameter.FLUX, value=flux_x)
+        experiment.set_parameter(alias="flux_z1", parameter=ql.Parameter.FLUX, value=flux_z)
+        experiment.execute_qprogram(qp)
+```
+
+Note that not giving a crosstalk matrix implies working with the identity. Warnings rise while creating the experiment to inform the user of this.
+[#899](https://github.com/qilimanjaro-tech/qililab/pull/899)
```
fedonman, line 38: Take a look in the code example and make sure it's updated to the latest changes, e.g. `set_crosstalk` instead of `add_crosstalk` and so on.

fedonman (commented):
Three small comments, and is good to go! :rocket:

fedonman (approved):
Awesome work Jordi! :100: Thank you very much for the patience doing all the requested changes! :pray: :rocket: 

200w (1)

PR #901, opened 2025-03-10
[QHC-855] Qblox Draw

flavie-lebars:
QBlox: An oscilloscope simulator has been implemented. It takes the sequencer as input, plots its waveforms and returns a dictionary (data_draw) containing all data points used for plotting.

The user can access the Qblox drawing feature in two ways:
  1. Via platform (includes runcard knowledge)
  `platform.draw(self, qprogram: QProgram, averages_displayed: bool = False)`
  ```python
  with platform.session():
    platform.draw(qprogram = qprogram)
  ```
  Note that if it is used with a Quantum Machine runcard, a ValueError will be generated.

  2. Via QProgram (includes runcard knowledge)
  `qprogram.draw(self, averages_displayed=False)`
  ```python
  qp = QProgram()
  qprogram.draw()
  ```
    
Both methods compile the qprogram internally to generate the sequencer and call `QbloxDraw.draw(self, sequencer, runcard_data=None, averages_displayed=False) -> dict`.


src/qililab/platform/platform.py
```diff
@@ -315,6 +316,9 @@ def __init__(self, runcard: Runcard):
         self.digital_compilation_settings = runcard.digital
         """Gate settings and definitions (``dataclass``). These setting contain how to decompose gates into pulses."""
 
+        self.alias = [a["alias"] for a in map(asdict, runcard.buses)]
```
jjmartinezQT, line 319: Let's move this one into the data_draw_oscilloscope method if it is only use there.

src/qililab/platform/platform.py
```diff
@@ -1367,3 +1427,18 @@ def compile(
         )
 
         return compiled_programs, final_layout
+
+    def draw_oscilloscope_platform(self, qprogram: QProgram, averages_displayed: bool = False):
+        """Draw the QProgram using QBlox Compiler
```
jjmartinezQT, line 1432: Pulse plotting only works for Qblox for now. what happens if the user sends a QuantumMachines QProgram here?

jjmartinezQT (dismissed):

src/qililab/platform/platform.py
flavie-lebars: done and added self.runcard in the init as discussed

src/qililab/platform/platform.py
flavie-lebars: done, now if a quantum machine runcard is provided it raises an error instead of a warning

src/qililab/platform/platform.py
```diff
@@ -465,6 +469,66 @@ def get_parameter(self, alias: str, parameter: Parameter, channel_id: ChannelID
         element = self.get_element(alias=alias)
         return element.get_parameter(parameter=parameter, channel_id=channel_id)
 
+    def data_draw_oscilloscope(self):
+        """From the runcard retrieve the parameters necessary to draw the qprogram."""
+        data_osci = {}
+        # All the aliases of the platform in a list
+        alias = [a["alias"] for a in map(asdict, self.runcard.buses)]
```
fedonman, line 476: `[bus.alias for bus in self.runcard.buses]` should work here.

src/qililab/platform/platform.py
```diff
@@ -465,6 +469,66 @@ def get_parameter(self, alias: str, parameter: Parameter, channel_id: ChannelID
...
+        buses = [self.buses.get(alias=a) for a in alias]
```
fedonman, line 477: Also, what exactly are we doing here? If you want a list of all the buses, you can do directly `buses = [bus for bus in self.platform.buses]`

src/qililab/instruments/qblox/qblox_draw.py
```diff
@@ -0,0 +1,534 @@
...
+# limitations under the License.
+
+import re
+
+import numpy as np
+import plotly.graph_objects as go
+from plotly.subplots import make_subplots
+
+
+class QbloxDraw:
```
fedonman, line 22: I can see some refactorings to make the code simpler, but since it works and is a long file, let's merge it as it is, and we can do it in the future. Nice work!!

src/qililab/platform/platform.py
```diff
@@ -299,6 +300,9 @@ def __init__(self, runcard: Runcard):
         self.name = runcard.name
         """Name of the platform (``str``) """
 
+        self.runcard = runcard
```
fedonman, line 303: Why do we need to store the runcard here?

src/qililab/platform/platform.py
```diff
@@ -465,6 +469,66 @@ def get_parameter(self, alias: str, parameter: Parameter, channel_id: ChannelID
...
+
+        data_oscillocope = {}
+        for bus, bus_param in data_osci.items():
+            data_oscillocope[bus] = {}
+            for param, value in bus_param.items():
+                try:
+                    data_oscillocope[bus][param.value] = value
+
+                except AttributeError:
+                    data_oscillocope[bus][param] = value
```
fedonman, line 528: Why do we copy the data from `data_osci` to a new dictionary? We could return the first dictionary directly.

src/qililab/platform/platform.py
```diff
@@ -465,6 +469,66 @@ def get_parameter(self, alias: str, parameter: Parameter, channel_id: ChannelID
...
+                            Parameter.IF,
+                            Parameter.GAIN_I,
+                            Parameter.GAIN_Q,
+                            Parameter.OFFSET_I,
+                            Parameter.OFFSET_Q,
+                            Parameter.HARDWARE_MODULATION,
+                        ]
+                        for p in param:
+                            val = self.get_parameter(bus.alias, p)
+                            data_osci[bus.alias][p] = val
```
fedonman, line 498: if you need the "string" representation of the parameter you can do `data_osci[bus.alias][p.value] = val` here. Then you won't need this copying loop at the end.

src/qililab/platform/platform.py
```diff
@@ -465,6 +469,66 @@ def get_parameter(self, alias: str, parameter: Parameter, channel_id: ChannelID
...
+                        static_offset_q = 0
+                        param = [
+                            Parameter.IF,
+                            Parameter.GAIN_I,
+                            Parameter.GAIN_Q,
+                            Parameter.OFFSET_I,
+                            Parameter.OFFSET_Q,
+                            Parameter.HARDWARE_MODULATION,
+                        ]
+                        for p in param:
```
fedonman, line 496: Please avoid variable abbreviations 🙏 We should use full names, so it is more readable. The list should be `parameters` and the loop `for parameter in parameters:`.

src/qililab/platform/platform.py
```diff
@@ -465,6 +469,66 @@ def get_parameter(self, alias: str, parameter: Parameter, channel_id: ChannelID
...
+                            val = self.get_parameter(bus.alias, p)
+                            data_osci[bus.alias][p] = val
+
+                        data_osci[bus.alias]["instrument_name"] = instrument.name.value
+
+                        if instrument.name == InstrumentName.QBLOX_QCM:
```
fedonman, line 502: Do we need the`out_offsets` of QCM only? What about the other 3 modules?

src/qililab/platform/platform.py
```diff
@@ -1367,3 +1431,17 @@ def compile(
         )
 
         return compiled_programs, final_layout
+
+    def draw_oscilloscope_platform(self, qprogram: QProgram, averages_displayed: bool = False):
```
fedonman, line 1435: We could rename this to `draw_oscilloscope`.

src/qililab/platform/platform.py
```diff
@@ -465,6 +469,66 @@ def get_parameter(self, alias: str, parameter: Parameter, channel_id: ChannelID
         element = self.get_element(alias=alias)
         return element.get_parameter(parameter=parameter, channel_id=channel_id)
 
+    def data_draw_oscilloscope(self):
```
fedonman, line 472: Should this method be public or private?

fedonman (changes requested):
Awesome work Flavie!! 💯  I made some comments on the `platform` part of the implementation.

docs/releases/changelog-dev.md
```diff
@@ -2,6 +2,10 @@
 
 ### New features since last release
 
+- An oscilloscope simulator has been implement for QBlox. It takes the sequencer as an input and plots it as if using an osilloscope.
```
DZPM, line 5: Typo, "osilloscope" -> oscilloscope

src/qililab/instruments/qblox/qblox_draw.py
```diff
@@ -0,0 +1,534 @@
...
+            gain = param.get("gain_i", 1)
+        elif i_or_q == "Q":
+            scaling_factor, max_voltage = self._get_scaling_factors(
+                param, param.get("offset_q", 0), param.get("static_offset_q", 0)
+            )
+            gain = param.get("gain_q", 1)
+        return scaling_factor, max_voltage, gain
+
+    def _get_scaling_factors(self, param, dynamic_off, static_off):
+        if param["hardware_modulation"]:
```
DZPM, line 69: This could return a `KeyError`, if `"hardware_modulation"` is not set.

Can we guarantee that `"hardware_modulation"` will always be defined?
Same for `param["max_voltage"]`

src/qililab/instruments/qblox/qblox_draw.py
```diff
@@ -0,0 +1,534 @@
...
+
+            # parse all lines of qprogram and deal with special case of reset phase
+            for line in program_line:
+                stripped_line = line.strip()
+                if stripped_line.startswith("reset_ph"):
+                    processed_lines.append("reset_ph             2")
+                else:
+                    processed_lines.append(stripped_line)
+            line_data = "\n".join(processed_lines)
+            pattern = r"(\w+):|(\w+)(?:\s+([^\n]*))?"
```
DZPM, line 276: I'm bad with regexp, would be nice a comment explaining what are we parsing here, maybe an example of a typical line

src/qililab/instruments/qblox/qblox_draw.py
```diff
@@ -0,0 +1,534 @@
...
+
+                elif Q1ASM_line[-1] not in instructions_ran:  # run if no loop label
+                    self._call_handlers(Q1ASM_line, param, register, data_draw[bus], wf)
+
+                else:
+                    pass
+
+            # remove the latest freq, phase and offset data points if nothing played after
+            for key in ["intermediate_frequency", "phase", "offset_i", "offset_q"]:
+                if param[f"{key}_new"]:
```
DZPM, line 444: Again, couldn't this raise a `KeyError`?

src/qililab/qprogram/qprogram.py
```diff
@@ -716,3 +717,21 @@ def measure(
...
+        """
+
+        from qililab.instruments.qblox.qblox_draw import QbloxDraw
+        from qililab.qprogram.qblox_compiler import QbloxCompiler
+
+        draw = QbloxDraw()
+        compiler = QbloxCompiler()
+        results = compiler.compile(self)
+        result_draw = draw.draw_oscilloscope(result=results, averages_displayed=averages_displayed)
+        logger.warning("The drawing feature is currently only supported for QBlox.")
```
DZPM, line 736: If it's unsupported, why a `logger.warning` instead of a `logger.error`?

Some people will ignore loggers, so maybe even a custom `UnsupportedOperationException`?

DZPM:
Good PR, and the tests are clear. Just added some questions and suggestions.

src/qililab/instruments/qblox/qblox_draw.py
flavie-lebars: thanks for all advices David! for this, is it necessary to raise the key error? If there is no runcard, the draw_oscilloscope automatically sets the hardware modulation as true (line 353)

src/qililab/instruments/qblox/qblox_draw.py
flavie-lebars: same as the other one, I alaways set them at the beginning of draw_oscilloscope but if you believe it should still raise a key error I can do it

src/qililab/qprogram/qprogram.py
flavie-lebars: If the user plots from qprogram we don't know the machine being used, hence I can't raise an error. On the other hand for the platform, an error is raised if quantum machine is being used.

src/qililab/platform/platform.py
fedonman: Or we could name this `draw` globally. (I mean we don't have any other draw method to differentiate from)

src/qililab/platform/platform.py
flavie-lebars: yes much better thanks😅

src/qililab/platform/platform.py
flavie-lebars: i have added for the QRM baseband, but I spoked with Adri and it doesn't seem that they have the DAC offset for RF modules

jordivallsq: For QCM-RF/QRM-RF we have `out0_offset_path0`, `out0_offset_path1` and `out1_offset_path0`, `out1_offset_path1` for QCM-RF only. This should attribute to the dac offset (https://docs.qblox.com/en/main/cluster/qcm_rf.html#block-diagram) which is applied just before the mixing as seen in the diagram

jordivallsq (commented):
Great job!!! I left a couple of comments, I would have left more comments to refactor the `qblox_draw.py` but I agree with @fedonman comment as it works for now and we can focus on on simplifying some parts in the future

src/qililab/platform/platform.py
flavie-lebars: i have now added it for the rf modules as well, thanks!

docs/releases/changelog-dev.md
```diff
@@ -2,6 +2,10 @@
 
 ### New features since last release
 
+- An oscilloscope simulator has been implement for QBlox. It takes the sequencer as an input and plots it as if using an oscilloscope.
+  Qblox Draw can either be called through the platform (adding the knowledge of the runcard) or qprogram directly.
+  [#899](https://github.com/qilimanjaro-tech/qililab/pull/899)
```
fedonman, line 7: Let's add a code example on how to use the new feature. We should always try to add code examples in the CHANGELOG, and this is especially important for new features.

src/qililab/qprogram/qprogram.py
```diff
@@ -716,3 +717,21 @@ def measure(
...
+
+        from qililab.instruments.qblox.qblox_draw import QbloxDraw
+        from qililab.qprogram.qblox_compiler import QbloxCompiler
+
+        qblox_draw = QbloxDraw()
+        compiler = QbloxCompiler()
+        results = compiler.compile(self)
+        result_draw = qblox_draw.draw(result=results, averages_displayed=averages_displayed)
+        logger.warning("The drawing feature is currently only supported for QBlox.")
+        return result_draw
```
fedonman, line 737: What are we returning here? If we need to return something add it to the method's signature and docstring. If we don't want to return, just remove this line.

fedonman (changes requested):
Two more comments.

docs/releases/changelog-dev.md
flavie-lebars: i have detailed the change log now, let me know if you would like more details!

DZPM (approved):
Looks good to me :+1:

fedonman (approved):
Great work! :100: :partying_face: 

200w

PR #903, opened 2025-03-17
Qhc 901 qblox first iteration hardware loop issues

flavie-lebars:
This is a temporary fix, awaiting for qblox to fix the bug. The first frequency and gain of a hardware loop is wrong. This does not affect the averages
This PR sets the frequency and the gain twice in a row whenever they are set to bypass this bug


fedonman (commented):
Awesome! Let's add a Changelog entry stating the temporary bug fix and is good to go 🚀

fedonman (approved):
💯 💯 

200w (2)

PR #906, opened 2025-03-17
[QHC-918] Implement Parallel Execution of QPrograms in Qblox

fedonman:
QBlox: Added support for executing multiple QProgram instances in parallel via the new method `platform.execute_qprograms_parallel(qprograms: list[QProgram])`. This method returns a list of `QProgramResults` corresponding to the input order of the provided qprograms. Note that an error will be raised if any submitted qprograms share a common bus.

```python
with platform.session():
  results = platform.execute_qprograms_parallel([qprogram1, qprogram2, qprogram3])
```

### DIFF
--- a/tests/platform/test_platform.py
+++ b/tests/platform/test_platform.py
@@ -1090,3 +1090,73 @@ class TestMethods:
                 for flux_bus in platform.analog_compilation_settings.flux_control_topology
                 if flux_bus.flux == flux
             )
+    
+    
+    def  test_parallelisation_same_bus_raises_error(self, platform: Platform):
+        """Test that if parallelisation is attempted on qprograms using at least one bus in common, an error will be raised"""
+        error_string = "QPrograms cannot be executed in parallel."
+        qp1 = QProgram()
+        qp2 = QProgram()
+        qp3 = QProgram()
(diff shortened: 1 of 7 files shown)


tests/platform/test_platform.py
```diff
@@ -1090,3 +1090,92 @@ def test_get_element_flux(self, platform: Platform):
...
+            patch.object(Bus, "acquire_qprogram_results") as acquire_qprogram_results,
+            patch.object(QbloxModule, "sync_sequencer") as sync_sequencer,
+            patch.object(QbloxModule, "desync_sequencer") as desync_sequencer,
+        ):
+            qp_list = [qp1,qp2,qp3]
+            with pytest.raises(ValueError, match=error_string):
+                platform.execute_qprograms_parallel(qp_list, debug=True)
+
+
+    def test_parallelisation_execute_quantum_machine(self, platform_quantum_machines: Platform):
```
jjmartinezQT, line 1119: This is just a picky detail, but I suggest to change the name of this unit test to something that reflects we do not support this feature in QM.

jjmartinezQT (changes requested):
Great work, and fast, small change request then good to go.

jjmartinezQT (approved):
Amazing, great work.

PR #907, opened 2025-03-21
[QHC-924] Remove GRES check in `slurm.py`

GuillermoAbadLopez:
@JavierSab and @islegmar pointed out that not everyone using `slurm` will have GRES, concretely BSC won't, so this would error...

With this change we will need to release version `0.29.1`, to send BSC, correcting this bug

### DIFF
--- a/tests/test_slurm.py
+++ b/tests/test_slurm.py
@@ -52,16 +52,16 @@ class TestSubmitJob:
             )
 
     def test_submit_job_no_gres_provided(self, ip):
-        """Check ValueError is raised in case GRES is not provided."""
+        """Check no Error is raised in case GRES is not provided."""
         ip.run_cell(raw_cell="a=1\nb=1")
-        with pytest.raises(
-            ValueError, match="GRES needs to be provided! See the available ones typing 'sinfo -o '%G'' in the terminal"
-        ):
-            ip.run_cell_magic(
-                magic_name="submit_job",
-                line=f"-o results -l {slurm_job_data_test} -n unit_test -e local",
-                cell="a+b",
-            )
+
+        ip.run_cell_magic(
+            magic_name="submit_job",
+            line=f"-o results -l {slurm_job_data_test} -n unit_test -e local",
+            cell="results = a+b ",
+        )
+        time.sleep(4)
+        assert ip.user_global_ns["results"].result() == 2
 
     def test_submit_job_with_random_file_in_logs_folder(self, ip):
         """Check non-submitit files are deleted if found in logs folder"""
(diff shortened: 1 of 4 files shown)


GuillermoAbadLopez:
> Do we need a slurm file at all in Qililab?

Good question, if its there I guess the answer is yes, but I have no idea, @JavierSab and @islegmar, will know better to answer that

jjmartinezQT:
> > Do we need a slurm file at all in Qililab?
> 
> Good question, if Adri added it there I guess the answer is yes, but I have no idea, @JavierSab and @islegmar, will know better to answer that 🤔

I will wait for the assesment there, but Qililab should hold no dependency on slurm execution at all. If the reason is because the calibration still lives within Qililab we can approve this for the time being, but Calibration will go out to the calibration repo. This dependency is not good and enforcing a purpose Qililab is not meant for.

jjmartinezQT:
> > > Do we need a slurm file at all in Qililab?
> > 
> > 
> > Good question, if Adri added it there I guess the answer is yes, but I have no idea, @JavierSab and @islegmar, will know better to answer that 🤔
> 
> I will wait for their assesment then, but Qililab should hold no dependency on slurm execution at all. If the reason is because the calibration still lives within Qililab we can approve this for the time being, but Calibration will go out to the calibration repo. This dependency is not good and enforcing a purpose Qililab is not meant for.

After the meeting with access today this change is ok to go ahead, please @GuillermoAbadLopez get anyone from QHC to help with a proper review as I won't probably have much time to do it myself. Thanks.

tests/test_slurm.py
```diff
@@ -52,16 +52,16 @@ def test_submit_job_output_not_assigned(self, ip):
...
-                line=f"-o results -l {slurm_job_data_test} -n unit_test -e local",
-                cell="a+b",
-            )
+
+        ip.run_cell_magic(
+            magic_name="submit_job",
+            line=f"-o results -l {slurm_job_data_test} -n unit_test -e local",
+            cell="results = a+b ",
+        )
+        time.sleep(4)
```
DZPM, line 63: Why this `time.sleep(4)` ?

To make sure it runs, right. why not 3, or 5? Will take different in different machines

JavierSab (approved):
lgtm!

tests/test_slurm.py
GuillermoAbadLopez: I just deleted the time.sleep and the check (since its already checked above)

There now we will only check it doesn't error, which is what it was supposed to do 👍

DZPM (approved):

DZPM:
Good! If the missing `sleep(4)` breaks a test later (maybe in less powerful machine), we can easily add it back

PR #908, opened 2025-03-25
Quickfix d5a

jordivallsq:
Fixed a minor error with the D5a crashing for the incorrect number of DACs (4 insteaqd of 16)


GuillermoAbadLopez (approved):
Looks good to me (if indeed the number of dacs is 16)

PR #918, opened 2025-03-28
QHC 933 - (BUG) and improvement QbloxDraw

flavie-lebars:
- bug corrected, the user can now play the same waveform in one play …instruction
- bug corrected, if the Q1ASM had a set gain of 0 it would replace it to 1 by default because of "i_val or default", a 0 would be considered faulty and hence replaced by a the default which was 1- fixed by removing default since those lines can't play unless a gain is given
- the voltage bounds of the RF modules as been set to be the same as their baseband counterpart (0.5 for QRM/RF and 2.5 for QCM/RF)
- if drawing is done through qprogram directly, the y axis of the plotting now reads Amplitude [a.u.]
- subplots have been removed, all lines plot on the same plots and can be toggled on/off

### DIFF
--- a/src/qililab/instruments/qblox/qblox_draw.py
+++ b/src/qililab/instruments/qblox/qblox_draw.py
@@ -16,7 +16,6 @@ import re
 
 import numpy as np
 import plotly.graph_objects as go
-from plotly.subplots import make_subplots
 
 
 class QbloxDraw:
--- a/src/qililab/instruments/qblox/qblox_draw.py
+++ b/src/qililab/instruments/qblox/qblox_draw.py
@@ -41,7 +41,7 @@ class QbloxDraw:
         elif action_type == "set_awg_offs":
             param = self._handle_offset(program_line, param)
         elif action_type == "set_awg_gain":
-            param = self._handle_gain_draw(program_line, param, 1, register)
+            param = self._handle_gain_draw(program_line, param, register)
         elif action_type == "play":
             data_draw = self._handle_play_draw(data_draw, program_line, waveform_seq, param)
         elif action_type == "wait":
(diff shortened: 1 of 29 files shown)


src/qililab/instruments/qblox/qblox_draw.py
```diff
@@ -347,12 +346,10 @@ def draw(self, sequencer, runcard_data=None, averages_displayed=False) -> dict:
                 parameters[bus]["intermediate_frequency"] = [IF]
                 parameters[bus]["ac_offset_i"] = parameters[bus]["offset_i"]
                 parameters[bus]["ac_offset_q"] = parameters[bus]["offset_q"]
-                if parameters[bus]["instrument_name"] == "QCM":
+                if parameters[bus]["instrument_name"] == "QCM" or parameters[bus]["instrument_name"] == "QCM-RF":
```
fedonman, line 349: Hint: Use `if parameters[bus]["instrument_name"] in {"QCM", "QCM-RF"}` for checking multiple equalities.

src/qililab/instruments/qblox/qblox_draw.py
```diff
@@ -347,12 +346,10 @@ def draw(self, sequencer, runcard_data=None, averages_displayed=False) -> dict:
...
                     parameters[bus]["max_voltage"] = 2.5
-                elif parameters[bus]["instrument_name"] == "QRM":
+                elif parameters[bus]["instrument_name"] == "QRM" or parameters[bus]["instrument_name"] == "QRM-RF":
```
fedonman, line 351: Same here.

fedonman (approved):
Great work!! :100: Left a pythonic suggestion, and is good to go :rocket: 

200w (2)

src/qililab/instruments/qblox/qblox_draw.py
flavie-lebars: that's much better thanks!

PR #924, opened 2025-04-01
ensure the plots of qblox draw don't output on the user's browser whe…

flavie-lebars:
when running tests via vs code, the tests of qbloxdraw have been modified such that the plots don't open on the user's browser


jjmartinezQT (approved):
look goods to me

PR #925, opened 2025-04-02
[QHC-949] Parallelize Tests

fedonman:
Github Actions now use `pytest-xdist` plugin to run tests in parallel. To run tests in parallel locally use `uv run pytest -n auto --dist loadfile`. The `--dist loadfile` flag is mandatory to avoid conflicts between tests that edit shared data, and should be planned for removal in the future.


jjmartinezQT (approved):
good job

PR #926, opened 2025-04-02
fixed D5a

jordivallsq:


src/qililab/instruments/qblox/qblox_d5a.py
```diff
@@ -151,8 +136,16 @@ def _set_ramping_rate(self, value: float | str | bool, channel_id: int, channel:
     @check_device_initialized
     def initial_setup(self):
         """performs an initial setup."""
-        for dac_index in self.dacs:
```
DZPM, line 154: Why did you move this inline method?

DZPM (approved):

src/qililab/instruments/qblox/qblox_d5a.py
jordivallsq: Because it lead to errors and to have the same structure as S4g for the sake of code symmetry

PR #928, opened 2025-04-04
Database platform implementation

jordivallsq:
The main purpose of this PR is to modify stream array to use databases straight from platform.stream_array(). The main changes implemented include:

- Platform: added `stream_array()` (to call platform.stream_array and avoid to hardcode the platform as an input to write it inside the database) and `save_measurement_results()` (to save a measurement already finished inside the database, for cases like old data or vna results).
- database.py: File locating all the database structure following [Oscar definition](https://www.notion.so/qilimanjaro/Database-structure-1a87eec14c53806dafe5e40df2a0dd9f) and following SQLAlchemy code architecture:
      - Cooldown metadata.
      - Sample metadata.
      - Measurement metadata (including all the functions necessary to save the measurement results.
      - DatabaseManager, a class to operate the database, the only one in the file which is not part of the SQLAlchemy structure but works with it). with functions to set the sample, cooldown, measurements and results and to modify the database (`head` and `tail`)
- StreamArray: modified version of stream array working with the database structure, the operation is similar to the previous one but now it saves directly to the database.

Supplementary modifications:

- Added type checks on qblox modules parameters: this was a fix for some issues that Oscar encountered while working with the databases as the values were not correctly set by the machine.
- Added SQLAlchemy and xarray (usefull tool for measurements) to the dependencies as they are required for databases.
- Relocated `save_results` and `load_results` from `data_management` to `result/result_management.py` for structure consistency. The load_results functions has been slightly changed to take into account different structures of data.

### DIFF
--- a/src/qililab/platform/platform.py
+++ b/src/qililab/platform/platform.py
@@ -1610,7 +1610,7 @@ class Platform:
 
         return result
 
-    def database_saving(
+    def db_real_time_saving(
         self,
         shape: tuple,
         loops: dict[str, np.ndarray],
@@ -1675,7 +1675,7 @@ class Platform:
             optional_identifier=optional_identifier,
         )
 
-    def save_measurement_results(
+    def db_save_results(
         self,
         experiment_name: str,
         results: np.ndarray,
@@ -1685,7 +1685,7 @@ class Platform:
         optional_identifier: str | None = None,
     ):
         """Uses the same StreamArray class as for live saving but it saves full chunks of data in the same format as platform.stream_array.
-        
+
         Example usage of this function:
 
             .. code-block:: python
--- a/tests/platform/test_platform.py
+++ b/tests/platform/test_platform.py
@@ -1362,12 +1362,12 @@ class TestMethods:
         with pytest.raises(AttributeError, match="Mixers calibration not implemented for this instrument."):
             platform.calibrate_mixers(alias=non_rf_readout_bus, cal_type=cal_type, channel_id=channel_id)
 
-    def test_database_saving(self, platform: Platform):
-        """Test database_saving function to save database from platform"""
+    def test_db_real_time_saving(self, platform: Platform):
+        """Test db_real_time_saving function to save database from platform"""
 
         shape = (2, 2)
         loops = {"test_amp_loop": np.arange(0, 2)}
-        experiment_name = "test_database_saving"
+        experiment_name = "test_db_real_time_saving"
         mock_database = MagicMock()
         db_manager = mock_database
         optional_identifier = "optional_identifier"
@@ -1376,19 +1376,19 @@ class TestMethods:
         qprogram = QProgram()
         qprogram.play(bus="drive_line_q0_bus", waveform=drive_wf)
 
-        database_saving = platform.database_saving(shape, loops, experiment_name, db_manager, qprogram, optional_identifier)
+        db_real_time_saving = platform.db_real_time_saving(shape, loops, experiment_name, db_manager, qprogram, optional_identifier)
 
-        assert database_saving.loops == loops
-        assert database_saving.results.shape == shape
-        assert database_saving.optional_identifier == optional_identifier
-        assert database_saving.platform == platform
-        assert database_saving.qprogram == qprogram
-        assert database_saving.db_manager == mock_database
-        assert database_saving.experiment_name == experiment_name
+        assert db_real_time_saving.loops == loops
+        assert db_real_time_saving.results.shape == shape
+        assert db_real_time_saving.optional_identifier == optional_identifier
+        assert db_real_time_saving.platform == platform
+        assert db_real_time_saving.qprogram == qprogram
+        assert db_real_time_saving.db_manager == mock_database
+        assert db_real_time_saving.experiment_name == experiment_name
 
     @patch("h5py.File")
-    def test_save_measurement_results(self, mock_h5file, platform: Platform):
-        """Test save_measurement_results functionto save from database from Platform"""
+    def test_db_save_results(self, mock_h5file, platform: Platform):
+        """Test db_save_results functionto save from database from Platform"""
 
         experiment_name = "experiment_name"
         loops = {"test_amp_loop": np.arange(0, 1)}
@@ -1402,13 +1402,13 @@ class TestMethods:
         qprogram = QProgram()
         qprogram.play(bus="drive_line_q0_bus", waveform=drive_wf)
 
-        platform.save_measurement_results(experiment_name, results, loops, db_manager, qprogram, optional_identifier)
+        platform.db_save_results(experiment_name, results, loops, db_manager, qprogram, optional_identifier)
 
         assert mock_h5file.called
 
     @patch("h5py.File")
-    def test_save_measurement_results_loop_dict(self, mock_h5file, platform: Platform):
-        """Test save_measurement_results functionto save from database from Platform"""
+    def test_db_save_results_loop_dict(self, mock_h5file, platform: Platform):
+        """Test db_save_results functionto save from database from Platform"""
 
         experiment_name = "experiment_name"
         loops = {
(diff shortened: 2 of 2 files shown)


jjmartinezQT:
Could we please add a description of the changes? Reviewer needs to understand motivation behind why some functions change from one place of Qililab to another so the review can be comprehensive. Thanks.

jjmartinezQT:
Stream array now only works with database? old behaviour has been removed or can we still use it if needed?

jordivallsq:
> Stream array now only works with database? old behaviour has been removed or can we still use it if needed? Rephrasing the question, how this affects for example BSC as a third party customer using Qililab in their workflow?

Stream array only works with database, I am remaking the old code for stream_array to revert that and have only a way to save the data only without a database

jjmartinezQT:
> > Stream array now only works with database? old behaviour has been removed or can we still use it if needed? Rephrasing the question, how this affects for example BSC as a third party customer using Qililab in their workflow?
> 
> Stream array only works with database, I am remaking the old code for stream_array to revert that and have only a way to save the data only without a database

Yes please, this is an architectural change that cannot affect backwards compatibility.

jjmartinezQT:
> > > > Stream array now only works with database? old behaviour has been removed or can we still use it if needed? Rephrasing the question, how this affects for example BSC as a third party customer using Qililab in their workflow?
> > > 
> > > 
> > > Stream array only works with database, I am remaking the old code for stream_array to revert that and have only a way to save the data only without a database
> > 
> > 
> > Yes please, this is an architectural change that cannot affect backwards compatibility.
> 
> Reitroduced the function `stream_results` with the same functionality as before, the only existing change with the past is the name of the class inside (StreamArray -> RawStreamArray) which is not percieved by the user while caling `ql.stream_array(...)`

Thanks! Can we include some minimal examples to check what the workflow is with/without database for results? I am trying to get around the new platform entry points.

jordivallsq:
I am on it, I will update the docs for all the new features. The example shall be in stream_array.py or in the rellevant platform.py function? I would say that since the platform function is the one used it shall have the example, as it uses platform

src/qililab/platform/platform.py
```diff
@@ -1590,3 +1592,119 @@ def draw(self, qprogram: QProgram, time_window: int | None = None, averages_disp
         result = qblox_draw.draw(sequencer, runcard_data, time_window, averages_displayed)
 
         return result
+
+    def stream_array(
```
jjmartinezQT, line 1596: Still have not comed up with a good name myself either, but if this is the platform entry point to save results using databases, maybe 'database_saving' could be the name? We need something very descriptive so if an user that faces Qililab for the first time sees the method he/she can know what it is about by the name.

src/qililab/platform/platform.py
```diff
@@ -1590,3 +1592,119 @@ def draw(self, qprogram: QProgram, time_window: int | None = None, averages_disp
...
+                platform.initial_setup()
+                platform.turn_on_instruments()
+
+                (experiment definition)
+
+                stream_array = platform.stream_array(
+                    shape=(len(if_sweep), 2),
+                    loops={"frequency": if_sweep},
+                    platform=platform,
+                    experiment_name="resonator_spectroscopy_cw",
```
jjmartinezQT, line 1628:
```suggestion
                    experiment_name="resonator_spectroscopy",
```

Sounds like a picky request I know but to start going back to documenting the library for external users let's keep the most simple name of the experiment.

src/qililab/platform/platform.py
```diff
@@ -1590,3 +1592,119 @@ def draw(self, qprogram: QProgram, time_window: int | None = None, averages_disp
...
+            shape=shape,
+            loops=loops,
+            platform=self,
+            qprogram=qprogram,
+            experiment_name=experiment_name,
+            db_manager=db_manager,
+            optional_identifier=optional_identifier,
+        )
+
+    def save_measurement_results(
```
jjmartinezQT, line 1662: I am trying to think a way, if it is worthy, of having only one entry point to save results. This two saving methods seem very similar but only changing in small particularities, do you think it is worth the effort of combining them? Maybe once there is an example here in the docstrings I see it more clearly it is not a good idea to combine them.

I have been thinking for a bit now if it is a good idea to combine live saving with the platform.session functionality but not completely sure. Maybe a future idea after discussing with experimentalists to get UI feedback.

src/qililab/result/result_management.py
```diff
@@ -0,0 +1,132 @@
...
+
+import os
+from datetime import datetime
+from pathlib import Path
+
+import h5py
+import numpy as np
+
+
+def save_results(results: np.ndarray, loops: dict[str, np.ndarray], data_path: str, name: str | None = None) -> str:
```
jjmartinezQT, line 23: Having this file to manage results is a good idea, well thought.

src/qililab/platform/platform.py
```diff
@@ -1590,3 +1592,119 @@ def draw(self, qprogram: QProgram, time_window: int | None = None, averages_disp
...
+                db_manager.set_sample_and_cooldown(sample=sample, cooldown=cooldown)
+
+                platform = ql.build_platform(runcard=runcard)
+                platform.connect()
+                platform.initial_setup()
+                platform.turn_on_instruments()
+
+                (experiment definition)
+
+                stream_array = platform.stream_array(
```
jordivallsq, line 1624:
```suggestion
                stream_array = platform.database_saving(
```

src/qililab/platform/platform.py
jordivallsq, line 1596:
```suggestion
    def database_saving(
```

src/qililab/platform/platform.py
jordivallsq: Right now those are two separate functions because there are a few differences in terms of input variables but they are really important, mainly results and shape, those are two mutually exclusive variables that one function have and not the other but they are not optional for either function. I could merge both into the same function but it would be messy, require some exceptions to prevent issues and the documentation for both will look confusing

src/qililab/platform/platform.py
jordivallsq: Right now I changed into the suggested name, I also thought on stream_database relating to the live saving but wasn't accurate

tests/result/test_database.py
```diff
@@ -0,0 +1,534 @@
...
+
+        # Act
+        measurement = db_manager.add_measurement("exp1", experiment_completed=True)
+
+        # Assert
+        expected_path = "/home/jupytershared/data/sampleA/cdX/2023-01-01/12_00_00/exp1.h5"
+        assert measurement.result_path == expected_path
+        assert db_manager._mock_session.add.called_once
+        assert db_manager._mock_session.commit.called_once
+        assert mock_makedirs.called_once_with("/home/jupytershared/data/sampleA/cdX/2023-01-01/12_00_00")
```
jordivallsq, line 370:
```suggestion
        assert mock_makedirs.called_once_with("/mnt/home.local/jupytershared/data/sampleA/cdX/2023-01-01/12_00_00/exp1.h5")
```

tests/result/test_database.py
```diff
@@ -0,0 +1,534 @@
...
+
+        fixed_time = datetime.datetime(2023, 1, 1, 12, 0, 0)
+        mock_datetime.datetime.now.return_value = fixed_time
+        mock_datetime.datetime.strftime = datetime.datetime.strftime  # fallback
+
+        # Act
+        measurement = db_manager.add_measurement("exp1", experiment_completed=True)
+
+        # Assert
+        expected_path = "/home/jupytershared/data/sampleA/cdX/2023-01-01/12_00_00/exp1.h5"
```
jordivallsq, line 366:
```suggestion
        expected_path = "/mnt/home.local/jupytershared/data/sampleA/cdX/2023-01-01/12_00_00/exp1.h5"
```

jjmartinezQT (approved):
Great one.

PR #929, opened 2025-04-04
Update pyproject.toml for right qblox instruments version

jjmartinezQT:
This PR reverts to the right version for qblox-instruments, which we lost somehow when we introduced uv.


fedonman:
I added the changes to `uv.lock` too. Run `uv sync --all-groups` when changing dependencies to reflect changes to the `uv.lock`.

fedonman:
Just `uv sync` works too, but `uv sync --all-groups` takes into accound the `docs` dependencies as well.

fedonman (approved):
approved-target-lady

PR #931, opened 2025-04-10
Fixed the experiment class results transposed bug

jordivallsq:


fedonman (commented):
Nice catch Jordi!! Have we tested in hardware? Also, is it possible to write a test for this so it's not introduced again in the future?

jordivallsq:
> Nice catch Jordi!! Have we tested in hardware? Also, is it possible to write a test for this so it's not introduced again in the future?

Let me see how to create a test for this, comparing the results shape and the generated shape. I will check the current tests.
It has been tested in Hardware by me and not it does not crash nor fills empty spaces with 0s

jordivallsq:
I implemented the suggested tests

tests/qprogram/test_experiment_executor.py
```diff
@@ -240,3 +256,12 @@ def test_execute(self, platform, experiment, qprogram, crosstalk):
...
         platform.assert_has_calls(expected_calls, any_order=False)
+
+    def test_execute_check_metadata(self, platform, experiment_two_loops):
+        """Tets to validate the correct order of the results"""
+        executor = ExperimentExecutor(platform=platform, experiment=experiment_two_loops)
+        _ = executor.execute()
+
+        results_metadata_shape = executor._metadata["qprograms"]["QProgram_0"]["measurements"]["Measurement_0"]["shape"]
+
+        assert (2, 3, 11, 2) == results_metadata_shape
```
fedonman, line 267: Why 2 is first and 3 is second? First loop `for_loop(bias, 0.0, 1.0, 0.5):` runs 3 times, and second loop `with experiment_two_loops.loop(frequency, values=np.array([2e9, 3e9])):` runs 2. So it should be `(3, 2, 11, 2)`

fedonman (changes requested):
I am requesting changes since it seems we have introduced another bug with this. Results shape should always be `(Experiment Loop 1, ..., Experiment Loop N, QProgram Loop 1, ..., QProgram Loop M, 2)`. With this change experiment loops are inverted.

Idea: Maybe we need to do this inversion only when storing the data? And not change the creation of metadata.

tests/qprogram/test_experiment_executor.py
jjmartinezQT: This test does not really validates the correct order of the results, rather the metadata one. I am trying to inverse the real order of results so it is faithful to original metadata design idea and it is the storing of results what must change. This test has to be better thought as well.

fedonman:
Closing since this is fixed in https://github.com/qilimanjaro-tech/qililab/pull/932

PR #932, opened 2025-04-17
[QHC-950] Fix ExperimentResults Data

fedonman:


jjmartinezQT (approved):
Check the change and unittests, great work.

nodding-yes

PR #933, opened 2025-04-29
QHC 960 improve the drawing for qblox

flavie-lebars:
- Bug corrected, the move commands were being read correctly once but were not being updated after -  this caused problem with loops
- In the plots, the name of the flux will now appear as "bus name" Flux, similar to I and Q when hardware modulation is enabled.
- An additional argument has been added, timeout allowing the user to stop the plotting after the specified number of ns have been plotted. The plotting might not be the precise number of ns inputted. For example, if the timeout is 100 ns but there is a play operation of 150 ns, the plots will display the data until 150 ns


src/qililab/instruments/qblox/qblox_draw.py
```diff
@@ -107,7 +111,7 @@ def _handle_play_draw(self, data_draw, program_line, waveform_seq, param):
                 param[key].extend([param[key][-1]] * len(scaled_array))
         return data_draw
 
-    def _handle_wait_draw(self, data_draw, program_line, param):
+    def _handle_wait_draw(self, data_draw, program_line, param, time=None):
         """Play a wwait by appending the stored data list.
```
jjmartinezQT, line 115:
```suggestion
        """Play a wait by appending the stored data list.
```
small typo

jjmartinezQT (changes requested):
Looks good to me, small typo to fix in documentation.

jjmartinezQT (approved):
great work!

PR #936, opened 2025-05-06
Qhc 965 save the database information first locally and then inside

jordivallsq:


jjmartinezQT (approved):
gw

jordivallsq:
I have made some minor improvements to the platform to reduce the number of inputs in the functions `execute_experiment`, `db_real_time_saving` and `db_save_results`. Essentially the variable base_path is no longer mandatory while the default is the `platform.experiment_results_base_path` (as this can be also modified easily by the user).

The main code changed is located at `platform.py` lines 1683-1687, 1746-1750 and at `experiment_executor.py` lines 529-532. It is the same condition in different rellevant places.

jjmartinezQT (approved):
great work!

PR #938, opened 2025-05-13
[QHC-928] Experiment class database implementation

jordivallsq:
Added the database implementation inside `Experiment`. `load_db_manager` function added to platform to add the db_manager. For `Experiment` class, added an input variable (`platform.save_experiment_results_in_database`) to define if working with databases or not (default set as True), if True, the database is automatically generated using the `Experiment` variables as information.

Code example:
```
platform.load_db_manager(db_ini_path = "path")  # To load the database manager. Optional path.
platform.save_experiment_results_in_database = True  # Default is True.
experiment = Experiment(label="Experiment_name", description = "optional description")
platform.execute_experiment(experiment)
```


src/qililab/qprogram/experiment_executor.py
```diff
@@ -112,6 +114,10 @@ def __init__(
         self,
         platform: "Platform",
         experiment: Experiment,
+        database: bool = True,
+        db_manager: DatabaseManager | None = None,
+        base_path: str | None = None,
+        optional_identifier: str | None = None,
```
fedonman, line 120: Why do we need all these information here since we have access to the `platform`?

Platform is the one responsible for handling the database management, so we shouldn't have duplicated source of truths.

Users will `platform.load_db_manager()` and give all information needed once there.

src/qililab/platform/platform.py
```diff
@@ -938,6 +942,9 @@ def execute_annealing_program(
     def execute_experiment(
         self,
         experiment: Experiment,
+        base_path: str | None = None,
+        database: bool = True,
+        optional_identifier: str | None = None,
```
fedonman, line 947: These should be removed as well based on the other comment.

docs/releases/changelog-dev.md
```diff
@@ -49,8 +53,12 @@
 
 [#945](https://github.com/qilimanjaro-tech/qililab/pull/945)
 
+
 ### Breaking changes
 
+- Modified file structure for functions `save_results` and `load_results`, previously located inside `qililab/src/qililab/data_management.py` and now located at `qililab/src/qililab/result/result_management.py`. This has been done to improve the logic behind our libraries. The init structure still works in the same way, import `qililab.save_results` and import `qililab.load_results` still works the same way.
+[#928](https://github.com/qilimanjaro-tech/qililab/pull/928)
```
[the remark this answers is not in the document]
jordivallsq: same reply as before

src/qililab/platform/platform.py
```diff
@@ -983,7 +1013,10 @@ def execute_experiment(
         executor = ExperimentExecutor(
             platform=self,
             experiment=experiment,
+            database=database,
+            db_manager=self.db_manager,
```
fedonman, line 1017: We don't need to pass the `db_manager` as separate parameter. It can be accessed from `platform`.

Also, based on the other comments, `database` and `optional_identifier` could be potentially removed.

src/qililab/qprogram/experiment_executor.py
```diff
@@ -111,7 +113,10 @@ def __init__(
         self,
         platform: "Platform",
         experiment: Experiment,
+        database: bool = True,
+        db_manager: DatabaseManager | None = None,
         base_path: str | None = None,
+        optional_identifier: str | None = None,
```
fedonman, line 119: same here as previous comments, parameters could be removed.

src/qililab/qprogram/experiment_executor.py
```diff
@@ -153,12 +158,36 @@ def __init__(
...
+        self.optional_identifier: str | None = optional_identifier
+
+    def _create_database(
+        self,
+    ):
+        if not self.db_manager:
+            try:
+                self.db_manager = get_db_manager()
+            except ReferenceError:
+                raise ReferenceError("Missing initialization information at the desired database '.ini' path.")
```
fedonman, line 186: Only `Platform` should create databases.

src/qililab/qprogram/experiment_executor.py
```diff
@@ -255,6 +284,14 @@ def finalize_measurement_structure():
             execution_time=0.0,
             qprograms={},
         )
+        if self.database:
```
fedonman, line 287: Following previous comments, this could be taken from platform, `if self.platform.save_experiment_results_in_databse:` or `if self.platform.db_manager is not None:`

src/qililab/qprogram/experiment_executor.py
```diff
@@ -581,13 +618,19 @@ def execute(self) -> str:
         # Create file path to store results
         results_path = self._create_results_path(executed_at=executed_at)
 
+        # Load / Create database manager, cooldown and sample names
+        if self.database:
+            self._create_database()
```
fedonman, line 623: Database would not be created here.

fedonman (changes requested):
Left some more comments mainly on the Platform's responsibility for the database management.

I have another worry: `ExperimentResultsWriter` was supposed to be responsible for writing results to the H5 file. Now it has become responsible for also sending the data to live plot and to database too. This makes the class a bit overcomplicated. I think the best option would be to somehow split the writers to different classes (one for writing to h5 file, one for sending the data to LivePlot, one for writing the (different) data to the database. I will think a bit about it and we can discuss in a daily.

docs/releases/changelog-dev.md
```diff
@@ -39,16 +39,41 @@
...
+platform.load_db_manager(db_ini_path = "path")  # To load the database manager. All parameters are optional.
+platform.execute_experiment(..., optional_identifier = "string")  # The typical experiment execution.
+...
+```
+
+By deafault the results and metadata will be saved in database format, so unless specified the user can also use this:
+
+```
+platform.execute_experiment(..., optional_identifier = "string")
+```
```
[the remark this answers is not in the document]
jordivallsq: maybe the naming is not the most adequate but is basically an additional comments

src/qililab/qprogram/experiment_executor.py
jordivallsq: ok, the function already exists, I only need to call platform from here

src/qililab/qprogram/experiment_executor.py
jordivallsq: I implemented `f self.platform.save_experiment_results_in_databse`

docs/releases/changelog-dev.md
```diff
@@ -39,16 +39,41 @@
...
+
+To NOT use the database saving format it just requires to set the variable database as False:
+
+```
+platform.execute_experiment(..., database = False)
+```
```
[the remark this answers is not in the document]
jordivallsq: I did exactly this

docs/releases/changelog-dev.md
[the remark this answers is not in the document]
jordivallsq: I implemented otional identifier as a platform variable

src/qililab/platform/platform.py
```diff
@@ -344,6 +345,15 @@ def __init__(self, runcard: Runcard):
         self.flux_parameter: dict[str, int | float | bool | str] = {}
         """Flux dictionary with information for the get parameter (only used on FLUX parameters)"""
 
+        self.db_manager: DatabaseManager | None = None
+        """Database manager for experiment class and db stream array"""
+
+        self.save_experiment_results_in_database: bool = True
+        """Database trigger to define if the experiment metadata will be saved in a database or not"""
+
+        self.db_optional_identifier: str | None = None
```
fedonman, line 354: My only concern is this field. 

First, the name is a bit misleading. `identifier` is used to something that is unique and we query based on that. If our purpose is to have a comment or additional information, using these exact words is better.,

Secondly, this is gonna change for every experiment right? So maybe it is better to be an optional field of `Experiment` class? E.g. `description: str | None`. And take the information from there when you save the metadata, like you do with the `Experiment.label`.

fedonman (commented):
Awesome work addressing the comments! I have one final concern.

Also, please c/p the changelog entry to the PR description.

src/qililab/platform/platform.py
jordivallsq: I agree, this makes more sense, I moved the description to the experiment class init and changed optional identifier names for platform functions

src/qililab/result/experiment_results_writer.py
```diff
@@ -82,6 +84,24 @@ class ExperimentMetadata(TypedDict, total=False):
...
+        executed_at (datetime): Timestamp when the experiment started execution.
+        execution_time (float): Time taken for the execution in seconds.
+        qprograms (dict[str, QProgramMetadata]): Quantum programs included in the experiment.
+    """
+
+    experiment_name: str
+    base_path: str
+    cooldown: str | None
+    sample_name: str | None
+    optional_identifier: str | None
```
fedonman, line 102: Can this also change to `description` or `experiment_description` easily or a change in the DB Schema is needed?

jordivallsq: I thought the same but it would change the current db schema, I can ask Oscar but is a change important enough for a diferent PR

fedonman:
@jordivallsq Edit changelog entry with the latest code changes and address this comment, and is good to go! 🚀 

> Awesome work addressing the comments! I have one final concern.
> 
> Also, please c/p the changelog entry to the PR description, and the `[QHC-928]` prefix to the title

fedonman (approved):
Great work! 💯 🚀 

giphy-downsized

PR #942, opened 2025-05-26
[QHC-973] Adding improved error, to trace origin in Runcard

GuillermoAbadLopez:


jordivallsq (approved):
I see no issues in the implementation, quite straightforward. Nice work!

PR #945, opened 2025-06-02
Qhc 969 qblox draw to improve following demo

flavie-lebars:
Improvements:

- The code is now tracking real time/classical time. The implications are the following:
  - A play can interrupt a preceding play - this replicates the hardware behaviour.
  - Acquisitions are now plotted, when hovering the mouse on the plot, the user can see the index of the acquisition (they are plotted by default but this can be set to False if desired with the argument 'acquisition_showing').
  - Acquire and Play can overlap each other as they are real time commands - this replicates the hardware behaviour.
        The integration length of the acquire is retrieved from the platform if given. If plotting directly from qprogram, the integration length is set as the duration of the acquire.
- The sub and not commands have been implemented.
- If QbloxDraw is given a Q1ASM command it is not programmed for, it will raise a NotImplementedError.
- The _handle_play() used to loop through all the waveform indices, now it exits in the loop as soon as I and Q have been found, this is more efficient.
- When running the code from the platform, the bus_mapping can be provided.
- The overall plotting design has been improved. A plotly colour palette is used. The I and Q are the same color but in a different shade.

Bug fixes:
- The sequencer offsets given from the runcard (offset_i and offset_q in the runcard) were being applied similarly to the DAC offsets, when they should have been treated like the Q1ASM offsets - this has been fixed and those sequencer offsets have been renamed sequencer_runcard_offset_i and  sequencer_runcard_offset_q instead of ac_offsets_i and ac_offsets_q for improved clarity.
- get_value() in the QbloxDraw class now checks that the given string is a float, it used to check x.isdigit() which didn't work for negative values.

### DIFF
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -38,11 +38,11 @@
 - QbloxDraw:
     - The code is now tracking real time/classical time. The implications are the following:
       - A play can interrupt a preceeding play - this replicates the hardware behaviour.
-      - Acquisition are now plotted, when hoovering the mouse on the plot, the user can see the index (they are plotted by default but this can be set to False if desired).
+      - Acquisitions are now plotted, when hovering the mouse on the plot, the user can see the index of the acquisition (they are plotted by default but this can be set to False if desired with the argument 'acquisition_showing').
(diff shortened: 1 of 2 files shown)


jjmartinezQT (approved):
Very good improvements!

PR #948, opened 2025-06-21
introducing environment variable for runcard

jjmartinezQT:
This PR brings a request from BSC to try to control a bit runcard usage from their users potential misusage.

If the environment variable is present in the python environment, the runcard argument provided by the user when calling build_platform() will be overwritten by the value the environment variable points to. This is the first step of two that we are introducing to try limit the potential dangers of QuantumSpain users being able to use Qililab in the QuantumSpain computers without an API supervising what is executed or which runcard is used.

### DIFF
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -4,9 +4,15 @@
 
 ### Improvements
 
+- A safe control for misusage of runcard in client deployments has been set. If the environment variable `ENVIRONMENT_RUNCARD` is present, the user provided
+parameter for the runcard in `data_management.build_platform` method will be ignored and instead the value of the environment variable will be used to build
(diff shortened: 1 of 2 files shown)


DZPM (approved):

src/qililab/data_management.py
```diff
@@ -47,6 +48,9 @@ def save_platform(path: str, platform: Platform) -> str:
         Qililab will use the name of the platform to create the YAML file. If ``platform.name == "galadriel"``, a file
         will be created in ``examples/runcards/galadriel.yml``.
     """
+    if os.environ.get("ENVIRONMENT_RUNCARD"):
+        raise ValueError("No permissions to save the runcard to disk.")
```
DZPM, line 52: Should we document this in the save_platform() docstring?

jjmartinezQT: Yes once this mitigation action has been agreed for BSC I will document it.

PR #950, opened 2025-06-26
Qhc 1014 implement hardware loop over time for qblox

flavie-lebars:
Implement hardware loop over time in qblox.
Also introduces VariableExpression to qprogram, this is the building block required to set a variable expression such as qp.wait(bus, time+constant). For now, the operations are limited to time+constant, constant+time, time-constant, constant-time; and they are only implement for a loop over the Time domain

### DIFF
--- a/src/qililab/platform/platform.py
+++ b/src/qililab/platform/platform.py
@@ -1610,6 +1610,9 @@ class Platform:
         qblox_draw = QbloxDraw()
         compiler = QbloxCompiler()
         sequencer = compiler.compile(qprogram)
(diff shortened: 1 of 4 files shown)


fedonman (changes requested):
Amazing work Flavie!! A dream coming true :) 

Obviously, the main concern is that implementation is by its nature complex, so we need to take care that we remember in the future what we have actually implemented and how.

I did a preliminary reviewing about variables naming, helpful docstrings and comments and eliminating magic numbers.

Apply these and I will continue with some refactorings in the actual code.

PR #952, opened 2025-06-30
Sudden Net Zero waveform

jjmartinezQT:
This PR brings the SNZ waveform to QProgram set of waveforms.


src/qililab/waveforms/snz.py
```diff
@@ -0,0 +1,82 @@
...
+        # raise warning if we are rounding
+        if (self.duration / resolution) % 1 != 0 or (half_pulse_t / resolution) % 1 != 0:
+            logger.warning(  # pragma: no cover
+                "Envelope length rounded to nearest value %d from division full_snz_duration (%s) / resolution (%s) = %s",
+                len(envelope), self.duration, resolution, self.duration / resolution
+            )
+
+        envelope[:half_pulse_t] = self.amplitude * np.ones(half_pulse_t)  # positive square halfpulse
+        envelope[half_pulse_t] = self.b * self.amplitude  # impulse b
+        envelope[half_pulse_t + 2 + self.t_phi:] = 0  # t_phi
```
jordivallsq, line 70: This should be deleted as this function is overwritten by line 72 and in the case that you intended to write `envelope[half_pulse_t+1:] = 0` this is already true as you define `envelope = np.zeros(round(self.duration / resolution))`. So anyways, this line is redundant.

src/qililab/waveforms/snz.py
```diff
@@ -0,0 +1,82 @@
...
+        envelope = np.zeros(round(self.duration / resolution))
+        # raise warning if we are rounding
+        if (self.duration / resolution) % 1 != 0 or (half_pulse_t / resolution) % 1 != 0:
+            logger.warning(  # pragma: no cover
+                "Envelope length rounded to nearest value %d from division full_snz_duration (%s) / resolution (%s) = %s",
+                len(envelope), self.duration, resolution, self.duration / resolution
+            )
+
+        envelope[:half_pulse_t] = self.amplitude * np.ones(half_pulse_t)  # positive square halfpulse
+        envelope[half_pulse_t] = self.b * self.amplitude  # impulse b
```
jordivallsq, line 69: Is b a coefficient or a voltage value? in case it is a voltage value you should change this line (and line 71) to not multiply with the amplitude and if its otherwise you should state in the documentation that is a coefficient multiplying the amplitude

jordivallsq (approved):
Nice implementation!!!
I left some comments about the snz function

PR #955, opened 2025-07-09
Qhc 954 implement active reset in qblox

flavie-lebars:
**Active reset for transmon qubits in QBlox**

  Implemented a feedback-based reset for QBlox: measure the qubit, and if it is in the \|1⟩ state apply a corrective DRAG pulse; if it is already in \|0⟩ (ground state), do nothing. This replaces the relaxation time at the end of each experiment with a much faster, conditional reset.
  This has been implemented as **`qprogram.qblox.measure_reset(bus: str, waveform: IQPair, weights: IQPair, control_bus: str, reset_pulse: IQPair, trigger_address: int = 1, save_adc: bool = False)`** 

  It is compiled by the QBlox compiler as:
    1. `latch_rst 4` on the control_bus
    2. play readout pulse 
    3. acquire
    4. sync the readout and control buses
    5. wait 400 ns on the control bus (trigger-network propagation)
    6. `set_conditional(1, mask, 0, duration of the reset pulse)` → enable the conditional
    7. Play the reset pulse on the control bus
    8. `set_conditional(0, 0, 0, 4)` → disable the conditional  
    For the control bus, `latch_en 4` is added to the top of the Q1ASM to enable trigger latching.

  Notes:
    - The 400 ns wait inside `measure_reset` corresponds to the propagation delay of the Qblox trigger network. This figure is conservative as the official guideline is 388ns.
    - Users may supply any IQPair for the reset_pulse, though DRAG pulses are recommended to minimize leakage.
    - After `measure_reset`, users should insert a further wait as needed to allow the readout resonator to ring down before subsequent operations.
    - On compilation, `cluster.reset_trigger_monitor_count(address)` is applied to zero the module’s trigger counter. And the qcodes parameters required to set up the trigger network are implemented by the QbloxQRM class.
    - The Qblox Draw class has been modified so that `latch_rst` instructions are interpreted as a `wait`, and all `set_conditional` commands are ignored.

### DIFF
--- a/src/qililab/qprogram/qprogram.py
+++ b/src/qililab/qprogram/qprogram.py
@@ -214,20 +214,20 @@ class QProgram(StructuredProgram):
                     if isinstance(control_bus, str) and control_bus in bus_mapping:
                         setattr(block.elements[index], "control_bus", bus_mapping[control_bus])
                     new_latch_enabled = []
-                    for bus in self.qblox.latch_enabled:
+                    for bus in copied_qprogram.qblox.latch_enabled:
                         if bus in bus_mapping:
                             new_latch_enabled.append(bus_mapping[bus])
                         else:
                             new_latch_enabled.append(bus)
-                    self.qblox.latch_enabled = new_latch_enabled
+                    copied_qprogram.qblox.latch_enabled = new_latch_enabled
 
                     new_trigger_network_required = {}
-                    for bus, value in self.qblox.trigger_network_required.items():
+                    for bus, value in copied_qprogram.qblox.trigger_network_required.items():
                         if bus in bus_mapping:
                             new_trigger_network_required[bus_mapping[bus]] = value
                         else:
                             new_trigger_network_required[bus] = value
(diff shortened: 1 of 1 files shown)


src/qililab/qprogram/operations/measure_reset.py
```diff
@@ -0,0 +1,39 @@
...
+from qililab.qprogram.operations.operation import Operation
+from qililab.waveforms import IQPair
+from qililab.yaml import yaml
+
+
+@yaml.register_class
+class MeasureReset(Operation):
+    def __init__(
+        self,
+        measure_bus: str,
```
fedonman: Minor thing based on the speculation that Operations will be subclassed in the future, e.g. `MeasureReset(Measure)` could we rename this to `bus` so we won't have breaking change in the future?

fedonman (commented):
Great implementation! 💯 💯 Left some minor comments, and one more important on the `Sync` during `MeasureReset`.

PR #959, opened 2025-07-22
QUICKFIX: Scope store enabled neds another order of operation calls

jordivallsq:
I implemented a super quickfix involving the last changes implemented in set_parameter for SCOPE_STORE_ENABLED where the order of the functions called affected the result, this issue didn't cause any error on qililab but the fix implemented here (https://github.com/qilimanjaro-tech/qililab/pull/956) didn't work


jjmartinezQT (approved):
Once tested in hardware this should be good to go.

jordivallsq:
We tested the quickfix with Oscar already and he is working with a branch that has it. I will fix the tests and merge it

PR #963, opened 2025-07-29
qhc-1054-qblox-draw-to-remove-figshow

flavie-lebars:
Previously, `platform.draw(qprogram)` and `qprogram.draw()` returned the plotly object and the raw data being plotted. Now they return only the plotly object. This change ensures: 
  - When calling `qprogram.draw()` or  `platform.draw(qprogram)` directly, the figure is displayed.
  - When assigning it to a variable (e.g., `plotly_figure = qprogram.draw()` or  `plotly_figure = platform.draw(qprogram)`), the figure is stored but not automatically shown (since `figure.show()` has been removed from QbloxDraw).

  If the user needs access to the underlying data, it can be retrieved as follows:
    ```
    plotly_figure = qprogram.draw()
    plotly_figure.data
    ```
Note: QbloxDraw class continues to return both, the plotly object and the dictionary of raw data.

### DIFF
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -4,12 +4,25 @@
 
 ### Improvements
 
+- Previously, `platform.draw(qprogram)` and `qprogram.draw()` returned the plotly object and the raw data being plotted. Now they return only the plotly object. This change ensures: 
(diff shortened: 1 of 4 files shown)


jjmartinezQT (approved):
great work.

PR #964, opened 2025-07-30
[QHC-1056] Changing `desynch_sequencers` loop, from `instruments` to `instrument_controllers`

GuillermoAbadLopez:
### SOLUTIONS:

The good solution would be to not do instrument.desync_sequencers() to the instruments not present in the buses, instrument_controllers or program, instead than desyncing all the instruments (the not connected to, since we only connect the ones present in instrument_controllers)

So in summary we should change the lines:

```python
```

for looping the instrument_controllers one, something like:

```python
for instrument_controller in self.instrument_controllers.elements:
    for instrument in instrument_controller.modules:
```

Or we could check which instruments are used in the program/buses (with a bit more complicated logic), and loop those instead.

So @juanjo.martinez if you think one solution is better than the others, I'll implement that one. Benefits I see with each one:

- intrument_controllers: it's what the `connect` loops, so this error would 100% not happen, since device will always exist (its created there). (but if some instrument is defined there, and not used in the program, we would desync it too, I don't know if that matters or not).

- program: related with the above, this one would only do desync to the actual used instrument for that circuit. But the error could appear if the runcard is illdefined and that instrument has not been connected.

- bus: I guess a this has only both the drawbacks, while no benefit?

#### (for now, I went with the first one, since ir was the easiest)

---

Other silly possible solutions:

- Doing the connection, to not only the instruments defined in instruments_controllers, but all, so they have a self.device
- Appart than leave the defined instruments, also leave them defined in instrument controllers too?

Already found silly solution:

- Comment/Remove redundant instruments in Runcard:
(We could add a specific error, pointing to this, so easier to recogn

### DIFF
--- a/tests/data.py
+++ b/tests/data.py
@@ -598,6 +598,7 @@ class Galadriel:
     ]
 
     instrument_controllers: list[dict] = [
+        qblox_qblox_cluster_controller,
         rohde_schwarz_controller_0,
         rohde_schwarz_controller_1,
         attenuator_controller_0,


src/qililab/platform/platform.py
```diff
@@ -1430,9 +1430,7 @@ def execute(
 
             # Create platform:
             platform = build_platform(runcard="<path_to_runcard>")
-            transp_config = DigitalTranspilationConfig(
-                routing=True, optimize=False, router=Sabre, placer=ReverseTraversal
-            )
+            transp_config = DigitalTranspilationConfig(routing=True, optimize=False, router=Sabre, placer=ReverseTraversal)
```
GuillermoAbadLopez, line 1433: This is just a autoformater change

GuillermoAbadLopez:
@flavie-lebars I am getting an error on the draw tests, can you maybe help me in here? 🙃

GuillermoAbadLopez:
Problem from tests solved in https://github.com/qilimanjaro-tech/qililab/pull/965

jjmartinezQT:
Chip will be ready finally on Monday, will let you know Guille when it is condensed and cold to use.

PR #965, opened 2025-07-31
Test qblox draw bug fixed

flavie-lebars:
The tests comparing the expected results to figure.data are failing sometimes. The comparison is currently done with the position in the list, but this can change, hence this PR ensures the checks are done checking the bus rather than the position in the list.


jjmartinezQT (approved):
great work

PR #968, opened 2025-08-08
Qhc 992 QDACII Qprogram compiler and actvive reset

jordivallsq:
Implementation of the triggered pulses. The drivers for the QDAC triggering and synchronization have been implemented in instruments and for the Qblox. Added wait_trigger and set_trigger inside the qprogram (including syncronization with other buses for Qblox).

Added Qdac compiler for Qprogram with all the necessary implementations for active reset.


src/qililab/qprogram/qprogram.py
```diff
@@ -294,15 +304,29 @@ def play(self, bus: str, waveform: Waveform | IQPair) -> None:
...
 
-    def play(self, bus: str, waveform: Waveform | IQPair | str) -> None:
+    def play(
+        self,
+        bus: str,
+        waveform: Waveform | IQPair | str,
+        dwell: int | None = None,
+        delay: int | None = None,
+        repetitions: int | None = None,
+    ) -> None:
```
fedonman, line 329: What are the `dwell`, `delay` and `repetitions` parameters do? Add docstrings to this method and overload methods.

fedonman: If they are only used in instrument-specific setting, like the QDAC, they should be only included in `_QdacInterface` so the users access them with `qp.qdac.play(...)`

src/qililab/qprogram/qprogram.py
```diff
@@ -477,23 +513,34 @@ def set_offset(self, bus: str, offset_path0: float, offset_path1: float | None =
...
+    def set_markers(self, bus: str, mask: str):
+        """Set the markers based on a 4-bit binary mask.
+
+        Args:
+            bus (str): Unique identifier of the bus.
+            mask (str): A 4-bit mask, where 0 means that the associated marker is open (no signal), and 1 means that the marker is closed (signal).
+        """
+        operation = SetMarkers(bus=bus, mask=mask)
+        self._active_block.append(operation)
+        self._buses.add(bus)
```
fedonman, line 536: These are Qblox-only operations, right? Then we should add them in `_QbloxInterface`.

src/qililab/qprogram/qprogram.py
```diff
@@ -717,6 +764,71 @@ def measure(
...
+                waveform (str): An identifier of a named waveform.
+            """
+
+        def play(
+            self,
+            bus: str,
+            waveform: Waveform | IQPair | str,
+            dwell: int | None = None,
+            delay: int | None = None,
+            repetitions: int | None = None,
```
fedonman, line 810: Add docstrings for `dwell`, `delay` and `repetitions` to all overload methods.

src/qililab/qprogram/quantum_machines_compiler.py
```diff
@@ -163,6 +165,7 @@ def compile(
         thresholds: dict[str, float] | None = None,
         threshold_rotations: dict[str, float] | None = None,
         calibration: Calibration | None = None,
+        qdac_buses: list = [],
```
fedonman, line 168: 1.Don't use mutable types for default argument values. 2. `list` should be annotated to what type it includes. E.g. do `qdac_buses: list[str] | None = None` and initialize `qdac_buses` within the method if needed.

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -159,6 +166,8 @@ def __init__(self) -> None:
         self._qprogram: QProgram
         self._buses: dict[str, BusCompilationInfo]
         self._sync_counter: int
+        self._markers: dict[str, str] | None
+        self._qdac_buses: list[str]
```
fedonman, line 170: It's more clear to define the buses that belong to Qblox and instead of this check:

```
        if element.bus in self._qdac_buses:
            return
```

do 

```
        if element.bus not in self._qblox_buses:
            return
```

It is also future-proof, since if we add another instrument like QDAC we would have to create a new list and add a new check everywhere, whereas by checking the qblox buses the code will always remain the samee.

tests/data.py
```diff
@@ -649,6 +649,227 @@ class Galadriel:
...
+        {RUNCARD.ALIAS: "readout", RUNCARD.INSTRUMENTS: ["qmm"], RUNCARD.CHANNELS: ["readout"]},
+        {RUNCARD.ALIAS: "qdac_bus_1", RUNCARD.INSTRUMENTS: ["qdac"], RUNCARD.CHANNELS: [1]},
+        {RUNCARD.ALIAS: "qdac_bus_2", RUNCARD.INSTRUMENTS: ["qdac"], RUNCARD.CHANNELS: [2]},
+    ]
+
+    runcard = {
+        RUNCARD.NAME: name,
+        RUNCARD.BUSES: buses,
+        RUNCARD.INSTRUMENTS: instruments,
+        RUNCARD.INSTRUMENT_CONTROLLERS: instrument_controllers,
```
fedonman, line 869: Would be amazing if you can avoid adding to this file and instead create a `runcard.yml` and make a fixture using `platform = build_platform(runcard)` like you do in the `test_qblox_cluster_controller`.

src/qililab/qprogram/quantum_machines_compiler.py
fedonman: And the same comment as in QbloxCompiler applies here too: It's better to validate against a list containing the quantum machines buses.

fedonman (changes requested):
Amazing work! 💯 I have done an initial review with some requested changes, and will continue with a deeper one involving the implementations.

Other things that are missing:
1. Changelog Entry
2. Title is misleading since the active reset is not included (?)

src/qililab/qprogram/quantum_machines_compiler.py
```diff
@@ -178,15 +181,16 @@ def traverse(block: Block):
...
-                if not handler:
-                    raise NotImplementedError(
-                        f"{element.__class__} operation is currently not supported in Quantum Machines."
-                    )
-                if isinstance(element, (InfiniteLoop, ForLoop, Loop, Average, Parallel, Block)):
-                    with handler(element):
-                        traverse(element)
-                else:
-                    handler(element)
+                if not isinstance(element, SetTrigger):
```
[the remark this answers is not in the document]
jordivallsq: I needed to ignore set trigger in case the qdac uses it but now it behaves like any other element. I created a handler ther skips it in case of being for the QDAC and raises a not implemented error if the user tries to execute it for QM.

src/qililab/qprogram/qblox_compiler.py
jordivallsq: adde `self._qblox_buses` and `self._qm_buses` inside their respective compilers

tests/data.py
jordivallsq: Sure! I moved them to an independent yml with a format approppiate for a runcrad.
I actually prefer this but I followed the examples that were already created. For the future I will create the new runcards and move the old ones in the folder that I created for it.

jordivallsq:
I added all the requested changes, feel free to unresolve the conversations, I just marked them as resolved to follow my progress

src/qililab/instruments/qdevil/qdevil_qdac2.py
```diff
@@ -158,31 +163,153 @@ def upload_waveform(self, waveform: Waveform, channel_id: ChannelID):
...
+            self.clear_trigger(trigger)
+
+        self._triggers[str(trigger)] = self._cache_dc[channel_id].allocate_trigger()
+
+        channel = self.device.channel(channel_id)
+        channel.write_channel(f'sour{"{0}"}:dc:mark:pstart {self._triggers[str(trigger)].value}')
+
+        self.device.connect_external_trigger(port=out_port, trigger=self._triggers[str(trigger)], width_s=width_s)
+
+    def set_start_marker_internal_trigger(self, channel_id: ChannelID, trigger: str):
```
fedonman: method docstring missing

src/qililab/platform/platform.py
```diff
@@ -354,6 +358,11 @@ def __init__(self, runcard: Runcard):
         self.db_manager: DatabaseManager | None = None
         """Database manager for experiment class and db stream array"""
 
+        self.save_experiment_results_in_database: bool = False
+        """Database trigger to define if the experiment metadata will be saved in a database or not"""
```
fedonman: This was moved to `QililabSettings`.

src/qililab/platform/platform.py
```diff
@@ -1148,7 +1157,7 @@ def execute_experiment(
 
     def compile_qprogram(
         self, qprogram: QProgram, bus_mapping: dict[str, str] | None = None, calibration: Calibration | None = None
-    ) -> QbloxCompilationOutput | QuantumMachinesCompilationOutput:
+    ) -> tuple(QbloxCompilationOutput | QuantumMachinesCompilationOutput, QdacCompilationOutput | None):  # type: ignore[valid-type]
```
fedonman: Returning `tuple` here will complicate things if we add more instruments e.g. Joel's or other vendors. Why not add add a class e.g. `QProgramCompilationOutput` that contains all instrument-specific compilations as properties? Something like:
```
class QProgramCompilationOutput:
    qblox: QbloxCompilationOutput | None
    qdac: QDacCompilationOutput | None
    quantum_machines: QuantumMachinesCompilationOutput | None
```

So we can easily add more instruments, and the return value of the method is very clear.

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -485,7 +555,9 @@ def _handle_add_waits(self, bus: str, duration: int):
         """
 
         if self._buses[bus].upd_param_instruction_pending:
-            if 4 < duration < 8:  # you cannot play an update param and then a wait bc both have a minimum of 4
+            if (
+                4 < duration and duration < 8
+            ):  # you cannot play an update param and then a wait bc both have a minimum of 4
```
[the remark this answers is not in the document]
jordivallsq: Fixed this issue, my formatter is already at 120 characters, but for some reason works differently.

I deactivated it already so for the latest PRs this will not be an issue unless I manually reformat things.

src/qililab/qprogram/qprogram.py
```diff
@@ -477,23 +504,23 @@ def set_offset(self, bus: str, offset_path0: float, offset_path1: float | None =
         self._active_block.append(operation)
         self._buses.add(bus)
 
+    @requires_domain("duration", Domain.Time)
+    def set_trigger(self, bus: str, duration: int, outputs: list[int] | int | None = None, position: str = "start"):
+        """Adds a delay on the bus and wait for an trigger signal to arrive.
+        Args:
+            bus (str): Unique identifier of the bus.
+            duration (int): Duration of the delay after the trigger is received. Minimum of 4 ns.
+        """
```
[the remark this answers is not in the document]
jordivallsq: Updated to the correct information

src/qililab/extra/quantum_machines/qprogram/quantum_machines_compiler.py
```diff
@@ -334,8 +344,8 @@ def _handle_for_loop(self, element: ForLoop):
 
         to_positive = stop >= start
         if to_positive:
-            return qua.for_(qua_variable, start, qua_variable <= stop, qua_variable + step)  # type: ignore[arg-type]
-        return qua.for_(qua_variable, start, qua_variable >= stop, qua_variable + step)  # type: ignore[arg-type]
+            return qua.for_(qua_variable, start, qua_variable <= stop, qua_variable + step)  # type: ignore
+        return qua.for_(qua_variable, start, qua_variable >= stop, qua_variable + step)  # type: ignore
```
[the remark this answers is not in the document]
jordivallsq: Fixed all type ignore, it was missing an operand as the same issue that is making us use arg-type raises a similar error under operand.

Same for all other comments with "Here too"

fedonman (approved):
Great implementation! 💯 💯

PR #970, opened 2025-08-19
[QHC 1065] Update Qibo from 0.2.15 to 0.2.20

jordivallsq:


jordivallsq:
> LGTM but verify that execution flow works correctly in hardware, cause I think current tests might not catch breaking changes of pandas / qibo.

Ok but I will request some assistance to test the pulses part, I can test executions using qprogram and all the functions used in house for measurements but circuits is trickier for me. I will request help from @GuillermoAbadLopez if needed

fedonman:
> Now all the tests pass but when I try to do a `pip install -e .[quantum-machines]` it crashes with the same issue of version dependencies ignoring the override dependencies>
> 
> ```
> ERROR: Cannot install qililab and qm-qua==1.2.1 because these package versions have conflicting dependencies.
> 
> The conflict is caused by:
>     qblox-instruments 0.16.0 depends on numpy
>     qibo 0.2.20 depends on numpy<3.0.0 and >=2.0.0
>     qm-qua 1.2.1 depends on numpy<2 and >=1.17.0; python_version >= "3.8" and python_version < "3.12"
> 
> To fix this you could try to:
> 1. loosen the range of package versions you've specified
> 2. remove package versions to allow pip to attempt to solve the dependency conflict
> ```
> 
> @Vyron do you know how to fix this?

All commands work for me:
```bash
uv sync --all-groups  # no QM installation
uv sync --all-groups --all-extras  # with QM installation
uv pip install -e .  # no QM installation
uv pip install -e .[quantum-machines]  # with QM installation
```

jordivallsq:
You are right, a `uv pip install -e .[quantum-machines]`, but then I try with only `pip install -e .[quantum-machines]` and it fails. It is ok as theintended way is working but I find it curious

fedonman:
> You are right, a `uv pip install -e .[quantum-machines]`, but then I try with only `pip install -e .[quantum-machines]` and it fails. It is ok as theintended way is working but I find it curious

I think `pip` by itself doesn't respect the `override-dependencies` field.

PR #977, opened 2025-08-27
qhc-1071 add calibration file to qbloxdraw

flavie-lebars:
QbloxDraw now supports passing a calibration file as an argument when plotting from both the platform and qprogram.


jordivallsq (approved):
Nice one! Thank you for the quick fix!

PR #978, opened 2025-08-27
qhc 1073 add phase to iqpair drag and prepare for future improvements

jordivallsq:
As Fabio suggested and to make measurements more straightforward iq pair needs to include a phase rotation. As this is a complex subject, for now we can implement simply a 90 degrees rotation and set a not implemented error for the rest.


fedonman:
I just saw it was created a long time ago, sorry I thought it was something urgent brought up.

It's true that Rmw is digital-only. But pulse programming already supports IQ rotation: Either with the `set_phase()` (recommended way) or manually rotating the IQWaveform (we can add a utility method if it's more user friendly)

For the calibration file we can edit it to include rotation, I have no issues with that. But it should be a new data structure specifically for the calibration file and not change the IQWaveform. E.g. it could be
```
class FrameWaveform:
   waveform: IQWaveform
   phase: float
```
or something similar. But again it would be more "correct" to implement with `SetPhase() + Play()` operations. The issue with rotating the IQWaveform is that many envelopes are created and the memory can overflow, whereas by using `SetPhase()` you could have unlimited rotations using the same envelope.

jordivallsq:
It seems like a good solution but I need to check the best way to implement something similar. I agree that the right way is to implement similar with set phase but my idea was to modify the envelope before reaching qblox compilation, this way we avoid unnecessary waveforms

PR #980, opened 2025-08-29
QUICKFIX check shape type on stream_array

jordivallsq:
Now the code checks if the shape of the streamarray is an int, if it is it transforms it to list.


flavie-lebars (approved):
only missing the changelog but looks good to me!

PR #981, opened 2025-08-29
Qhc 1072 implement real time predistorsions inside qblox drivers

flavie-lebars:


src/qililab/platform/platform.py
```diff
@@ -552,12 +568,65 @@ def _data_draw(self):
...
+        for pair in self.qblox_alias_module:
+            module_alias, output_id = next(iter(pair.items()))
+            qblox_instrument = self.instruments.get_instrument(module_alias)
+            for filter in qblox_instrument.filters:
+                if filter.output_id == output_id:
+                    if filter.fir_state in {DistortionState.ENABLED, DistortionState.DELAY_COMP}:
+                        qblox_active_filter = True
+        return qblox_active_filter
+
+    def _get_qblox_active_filter_exponential(self):
```
jordivallsq, line 587: Is this run every time a runcard is executed? What happens if there is no qblox? (same for _get_qblox_alias_module)

src/qililab/platform/platform.py
```diff
@@ -591,7 +661,71 @@ def set_parameter(
...
+    def _update_qblox_filter_state_exponential(self):
+        """Updates the exponential state as needed.
+            If any exponential idx is active ("enabled" or "delay_comp"), the exponential state of all other outputs will be updated to "delay_comp".
+            This ensures that there are no delays between outputs that have a filter "enabled" or "delay_comp" and outputs with bypassed filters.
+        """
+        if self.qblox_active_filter_exponential:
+            for pair in self.qblox_alias_module:
+                alias, output_id = next(iter(pair.items()))
+                for expo_idx in self.qblox_active_filter_exponential:
+                    parameter = getattr(Parameter, f"EXPONENTIAL_STATE_{expo_idx}")
```
jordivallsq, line 698: Does this accept integers in `exponential_amplitude`, `exponential_time_constant`... instead of lists? In  the changelog you added an example with ints but I cannot see where this is checked

src/qililab/instruments/qblox/qblox_module.py
```diff
@@ -314,6 +461,105 @@ def _set_out_offset(self, output: int, value: float | str | bool):
...
+            exponential_idx (int): index of the exponential filter
+            value (float): value to update
+        """
+        if value is not None:
+            filter_exp_amplitude = self.get_filter(output_id).exponential_amplitude
+            filter_exp_amplitude = cast("list[float]", filter_exp_amplitude)
+            filter_exp_amplitude[exponential_idx] = float(value)
+            if self.is_device_active():
+                getattr(self.device, f"out{output_id}_exp{exponential_idx}_amplitude")(float(value))
+                logger.warning("Distortion filter settings can cause transient effects.")
```
jordivallsq, line 518: At the end are we changing this whit the device active?

src/qililab/instruments/keithley/keithley_2600.py
```diff
@@ -44,7 +44,13 @@ class Keithley2600Settings(Instrument.InstrumentSettings):
     settings: Keithley2600Settings
     device: Keithley2600Driver
 
-    def set_parameter(self, parameter: Parameter, value: ParameterValue, channel_id: ChannelID | None = None):
+    def set_parameter(
+        self,
+        parameter: Parameter,
+        value: ParameterValue,
+        channel_id: ChannelID | None = None,
+        output_id: OutputID | None = None,
```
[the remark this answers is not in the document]
flavie-lebars: I would find it more confusing because channel_id is for the sequencer, and output_id is by output. But we can discuss it at the sprint planning/daily

src/qililab/platform/platform.py
flavie-lebars: the function checks at this line `if isinstance(instrument, QbloxModule):`

fedonman (commented):
Here are some of my thoughts:

1. Ideally, we should be able to get/set parameters based on all three identifiers: `channel_id`, `output_id`, `input_id` depending on the capabilities of the instrument. This is something that was tried in the Pydantic Refactoring PR. So, in essence, I agree with adding the `output_id` to the methods.
2. However, buses are connected to channels, not outputs. So, it doesn't make sense to add the `output_id` there, at least with no hard-check. For example, let's say `drive_q0_bus` is connected to sequencer 0 and sequencer 0 is connected to output 0. What if I do `platform.set_parameter("drive_q0_bus", ..., output_id=3)`? I will set an output not belonging to the bus. For the channels, there is a check the verifies that `channel_id` is the one that the bus is connected to, but how will we do it for outputs? It needs to be done though.
3. There are some inconsistencies if we merge this PR as it is:
    a. We introduce five  new parameters `EXPONENTIAL_AMPLITUDE`, `EXPONENTIAL_TIME_CONSTANT`, `EXPONENTIAL_STATE`, `FIR_COEFF`, `FIR_STATE`. However some of them are lists (`FIR_COEFF`, `FIR_STATE`) and some of them are split into four distinct Parameters, like `EXPONENTIAL_AMPLITUDE_0`, `EXPONENTIAL_AMPLITUDE_1`, `EXPONENTIAL_AMPLITUDE_2` and `EXPONENTIAL_AMPLITUDE_3`. Why not have them all defined as lists?

    b. If we add `output_id` to the get/set parameters it makes sense to be able to set other parameters that refer to outputs. Instead of having different parameters like `OFFSET_OUT0`, ..., `OFFSET_OUT3`, or `OUT0_LO_FREQ`, `OUT1_LO_FREQ` we should be able to target them as one Parameter like `Offset` or `LO_FREQUENCY` by the `output_id`.

     So ideally, I think that if we go with the `output_id` solution, the runcard should look like this:
   ```
    outputs:
        identifier: 0
        offset: 0.5  # or [0.5, 0.6] if it is RF with I/Q
        filter:
          exponential_amplitude: [0.8, -1]
          exponential_time_constant: [6, 8]
          exponential_state: [True, True, False]
          fir_coeff: [0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8, 0.8]
          fir_state: True
   ```
   
   Of course, this change might be too big and out of scope of this PR, but we should target it IMO.

flavie-lebars:
Hi Vyron, thanks!
I will reply point by point when needed:
2. I did not think of this, thanks, I will look into implementing this check
3. a. I split it like this because there are four exponential filters, so when doing for example EXPONENTIAL_AMPLITUDE_1 of output 0 it refers to element 1 of the list of exponential amplitude (-1 in the runcard example you gave). Let me know if you find a better way to do this. If I were to only have  EXPONENTIAL_AMPLITUDE the user would need to give all the list of exponential when doing a set_parameter.
3.b. I initially chose not to go with this approach as it would require to modify all existing runcards (since I kept the filters section fully optional) but I agree it is better this way, I will change this.

fedonman:
> @fedonman , thanks! I will reply point by point when needed: 2. I did not think of this, thanks, I will look into implementing this check 3. a. I split it like this because there are four exponential filters, so when doing for example EXPONENTIAL_AMPLITUDE_1 of output 0 it refers to element 1 of the list of exponential amplitude (-1 in the runcard example you gave). Let me know if you find a better way to do this. If I were to only have EXPONENTIAL_AMPLITUDE the user would need to give all the list of exponential when doing a set_parameter. 3.b. I initially chose not to go with this approach as it would require to modify all existing runcards (since I kept the filters section fully optional) but I agree it is better this way, I will change this.

Don't implement 3b. I agree it is way out of scope of the PR. We should discuss when/how to do it in the future.

PR #982, opened 2025-09-02
(BUG) QbloxDraw acquire_weighted to take weights as integration duration

flavie-lebars:
Previously, when plotting from the platform, the integration length was incorrectly taken from the runcard parameter. However, since Qililab currently only implements acquire_weighted, the integration length should instead be determined by the duration of the weight.
This has been corrected and now the behaviour of the acquire is the same when plotting from the platform or the qprogram
The integration length is defined as the duration of the acquire, not the weight, because Qililab ensures they are always equal.
As a result, two acquires cannot overlap in Qililab. However, in QbloxDraw’s logic, interruptions remain possible, similar to Play.


jordivallsq (approved):
Great work!! Now the qprogram draw with and without runcard share the correct integration_length.

PR #985, opened 2025-09-08
Quickfix to replace FlatTop variable sigma by smooth_duration

jordivallsq:
From petition from the measurement team I changed the name sigma to smooth_duration as matehmatically speacking that variable name had no sense. Now the variable refers to the duration of the smoothing in ns (or us, depending on the instrument).


fedonman (approved):
Image

PR #986, opened 2025-09-09
[QHC 1076] Added external load_by_id for slurm

jordivallsq:
Added `ql.load_by_id(id)` in qililab, This function allows to retrieve the data path of a measurement with the given id without creating a `DatabaseManager` instance. This function is intended to be used inside notebooks using slurm as `DatabaseManager` cannot be submitted.


flavie-lebars (approved):
looks good!!

PR #988, opened 2025-09-12
Qhc 1083 implement skeleton

fedonman:
* Removed all references to **`Qibo`**.

* Removed the **`pulse`** module.

* Restructured the **`digital`** module:

  * Introduced our own **`CircuitTranspiler`** and **`CircuitToQProgramCompiler`**.
  * **`CircuitTranspiler`** is now responsible for decomposing circuits into the native gate set, handling logical and physical layouts, and performing optimizations.
  * It operates as a linear pipeline (`list[CircuitTranspilerPass]`) designed to replace and improve upon the previous **Router**, **Placer**, and **Optimizer** that referenced Qibo implementations.
  * Each transpiler pass is required to have a concrete and well-defined purpose.

* Implemented the following transpiler passes:

  * `CancelIdentityPairsPass`
  * `CircuitToCanonicalBasisPass`
  * `FuseSingleQubitGatesPass`
  * `SabreLayoutPass`
  * `SabreSwapPass`
  * `CanonicalBasisToNativeSetPass`

* Refactored the **IQ waveform class hierarchy**:

  * Added an abstract parent class **`IQWaveform`**, from which **`IQPair`** now inherits.
  * Removed the `DRAG` class method from **`IQPair`** and replaced it with a dedicated **`IQDrag`** class.
  * This unification ensures all IQ waveforms are handled consistently and can be correctly serialized.

### DIFF
--- a/src/qililab/platform/platform.py
+++ b/src/qililab/platform/platform.py
@@ -1430,61 +1430,6 @@ class Platform:
 
         return results
 
-    # @staticmethod
-    # def _order_result(result: Result, circuit: Circuit, final_layout: list[int] | None) -> Result:
-    #     """Order the results of the execution as they are ordered in the input circuit.
-
-    #     Finds the absolute order of each measurement for each qubit and its corresponding key in the
-    #     same format as in qblox's aqcuisitions dictionary (#qubit, #qubit_measurement).
-
-    #     Then it orders results in the same measurement order as the one in circuit.queue.
-
-    #     Args:
-    #         result (Result): Result obtained from the execution
-    #         circuit (Circuit): Qibo circuit being executed
-    #         final_layouts (list[int], optional): Final layout of the original logical qubits in the physical circuit:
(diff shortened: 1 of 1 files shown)


src/qililab/platform/platform.py
```diff
@@ -1435,111 +1424,73 @@ def execute(
...
-            physical_qubit = qblox_result["qubit"]
-            original_logical_qubit = final_layout[physical_qubit] if final_layout else physical_qubit
-
-            # TODO:Check this is correct, or how it works with multiple measurements per qubit:
-            original_measure_num = order[original_logical_qubit, measurement]
-            results[original_measure_num] = qblox_result
+        return results
 
-        return QbloxResult(integration_lengths=result.integration_lengths, qblox_raw_results=results)
+    # @staticmethod
```
DZPM, line 1433: Why are we keeping all this commented code?

fedonman: Removed.

PR #989, opened 2025-09-16
[QUICKFIX] Counts documentation update

jordivallsq:
Fixed documentation for results `counts`, now it warns the user that instead of `num_avg` they must use `num_bins`.


src/qililab/result/qblox_results/qblox_bins_acquisitions.py
```diff
@@ -62,7 +62,9 @@ def counts(self) -> Counts:
         counts_object = Counts(n_qubits=len(self.bins))
         for seq in self.bins:
             if any(bin_avg > 1 for bin_avg in seq.avg_cnt):
-                raise ValueError("Counts cannot be used when using a hardware average higher than 1.")
+                raise ValueError(
```
DZPM, line 65: Shouldn't this trigger a test?

jordivallsq: The code is for the old version of qililab. If we want to add tests we can but this code should be deprecated.

PR #996, opened 2025-09-25
[QHC-1091] Modify `execute_qprograms_parallel` to take multiple bus mappings and Calibrations

jjmartinezQT:
`platform.execute_qprograms_parallel()` now takes a list of bus mappings to allow one bus mapping per qprogram.
Parameters for the function have now the same syntax and behaviour:
- bus_mapping (ist[dict[str, str] | None] | dict[str, str], optional). It can be one of the following:
    - A list of dictionaries mapping the buses in the :class:`.QProgram` (keys )to the buses in the platform (values). In this case, each bus mapping gets assigned to the :class:`.QProgram` in the same index of the list of qprograms passed as first parameter.
    - A single dictionary mapping the buses in the :class:`.QProgram` (keys )to the buses in the platform (values). In this case the same bus mapping is used for each one of the qprograms.
    - None, in this case there is not a bus mapping between :class:`.QProgram` (keys )to the buses in the platform (values) and the buses are as defined in each qprogram.
    It is useful for mapping a generic :class:`.QProgram` to a specific experiment.
    Defaults to None.
- calibrations (list[Calibration], Calibration, optional). Contains information of previously calibrated values, like waveforms, weights and crosstalk matrix. It can be one of the following:
    - A list of :class:`.Calibration` instances, one per :class:`.QProgram` instance in the qprograms parameter.
    - A single instance of :class:`.Calibration`, in this case the same `.Calibration` instance gets associated to all qprograms.
    None. In this case no `.Calibration` instance is used.
    - Defaults to None.

### DIFF
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -19,6 +19,7 @@ calibrations (list[Calibration], Calibration, optional). Contains information of
     Defaults to None.
 [#996](https://github.com/qilimanjaro-tech/qililab/pull/996)
 
+
 - `%% submit_job`: Added support for `sbatch --chdir` via a new `-c/--chdir` option that is propagated through `slurm_additional_parameters` and also enforced inside the job (`os.chdir(...)`) so it works with `-e local`. Made `--output` mandatory and hardened the output‑assignment check to recognize `Assign`, `AugAssign`, `AnnAssign`, walrus (`NamedExpr`), and tuple targets. Shipment of the notebook namespace is now safer: only picklable values (via `cloudpickle`) are sent, with common pitfalls (modules, loggers, private `_` names, IPython internals) excluded. `--low-priority` is a boolean flag mapping to a sane Slurm `nice=10000`. Paths are handled with `pathlib` plus `expanduser/expandvars`, the logs directory is created if missing, and imports are harvested conservatively from history (one‑line `import`/`from`, excluding `from __future__`). Parameter assembly only includes Slurm extras when provided, and the submitted function compiles the code string internally while accepting the output name and optional workdir. The job object is written to both `local_ns` and the global `user_ns` for IPython robustness. Log cleanup was rewritten to be cross‑platform and resilient: artifacts are grouped by numeric job‑ID prefix, non‑conforming entries are removed, and only the newest `num_files_to_keep` job groups are retained.
   [#994](https://github.com/qilimanjaro-tech/qililab/pull/994)
 


src/qililab/platform/platform.py
```diff
@@ -24,7 +24,7 @@
 from contextlib import contextmanager
 from copy import deepcopy
 from dataclasses import asdict
-from typing import TYPE_CHECKING, Any, Callable, cast
+from typing import TYPE_CHECKING, Any, Callable, Optional, cast
```
fedonman, line 27: nit: avoid using `Optional` it's a bit "old", `type | None` is preferred for 3.10+. Not important, just commenting it for having the same code style.

src/qililab/platform/platform.py
```diff
@@ -1227,42 +1227,76 @@ def execute_qprogram(
...
+
+        return {mapping.get(b, b) for b in qp_buses}
+
     def execute_qprograms_parallel(
         self,
         qprograms: list[QProgram],
-        bus_mapping: dict[str, str] | None = None,
-        calibration: Calibration | None = None,
+        bus_mappings: Optional[list[dict[str, str]]] = None,
+        calibration: Optional[Calibration] = None,
```
fedonman, line 1257: Since we allow for a list of `bus_mapping` then why don't we allow for a list of `Calibration` as well?

src/qililab/platform/platform.py
```diff
@@ -1227,42 +1227,76 @@ def execute_qprogram(
...
+            return set(qp_buses)
+
+        return {mapping.get(b, b) for b in qp_buses}
+
     def execute_qprograms_parallel(
         self,
         qprograms: list[QProgram],
-        bus_mapping: dict[str, str] | None = None,
-        calibration: Calibration | None = None,
+        bus_mappings: Optional[list[dict[str, str]]] = None,
```
fedonman, line 1256: What if I want to use the same `bus_mapping` for all QPrograms? Maybe we should make the type as `dict[str, str] | list[dict[str, str] | None`, so the user can:

1. Pass None -> don't use mappings
2. Pass single mapping -> use it for all QPrograms
3. Pass list of mappings -> Check that sizes are equal and use them 1-1 with the QPrograms

fedonman: This is also backwards-compatible with the current behaviour

src/qililab/platform/platform.py
jjmartinezQT: Sure thing.

src/qililab/platform/platform.py
jjmartinezQT: I had it like that initially but then had second thoughts, but will revert to mimic what you are proposing.

jjmartinezQT: Addressed now.

src/qililab/platform/platform.py
jjmartinezQT: Addressed.

src/qililab/platform/platform.py
jjmartinezQT: Extended to calibrations now.

src/qililab/platform/platform.py
```diff
@@ -1227,42 +1227,120 @@ def execute_qprogram(
...
-        It loads each qprogram into a different sequencer and uses the multiplexing capabilities of QBlox to run all sequencers at the same time.
-
-        **The execution can be done for (buses associated to) Qblox only. And it can only be done for qprograms that do not share buses.**
+            It loads each qprogram into a different sequencer and uses the multiplexing capabilities of QBlox to run all sequencers at the same time.
+
+            **The execution can be done for (buses associated to) Qblox only. And it can only be done for qprograms that do not share buses.**
+
+            Args:
+                qprograms (list[QProgram]): A list of the :class:`.QProgram` to execute.
+                bus_mapping (Optional[Sequence[Optional[dict[str, str]]]] | Optional[list[dict[str, str]]]): A list of dictionaries mapping the buses in the :class:`.QProgram` (keys )to the buses in the platform (values).
```
fedonman, line 1307: Remove `Optional`, added as `, optional)` like the others.

fedonman, line 1307: I think it would be useful to explain the three cases here, since this is the method that user will see.

docs/releases/changelog-dev.md
```diff
@@ -4,6 +4,9 @@
 
 ### Improvements
 
+- `platform.execute_qprograms_parallel()` now takes a list of bus mappings to allow one bus mapping per qprogram.
+[#996](https://github.com/qilimanjaro-tech/qililab/pull/996)
```
fedonman, line 8: Also, explain the three cases here.

fedonman (commented):
Minor things and good to go! 💯

src/qililab/platform/platform.py
jjmartinezQT: yes, makes sense.

docs/releases/changelog-dev.md
jjmartinezQT: addressed

fedonman (approved):
🚀 🚀

PR #997, opened 2025-09-25
Fix ExperimentExecutor with databases test creating a file

jordivallsq:
Due to a missing mock on the H5py.File a file was created everytime the tests were run. This has been fixed.

In addition, the assertions for the database have been modified following the rest of the tests code.


jordivallsq:
@fedonman Do we need to add a changelog entry to a feature fix that has not been released? I understand to put the PR to the fixed feature entry but if we create another entry the users that review the changelog will see the addition of a new feature and some comments behind a fix for a bug they've neever seen

DZPM (approved):

PR #998, opened 2025-09-26
Qhc 1109 qblox compiler to loop over bins and not acquisition index

flavie-lebars:


src/qililab/result/qprogram/qblox_measurement_result.py
```diff
@@ -74,3 +82,49 @@ def threshold(self) -> np.ndarray:
...
+
+                avg_cnt = results_unintertwined.raw_measurement_data["bins"]["avg_cnt"]
+                results_unintertwined.raw_measurement_data["bins"]["avg_cnt"] = avg_cnt[result::self.intertwined]
+
+                results_unintertwined_list.append(results_unintertwined)
+
+        else:
+            results_unintertwined_list = [self]
+
+        return results_unintertwined_list
```
fedonman, line 130: This should happen as post-processing of the results in the `platform._execute_qblox_compilation_output` method. The `QbloxCompilationOutput` that is input to that method has the `AcquisitionData` info, in which you have already added the `interwined` property, so the method has all the information it needs.

We want this to be transparent from the user, it is our way of internally storing/handling results and is relevant only when executing qprograms. The returned `QbloxResults` of the `platform._execute_qblox_compilation_output` should be exactly the same as before.

So I suggest:
a) Remove the `interwined` property.
b) Move this logic as post-processing of the results in the `platform._execute_qblox_compilation_output` method. Of course it will run inside a loop, for all `QbloxMeasurementResult` in `QProgramResult`.

Side note: I think it's easier and faster to create a new `QbloxMeasurementResult` each time, than deep copying.

fedonman (changes requested):
Super nice idea and implementation! Left one major comment I would like to get addressed before merging.

src/qililab/result/qprogram/qblox_measurement_result.py
flavie-lebars: thanks vyron! I have modified the code and tested it in hardware

fedonman (approved):
Great work! 💯 💯 Everything seems correct, but I would suggest doing a hardware test before merging!

PR #999, opened 2025-09-29
[QHC-1113] Fix Reshaping of `MeasurementResults` within `ExperimentResults`

fedonman:
- Fixed a bug in the reshaping of `MeasurementResults` within the `ExperimentResults`.
- Set database saving and live plotting to `False` by default during experiment execution.

### DIFF
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -4,6 +4,22 @@
 
 ### Improvements
 
+- `platform.execute_qprograms_parallel()` now takes a list of bus mappings to allow one bus mapping per qprogram.
+Parameters for the function have now the same syntax and behaviour:
+bus_mapping (ist[dict[str, str] | None] | dict[str, str], optional). It can be one of the following:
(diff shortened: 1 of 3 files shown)


src/qililab/qprogram/experiment_executor.py
```diff
@@ -466,7 +467,7 @@ def store_results(qprogram_results: QProgramResults, qprogram_index: int):
             # Determine the index based on current loop indices and store the results in the ExperimentResultsWriter
             for measurement_index, measurement_result in enumerate(qprogram_results.timeline):
                 indices = (qprogram_index, measurement_index, *tuple(index for _, index in self.loop_indices.items()))
-                self._results_writer[indices] = measurement_result.array.T  # type: ignore
+                self._results_writer[indices] = np.moveaxis(measurement_result.array, 0, -1)
```
elygoner, line 470: What is this for?

jordivallsq, line 470: I this the same treatment as we do in `qprogram_executor` results? I cannot find any other line with `moveaxis`

jordivallsq: Is is equivalent of this reshape inside `QbloxMeasurementResult`?
```
array = array.reshape((2, *self.shape))
```

jordivallsq (commented):
I have a few questions for the actual reshaping, aside for that it looks great!

PR #1001, opened 2025-09-29
Qhc 1110 numeric helpers fix verify and write tests

elygoner:
I fixed the _mat_U3 and the _zyz_from_unitary functions so that they work as intended, and I created tests for it.


GuillermoAbadLopez (commented):
We will leave the changelog for the bigger PR so go ahead 🙌🏽

PR #1008, opened 2025-09-30
Qhc 1085 fusesinglequbitgatespass verify and write tests

elygoner:


src/qililab/digital/circuit_transpiler_passes/fuse_single_qubit_gates_pass.py
```diff
@@ -102,7 +108,7 @@ def flush_touches(qubits: Tuple[int, ...]):
                 apply(q, U)
 
             elif isinstance(g, CZ):
-                flush_touches((g.control_qubits[0], g.target_qubits[0]))
+                flush_touches((*g.control_qubits, *g.target_qubits))
```
GuillermoAbadLopez, line 111: Hmmm, seems clearer now; both ways should be equivalent. 

Although, thanks to your change I am seeing a possible bug with this, given the current qilisdk `CZ` implementation, since a user can do something like:
``` a = CZ((2, 3), 0) ```
And therefore, we would have a wrong CZ gate, with multiple control qubits, which the transpiler would treat as a normal one. Although this would also be a problem with the original implementation.

So I will raise this in `qilisdk` repo, so we add checkers for the CZ at least, and both these implementations should be equivalent 👌

**Review by GuillermoAbadLopez — COMMENTED** [review]
Small review about the source code changes

File: src/qililab/digital/circuit_transpiler_passes/fuse_single_qubit_gates_pass.py (hunk 1 above)
Comment (elygoner) [reply src/qililab/digital/circuit_transpiler_passes/fuse_single_qubit_gates_pass.py:111]: In the previous pass (circuit to canonical) an error is razed is a control gate is working on multiple *target* qubits;
image

 and changes any gates^with more than one control qubit with:
image

Comment (elygoner) [reply src/qililab/digital/circuit_transpiler_passes/fuse_single_qubit_gates_pass.py:111]: He can add an error in here too, if you'd like, since you can manualy run the passes.

File: src/qililab/digital/circuit_transpiler_passes/fuse_single_qubit_gates_pass.py [hunk 2]
```diff
@@ -69,16 +69,22 @@ def flush(q: int):
             theta = _wrap_angle(theta)
             phi = _wrap_angle(phi)
             gamma = _wrap_angle(gamma)
-            if abs(theta) < 1e-9:
+            if _is_close_mod_2pi(theta, 0.0):
```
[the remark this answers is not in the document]
elygoner: We'll leave it for future refinement.

PR #1010, opened 2025-10-01
qhc-1111-optimise-q1asm-length

flavie-lebars:
[BUG] -  When dealing with real time classical time in the qblox draw the real duration was put instead of the wait duration

### DIFF
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -176,7 +176,7 @@ The data automatically selects between the local or shared domains depending on 
 
 ### Bug fixes
 
-- Qblox Draw- When dealing with real time and classical time, the real duration was put instead of the wait duration. Note: do not include in the next release.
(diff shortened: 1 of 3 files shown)


docs/releases/changelog-dev.md
```diff
@@ -176,6 +176,8 @@ The data automatically selects between the local or shared domains depending on
 
 ### Bug fixes
 
+- Qblox Draw- When dealing with real time and classical time, the real duration was put instead of the wait duration. Note: do not include in the next release.
```
jjmartinezQT, line 179:
```suggestion
- Qblox Draw- When dealing with real time and classical time, the real duration was put instead of the wait duration. Note: do not include this comment in the next release changelog as the bug was not in the previous release.
```

jjmartinezQT: adding this small suggestion as I am very mindless and I know I will think it is the fix that should not be included and not the changelog entry :D

PR #1013, opened 2025-10-02
Qhc 1084 circuittocanonicalbasispass verify and write tests

elygoner:
**changelog**
*Changed CircuitToCanonicalBasisPass with updated logic
*added tests for it


fedonman (changes requested):
We'll have to check the two implementations more carefully (original one and the one in this PR) because there are many differences in the emitted circuit. For example,

Previous implementation:

Image

This PR:

Image

I asked GPT to compare the two implementations and it favors the original one. Check https://chatgpt.com/s/t_68e3844a57308191a4d2fa1de392d57d and https://chatgpt.com/s/t_68e388c64638819198d9babac2106e67

So let's review them more carefully, Theory can help in this too. Personally, I would recommend adding tests for the original.

PR #1020, opened 2025-10-15
Qhc 1132 driver for the rswu sp16tr

jjmartinezQT:
This PR brings a new driver for the RSWU-SP16TR instrument.


docs/releases/changelog-dev.md
```diff
@@ -4,6 +4,365 @@
 
 ### Improvements
 
+- Implemented a new driver for the Becker Nachrichtentechnik RSWU-SP16TR
```
jordivallsq: Check the changelog and add only the entry from this PR (leave the titles).
This file gets whiped clean after every release.

jordivallsq (changes requested):
I left a couple of comments

jordivallsq (approved):
Good Work! I understood that you tested in hardware and it works as intended.
Regarding ruff comments, Ruff is complaining because it is trying to do do autoformatting and is not able to. You can install the ruff autoformat as a VSCode extension and wait for it to reformat all the files you update.

PR #1030, opened 2025-11-03
[QHC-1124] Crosstalk compensation in qblox / qdac compiler

jordivallsq:


src/qililab/platform/platform.py
```diff
@@ -1343,6 +1349,14 @@ def execute_compilation_output(
         output: QProgramCompilationOutput,
         debug: bool = False,
     ):
+        if isinstance(output.qdac, QdacCompilationOutput):
+            qdac_buses = [
```
flavie-lebars: can't we store these somewhere? because those lines are repeated several times in the code, unless they change between the compilation and the execution

src/qililab/qprogram/crosstalk_matrix.py
```diff
@@ -244,10 +252,11 @@ def set_crosstalk_from_bias(self, crosstalk: CrosstalkMatrix, bias_vector: dict[
         if not self.bias_vector:
             self.bias_vector = self.flux_vector.copy()
 
+        # Add duration logic!!!!!
```
flavie-lebars: did you forget this?

src/qililab/qprogram/qdac_compiler.py
```diff
@@ -38,6 +41,9 @@
 )
 from qililab.qprogram.qprogram import QProgram
 from qililab.typings.enums import Parameter
+from qililab.waveforms.arbitrary import Arbitrary
```
flavie-lebars:
```suggestion
from qililab.waveforms import Arbitrary, IQPair, Waveform
```

src/qililab/qprogram/qdac_compiler.py
```diff
@@ -93,24 +99,30 @@ def __init__(self) -> None:
...
         self,
         qprogram: QProgram,
         qdac: QDevilQDac2,
         qdac_buses: list["Bus"],
+        qdac_offsets: list[float],
         bus_mapping: dict[str, str] | None = None,
         calibration: Calibration | None = None,
+        crosstalk: CrosstalkMatrix | None = None,
     ) -> QdacCompilationOutput:
         """Compile QProgram to qpysequence.Sequence
```
flavie-lebars: why qpysequence if we are in the qdac compiler?

src/qililab/qprogram/qdac_compiler.py
```diff
@@ -286,9 +412,10 @@ def _handle_play(self, element: Play):
             self._qdac.upload_voltage_list(
                 waveform=waveform,
                 channel_id=self._channels[element.bus],
-                dwell_us=convert(element.dwell),
+                dwell_us=element.dwell,
```
flavie-lebars: why is the dwell conversion no longer needed?

src/qililab/qprogram/qprogram.py
```diff
@@ -63,6 +69,67 @@ def __init__(
         self.quantum_machines = quantum_machines
 
 
+class CrosstalkElements:
+    """_summary_"""
```
flavie-lebars: ??

flavie-lebars (changes requested):
Great job!
This is an initial review from my side, I will do a following one. The main concern I have is I do not understand why the transpilation to a new qprogram is done inside the qdac compiler and not qprogram

src/qililab/platform/platform.py
```diff
@@ -1227,6 +1228,7 @@ def compile_qprogram(
         qdac_buses = [
             bus for bus in buses if any(isinstance(instrument, QDevilQDac2) for instrument in bus.instruments)
         ]
+        qdac_offsets = [float(bus.get_parameter(Parameter.VOLTAGE)) for bus in qdac_buses]
```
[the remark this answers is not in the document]
jordivallsq: Since compiled_qdac can be used in both Qprogram and QM (I know that is not very relevant but it was not difficult either way) I moved all the functions to a common place to avoid repeating info and defining everything related to the qda to the same place

src/qililab/qprogram/qprogram.py
```diff
@@ -375,6 +442,377 @@ def traverse(block: Block):
...
+            for i, element in enumerate(block.elements):
+                if isinstance(element, (Play, SetOffset, SetGain)) and element.bus in crosstalk.matrix.keys():
+                    crosstalk_elements.check_flux_vector(element)
+
+                    crosstalk_elements.flux_vector[str(element.__class__)] = handle_flux_vector(
+                        flux_vector=crosstalk_elements.flux_vector[str(element.__class__)],
+                        element=element,
+                    )
+                    crosstalk_elements.add_element(element, i)
+                if isinstance(element, (Wait, Acquire, Measure, MeasureReset)) or (
```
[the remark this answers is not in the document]
jordivallsq: Added as an exception then, as I cannot think of the necessity of combining those two features.
But as a note, this interaction here is just restarting the flux vector values to not carry unnecessary information after waiting

src/qililab/qprogram/qprogram.py
```diff
@@ -63,6 +69,67 @@ def __init__(
...
+            element (Play | SetOffset | SetGain): _description_
+        """
+        operation = str(element.__class__)
+        if operation not in self.flux_vector_bus.keys() or element.bus in self.flux_vector_bus[operation]:
+            self.restart_flux_vector(operation)
+
+    def restart_flux_vector(self, operation: str | None = None):
+        """_summary_
+
+        Args:
```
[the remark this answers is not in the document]
jordivallsq: added

src/qililab/instruments/qdevil/qdevil_qdac2.py
```diff
@@ -172,6 +172,7 @@ def upload_voltage_list(
         dwell_us: int = 1,
         sync_delay_s: float = 0,
         repetitions: int = 1,
+        stepped: bool = False,
```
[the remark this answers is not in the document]
jordivallsq: added

jordivallsq (commented):
I have replied and applied the suggested changes and comments.
Thank you very much for the review.

src/qililab/platform/platform.py
```diff
@@ -1508,7 +1522,7 @@ def _execute_quantum_machines_compilation_output(
             compiled_program_id = cluster.compile(program=qua)
 
             if output.qdac:
-                if output.qdac.trigger_position == "front":
+                if output.qdac.trigger_position == "back":
```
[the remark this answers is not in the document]
jordivallsq: removed it

docs/releases/changelog-dev.md
```diff
@@ -2,6 +2,82 @@
 
 ### New features since last release
 
+- Implemented QBlox and QDAC-II automatic crosstalk compensation for `Qprogram`. The compiler automatically detects if there is a crosstalk matrix inside platform and implements the crosstalk for any bus inside the `Crosstalk` class. To do so, either use `platform.set_crosstalk(crosstalk)` or define a crosstalk inside `calibration` and use it through `execute_qprogram(..., calibration)`.
```
flavie-lebars, line 5: For completeness, I would add the third way you can set the crosstalk, which is `execute_qprogram(..., crosstalk)`

docs/releases/changelog-dev.md
```diff
@@ -2,6 +2,82 @@
...
+    ```
+    ...
+
+    with qp_qdac.average(10):
+        qp_qdac.set_offset(bus="qdac_flux2", offset_path0=offset_2_compensated)
+        qp_qdac.set_offset(bus="qdac_flux1", offset_path0=offset_1_compensated)
+        with qp_qdac.for_loop(variable=freq, start=10e6, stop=100e6, step=10e6):
+
+            # QDAC TRIGGER GENERATION
+            qp_qdac.qdac.play(bus="qdac_flux2", waveform=flux_wf_crosstalk_compensated_2, dwell=2)
```
flavie-lebars, line 55: is it normal that the order gets reversed? In the inital qp, flux1 was first

docs/releases/changelog-dev.md
```diff
@@ -2,6 +2,82 @@
...
+                waveform=ql.IQPair(I=r_wf, Q=r_wf),
+                weights=ql.IQPair(I=r_wf, Q=r_wf),
+            )
+
+            qp_qdac.wait(bus="readout", duration=100)
+    ```
+
+  - For QBlox:
+    The `QProgram` structures affected by the crosstalk are `qp.set_offset(...)`, `qp.set_gain(...)` and `qp.play(...)` for flux buses. The behavior is similar to the `QdacCompiler` but accounting for the complexity of Qblox compilation.
+    With crosstalk, the qprogram structure will look the same as usual, but for each bus inside the crosstalk matrix, the machine will send a flux pulse / offset to compensate the crosstalk. The Qprogram is now able to sum variables so this implementation allows for a whole flux vs flux run in a single `Q1ASM`.
```
flavie-lebars: Is this the Variable Expression PR? this is not available yet.

tests/qprogram/test_qblox_compiler.py
```diff
@@ -784,6 +787,114 @@ def fixture_dynamic_wait_three_buses_static_static() -> QProgram:
...
+    with qp.for_loop(variable=offset, start=0, stop=0.1, step=0.01):
+        qp.set_offset(bus="flux1", offset_path0=offset)
+        qp.wait(bus="drive", duration=10)
+        qp.wait(bus="flux1", duration=10)
+        qp.wait(bus="flux2", duration=10)
+        qp.set_gain(bus="flux2", gain=0.05)
+        qp.play(bus="flux1", waveform=square_wf)
+        qp.play(bus="drive", waveform=square_iq)
+        qp.sync(["drive","readout"])
+        qp.measure(bus=f"readout", waveform=square_iq, weights=square_iq)
```
flavie-lebars: no need for f string here

tests/qprogram/test_qblox_compiler.py
```diff
@@ -784,6 +787,114 @@ def fixture_dynamic_wait_three_buses_static_static() -> QProgram:
...
+    with qp.for_loop(variable=gain, start=0, stop=0.1, step=0.01):
+        qp.set_offset(bus="flux1", offset_path0=0.05)
+        qp.wait(bus="drive", duration=10)
+        qp.wait(bus="flux1", duration=10)
+        qp.wait(bus="flux2", duration=10)
+        qp.set_gain(bus="flux2", gain=gain)
+        qp.play(bus="flux1", waveform=square_wf)
+        qp.play(bus="drive", waveform=square_iq)
+        qp.sync(["drive","readout"])
+        qp.measure(bus=f"readout", waveform=square_iq, weights=square_iq)
```
flavie-lebars: same her

tests/qprogram/test_qdac_compiler.py
```diff
@@ -133,18 +257,352 @@ def test_play_and_set_trigger(self, qdac: QDevilQDac2, flux1: Bus, flux2: Bus):
...
+            qprogram=qp, qdacs=[qdac], qdac_buses=[flux1, flux2], qdac_offsets=[0, 0], crosstalk=crosstalk
+        )
+
+        assert isinstance(output, QdacCompilationOutput)
+        assert compiler._qdacs == [qdac]
+
+        assert qdac_bus.upload_voltage_list.call_count == 2
+        assert qdac_bus.set_parameter.call_count == 0
+
+    def test_crosstalk_compensation(self, qdac: QDevilQDac2, flux1: Bus, flux2: Bus):
```
flavie-lebars: you have this function twice

tests/qprogram/test_qdac_compiler.py
```diff
@@ -133,18 +257,352 @@ def test_play_and_set_trigger(self, qdac: QDevilQDac2, flux1: Bus, flux2: Bus):
...
+
+        qp.qdac.play(bus="flux1", waveform=flux_wf, dwell=2)
+        qp.qdac.play(bus="flux2", waveform=flux_wf_wrong, dwell=2)
+        qp.set_trigger(bus="flux1", duration=10e-6, outputs=1, position="start")
+
+        compiler = QdacCompiler()
+
+        with pytest.raises(
+            ValueError,
+            match=f"qp.play elements must have the same size.",
```
flavie-lebars: no need for f string

tests/result/test_stream_results.py
```diff
@@ -281,7 +287,7 @@ def test_stream_array_instantiation_bus_map(self, stream_array_bus_map: StreamAr
     def test_stream_array_instantiation_qubit_idx(self, stream_array_qubit_idx: StreamArray):
         """Tests the instantiation of a StreamArray object with target and secondary indexes."""
         # Create mock for the file context
-        debug_q1asm = "Bus readout_q0:\nsetup:\n                wait_sync        4              \n                set_mrk          0              \n                upd_param        4              \n\nmain:\n                move             1, R0          \nsquare_0:\n                play             0, 1, 100      \n                loop             R0, @square_0  \n                wait             100            \n                set_mrk          0              \n                upd_param        4              \n                stop                            \n\n\n"
+        debug_q1asm = "Compilation of the debug has failed with the following error:\n'NoneType' object has no attribute 'instruments'"
```
flavie-lebars, line 342: is this change expected?

src/qililab/waveforms/flat_top.py
```diff
@@ -30,7 +30,7 @@ class FlatTop(Waveform):
     Args:
         duration (int): Duration of the pulse (ns).
         amplitude (float): Maximum amplitude of the pulse.
-        smooth_duration (float, optional): duration of the smoothing component in ns.
+        smooth_duration (float): duration of the smoothing component in ns.
```
flavie-lebars: the typing shows int

src/qililab/qprogram/qprogram.py
```diff
@@ -375,6 +452,469 @@ def traverse(block: Block):
 
         return copied_qprogram
 
+    def with_crosstalk(self, crosstalk: CrosstalkMatrix):
```
flavie-lebars: if this one is specific to qblox, could you rename it `with_crosstalk_qblox` please

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -315,10 +319,14 @@ def traverse(block: Block):
             self._qprogram = self._qprogram.with_bus_mapping(bus_mapping=bus_mapping)
         if calibration is not None:
             self._qprogram = self._qprogram.with_calibration(calibration=calibration)
+            if calibration.crosstalk_matrix and crosstalk is None:
```
flavie-lebars, line 322: what if crosstalk is True?
I assume the taken crosstalk will be the one defined at compilation but we should maybe raise a warning

src/qililab/instrument_controllers/instrument_controller.py
```diff
@@ -46,6 +54,7 @@ class InstrumentControllerSettings(Settings):
     alias: str
     connection: Connection
     modules: list[InstrumentReference]
+    reference_clock: str = field(default_factory=lambda: ReferenceClock("internal"))
```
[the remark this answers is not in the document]
flavie-lebars: ok sounds good, please make sure to create a Linear ticket so we don't forget

src/qililab/instrument_controllers/instrument_controller.py
```diff
@@ -168,15 +177,26 @@ def set_parameter(
...
-        """Updates the reset settings for the controller."""
+        """Gets the reset settings for the controller."""
         if parameter == Parameter.RESET:
             return self.settings.reset
-        raise ValueError("Reset is the only property that can be set for an Instrument Controller.")
+        """Gets the reference clock settings for the controller."""
+        if parameter == Parameter.REFERENCE_CLOCK:
+            return self.settings.reference_clock
+        raise ValueError(
+            "Reset and reference clock are the only properties that can be get for an Instrument Controller."
```
flavie-lebars: I would change get by retrieved

src/qililab/platform/platform.py
```diff
@@ -885,7 +892,7 @@ def set_bias_to_zero(self, bus_list: list[str] | None = None):
             if flux not in self.flux_vector.flux_vector.keys():
                 self.flux_vector[flux] = 0
             self.flux_vector.bias_vector[flux] = 0
-            self.flux_vector.set_crosstalk_from_bias(self.crosstalk)
+        self.flux_vector.set_crosstalk_from_bias(self.crosstalk)
```
flavie-lebars, line 895: was this a bug before? if yes can you please document it in the changelog

src/qililab/platform/platform.py
```diff
@@ -1213,8 +1220,15 @@ def execute_experiment(
         return executor.execute()
 
     def compile_qprogram(
-        self, qprogram: QProgram, bus_mapping: dict[str, str] | None = None, calibration: Calibration | None = None
+        self,
+        qprogram: QProgram,
+        bus_mapping: dict[str, str] | None = None,
+        calibration: Calibration | None = None,
+        crosstalk: bool = True,
```
flavie-lebars, line 1227: does it make sense to have the crosstalk default as True?

src/qililab/platform/platform.py
```diff
@@ -1213,8 +1220,15 @@ def execute_experiment(
...
     ) -> QProgramCompilationOutput:
+        if not crosstalk and calibration is not None and calibration.crosstalk_matrix is not None:
+            calibration.crosstalk_matrix = None
```
flavie-lebars, line 1230: I am confused by this line, this actually modifies the caller's object

src/qililab/platform/platform.py
```diff
@@ -1391,11 +1405,11 @@ def _execute_qblox_compilation_output(self, output: QProgramCompilationOutput, d
 
             # Execute sequences
             if output.qdac:
-                if output.qdac.trigger_position == "front":
+                if output.qdac.trigger_position == "back":
```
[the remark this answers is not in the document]
flavie-lebars: ok can we document it in the bug section of the changelog please

src/qililab/result/stream_results.py
```diff
@@ -202,14 +201,9 @@ def _get_debug(self):
...
-                compiled = self.platform.compile_qprogram(
-                    qprogram=self.qprogram, bus_mapping=self.bus_mapping, calibration=self.calibration
-                ).qblox
-            else:
-                qblox_compiler = QbloxCompiler()
-                compiled = qblox_compiler.compile(qprogram=self.qprogram, calibration=self.calibration)
-
+            compiled = self.platform.compile_qprogram(
+                qprogram=self.qprogram, bus_mapping=self.bus_mapping, calibration=self.calibration
+            ).qblox
```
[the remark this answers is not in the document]
flavie-lebars: does this mean we can only get the debug if the user provides both a bus_mapping and calibration? And any other case it returns the debug exception?

src/qililab/qprogram/crosstalk_matrix.py
```diff
@@ -229,35 +231,41 @@ def __setitem__(self, key: str, flux: float) -> None:
...
 
-    def set_crosstalk_from_bias(self, crosstalk: CrosstalkMatrix, bias_vector: dict[str, float] | None = None):
+    def set_crosstalk_from_bias(
+        self, crosstalk: CrosstalkMatrix, bias_vector: dict[str, float | list[float] | np.ndarray] | None = None
+    ) -> dict[str, float | list[float] | np.ndarray]:
         """Set the crosstalk compensation on the existing flux vector. This function does the matrix product to calculate the correct flux
 
         Args:
-            crosstalk (CrosstalkMatrix): _description_
+            crosstalk (CrosstalkMatrix): crosstalk matrix to be apply
```
flavie-lebars: applied

src/qililab/qprogram/qdac_compiler.py
flavie-lebars: pls add it as a bug in the changelog

src/qililab/instruments/qdevil/qdevil_qdac2.py
```diff
@@ -223,14 +266,59 @@ def set_in_internal_trigger(self, channel_id: ChannelID, trigger: str):
 
         if str(trigger) not in self._triggers.keys():
             raise ValueError(f"Trigger with name {trigger} not created.")
-        if channel_id not in self._cache_dc.keys():
+        if f"{self.device.name}_{channel_id}" not in self._cache_dc.keys():
```
flavie-lebars, line 269: isn't there a way ok keeping something similar to channel_id for all the bwlo conditionals? I don't really like having an f string here

src/qililab/instruments/qdevil/qdevil_qdac2.py
```diff
@@ -239,25 +327,34 @@ def set_end_marker_external_trigger(
...
             )
         if str(trigger) in self._triggers.keys():
             self.clear_trigger(trigger)
 
-        self._triggers[str(trigger)] = self._cache_dc[channel_id].allocate_trigger()
+        self._triggers[str(trigger)] = self._cache_dc[f"{self.device.name}_{channel_id}"].allocate_trigger()
 
         channel = self.device.channel(channel_id)
-        channel.write_channel(f"sour{'{0}'}:dc:mark:pend {self._triggers[str(trigger)].value}")
+        if not step:
```
flavie-lebars, line 344: this should not be done in qililabm, but we should call the driver. And if this specific function is not in qcodes we should add it. I understand that we need to merge fairly quickly, but id this is the case and the function needs to be created on qcodes, can you please create a Linear ticket for this to be done later. (I will only comment it here but this is valid for all SCPI commands in qililab -  we should avoid)

src/qililab/qprogram/qprogram.py
```diff
@@ -375,6 +452,469 @@ def traverse(block: Block):
...
+                        element=element,
+                    )
+                    crosstalk_elements.add_element(element, i)
+                if isinstance(element, (Wait, Acquire, Measure)) or (
+                    isinstance(element, Play) and element.bus not in crosstalk.matrix.keys()
+                ):
+                    crosstalk_elements.restart_flux_vector()
+
+                if isinstance(element, Block):
+                    crosstalk_elements.restart_flux_vector(str(Play.__class__))
```
flavie-lebars, line 503: This might be a bug here, here play is the class, not the instance, like you have done in the rest of the code

src/qililab/qprogram/qprogram.py
```diff
@@ -375,6 +452,469 @@ def traverse(block: Block):
...
+            elements: dict[int, Play],
+        ):
+            """
+            Modifies Play creating a compensated Arbitrary pulse.
+            For Square and FlatTop pulses uses the same type instead of an Arbitrary pulse.
+            """
+            play_elements = [elements[element] for element in element_group if isinstance(elements[element], Play)]
+            if len({(play.dwell, play.delay, play.repetitions, play.wait_time) for play in play_elements}) > 1:
+                raise ValueError("Play elements must be the same for the same play pulse.")
+            dwell, delay, repetitions, stepped, wait_time = (
```
flavie-lebars, line 709: what happens if different plays have different values of dwell, delay etc ..?

flavie-lebars (commented):
Great work Jordi!
I have a few comments:
- I don't think that compiler is the appropriate word, becasue the input and outputs are both qprograms.
- As this is very complex, when you have time later, could you pleae add documentation on Notion.
- qprogram should be split and there should be a class to apply the crosstalk compensation I believe. We can see with Vyron and the refactor of qprogram. 
- the crosstalk compensation for qblox should be split into different passes if feasible, it is very hard to read/review and test the chunks properly at the moment.

docs/releases/changelog-dev.md
jordivallsq: Yes it was, I will remove that part then.

src/qililab/platform/platform.py
jordivallsq: Added to the changelog

src/qililab/instruments/qdevil/qdevil_qdac2.py
jordivallsq: I agree that the f string is not ideal but since we can use multiple qdacs the cache is instrument dependant, and I have to add the device name somewhere as an identification for the instrument. The cleanest change I could think is of having a dict of dicts like {self.device.name: `{channel_id: dc_list, ...}, ...}` but I don't really fancy this solution as makes debugging a bit more convoluted.

src/qililab/instruments/qdevil/qdevil_qdac2.py
jordivallsq: Completely agree! Task created

src/qililab/instrument_controllers/instrument_controller.py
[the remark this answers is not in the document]
jordivallsq: ticket created

src/qililab/instrument_controllers/instrument_controller.py
jordivallsq: modified

src/qililab/platform/platform.py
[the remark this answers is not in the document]
jordivallsq: Added to changelog

src/qililab/platform/platform.py
jordivallsq: The crosstalk should be considered as true because it is the default state for any experiment including crosstalk. In case the user wants to work in voltages they need to set crosstalk as False, otherwise the user would need to remove the crosstalk from platform and/or calibration.

That said, for any experiment not including a crosstalk (i.e. transmons) this condition is irrelevant because it will not include crosstalk anywhere.

I added a description of the variable inside execute_qprogram docstrings  and at the changelog

src/qililab/platform/platform.py
jordivallsq: this in fact modifies the calibration class when no crosstalk is desired. But it only modifies the calibration for the current execution as calibration is not a component of platform

src/qililab/qprogram/crosstalk_matrix.py
jordivallsq: fixed

src/qililab/qprogram/qblox_compiler.py
jordivallsq: In this case the crosstalk is already set as a Crosstalk class or None if it wasn't set with `platform.set_crosstalk` or if `crosstalk:bool=False`. It will never be True as is not the same parameter.
platform.py line 1327: `crosstalk=self.crosstalk if crosstalk else None,`

src/qililab/qprogram/qdac_compiler.py
jordivallsq: Added

src/qililab/qprogram/qprogram.py
jordivallsq: I was considering it as a more generic impementation but fair since QM is not considered in this PR

src/qililab/qprogram/qprogram.py
jordivallsq: It triggers the error in the lines just above
`ValueError("Play elements must be the same for the same play pulse.")`
let me know if the text is not clear enough

src/qililab/waveforms/flat_top.py
jordivallsq: it is indeed int, this was a typo

tests/qprogram/test_qblox_compiler.py
jordivallsq: removed

tests/qprogram/test_qblox_compiler.py
jordivallsq: removed

tests/qprogram/test_qdac_compiler.py
jordivallsq: thanks, removed

tests/qprogram/test_qdac_compiler.py
jordivallsq: removed

src/qililab/qprogram/qprogram.py
jordivallsq: As stated in the restart_flux_vector docstring from line 116 is a string from the class to be restarted. All operations used in the dictionaries are strings from the classes and all operations added are str(element.__class__) where element can be `Play | SetOffset | SetGain`.
In this case is Play is used because we need to create a new set of plays as we go inside a loop and we had a play already done there, this way we do not remove the play outside the loop.

tests/result/test_stream_results.py
jordivallsq: It is as I reworked the debug to take the bus mapping into account but neither the name is good fits the checks nor there is an example of the code working.
I conserved this tests and created a new test wit the previous result.

docs/releases/changelog-dev.md
jordivallsq: yes, this is due to the order of the lists/dictionaries created in `with_crosstalk_...`, since it has to use appends because block elements are lists it inverts the order, while the original is the last. This only affects the stacks of plays, set_offsets and set_gains, since the order of these instructions do not matter as they are sent at the same time this is not a problem. If you think that in the future this can create issues I can restructure it. 

But even if I maintain the order of the buses created there will be other buses in the crosstalk matrix that will be put at the bottom of the list in a somewhat arbitrary manner

docs/releases/changelog-dev.md
jordivallsq: Added an explanation on what `execute_qprogram(..., crosstalk)` as well as added in the docsrings of execute_qprogram and execute_parallel_qprograms

src/qililab/result/stream_results.py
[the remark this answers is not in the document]
jordivallsq: Both bus_mapping and calibration are optional fields in both `compile_qprogram` and `StreamArray`, defaulting to None, so if the user doesn't provide them, it will run compiler_qprogram anyway; the only reason for it to return an exception would be if the compilation fails (for example, if the user didn't provide the bus mapping when necessary)

jordivallsq (commented):
Thank you for your review @flavie-lebars, I replied to all your comments taking the necessary actions.

docs/releases/changelog-dev.md
flavie-lebars: I do think it is counterintuitive, but I cannot see an issue for now.

src/qililab/qprogram/qprogram.py
flavie-lebars: I think you forgot stepped

flavie-lebars (commented):
Looks good to me! One last thing, I remember we discussed to remove the `wait_sync `mid q1asm if not needed (ie: only one sequencer), will this be done in another PR? If yes make sure you have a task if not already

jordivallsq:
@flavie-lebars regarding the wait sync for lonely sequencers, I am currently removing it on the PR for fixing `_handle_wait_trigger`, I will add the fix in the changelog anyway

src/qililab/qprogram/qprogram.py
jordivallsq: fixed

jordivallsq:
@flavie-lebars, as far as I am concerned, everything has been replied and the checks pass.
The issue with the wait_sync is already fixed here: https://github.com/qilimanjaro-tech/qililab/pull/1099

Let me know if it is ready to merge

flavie-lebars (approved):
Great work Jordi!
FYI those previous comments still stand:
- I don't think that compiler is the appropriate word, because the input and outputs are both qprograms.

- As this is very complex, when you have time later, could you please add documentation on Notion.

- qprogram should be split and there should be a class to apply the crosstalk compensation I believe. We can see with Vyron and the refactor of qprogram.

- the crosstalk compensation for qblox should be split into different passes if feasible, it is very hard to read/review and test the chunks properly at the moment.

jordivallsq:
I missed those by mistake. Let me answer them:

- I completely agree, but as much as I checked, I don't see where I call this a compiler, only crosstalk compensation, as well as the functions being called `with_crosstalk_...`
- Of course, let me add a Notion page under QHC talking about this feature.
- That is what I was thinking about, I agree that the place for this feature could be individually outside in a class or even a function. I would wait for a refactor, but I will mention this on the Notion page as the next step.
- I don't know how feasible it is, but there is room for improvement once the class is refactored. I am adding this to part of the previous comment in Notion.

PR #1070, opened 2026-02-02
[QHC-1279] Database measurement calibration table and multiple autocalibration qubits

jordivallsq:
Oscar requested a new table on the measurement database to handle sequences of measurements such as the ones in seqtante. The main purpose will be for calibrating the crosstalk matrix but this can give many other users as it allows for total flexibility

We also need to modify the data type of the quibit index to be a string, as it can be either a int or a list.

Create the table and columns in the measurement database en pgAdmin and qililab database_measurement.py

### DIFF
--- a/src/qililab/result/database/database_measurements.py
+++ b/src/qililab/result/database/database_measurements.py
@@ -96,6 +96,29 @@ class Sample(base):  # type: ignore
         return f"{self.sample_name} {self.manufacturer} {self.additional_info}"
 
 
+class SequenceRun(base):  # type: ignore
+    """Creates and manipulates a sequence of measurements run metadata database"""
+
+    __tablename__ = "sequence_run"
+
+    sequence_id: Column = Column("sequence_id", Integer, primary_key=True)
+    date: Column = Column("date", DateTime)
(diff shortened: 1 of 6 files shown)


src/qililab/result/database/database_measurements.py
```diff
@@ -96,6 +96,29 @@ def __repr__(self):
...
+
+    def __init__(self, date, sequence_tree, sequence_completed, sample_name, cooldown):
+        self.date = date
+        self.sequence_tree = sequence_tree
+        self.sequence_completed = sequence_completed
+        self.sample_name = sample_name
+        self.cooldown = cooldown
+
+    def __repr__(self):
+        return f"{self.sequence_id} {self.date} {self.sequence_completed} {self.sample_name} {self.cooldown}"
```
DZPM: To improve readability, how about adding a bit more format:

return f"[{self.sequence_id}] ({self.date}) {self.sequence_completed} at {self.sample_name}: {self.cooldown}"

elygoner (approved):
It looks good but we should test it. Confirm that it is tested before merging. Also If we want the str qubit_idx in Seqtante we also need to do a PR on the BSC branch.

PR #1072, opened 2026-02-16
qhc-1295-bug-vna-set_parameter-with-enums

flavie-lebars:
[BUG] - VNA E5080B Driver. Enum Parameters were passing the entire Enum to the device causing an error when setting the parameter. Now the value of the Enum is passed to the device.


DZPM (approved):

jjmartinezQT (approved):
good catch.

PR #1074, opened 2026-02-19
[FIX] Measurement database folder processing for parallel qprogram with same folder path

jordivallsq:
Added target to measurements database manager and creates a folder based on target value.
added optional data_folder path to get_db_manager for easy access without Qilimanjaro's infrastructure.
Removed sequence run as it will have conflicts with bsc's database.


elygoner (approved):
LGTM, No notes. fix the coverages and you can merge.

PR #1075, opened 2026-02-24
qhc-1307-change-the-auto_norm-default-of-software-filters-to-false

flavie-lebars:
Previously, the software filters in the `PulseDistortion` module were normalised by default.
This PR changes the default value of `auto_norm` to `False`, as the previous behaviour was considered counterintuitive.

### DIFF
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -26,6 +26,10 @@
 
 ### Improvements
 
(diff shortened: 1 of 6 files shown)


jordivallsq (approved):
Well done! As long as the people from measurements give the ok (mainly affecting digital existing runcards) is ready to merge.

PR #1077, opened 2026-03-03
Qhc 1317 add resistances to calibration file

jordivallsq:
Added resistances inside `CrosstalkMatrix()` they can be set by `crosstalk.set_resistances()` in the same way as `crosstalk.set_offset`. Also they can be set inside the calibration file as resistances.


jjmartinezQT (approved):
Good to me, if Oscar has confirmed this is the format they want, then we are done with this one, good job.

PR #1079, opened 2026-03-06
Qhc-1319-save-state-of-crosstalk-during-calibration

jordivallsq:
Created crosstalk history element within Calibration. The crosstalk history describes the state of the crosstalk at each iteration of the crosstalk calibration, each time a crosstalk calibration is done the user can add an element to the list with `add_crosstalk_history` and use `save_crosstalk(experiment_name)` at every step of the calibration to update the `crosstalk_history["history"]`. The user can then save the final crosstalk information through `save_history` and remove any element on the history that was not correctly implemented with `remove_history_step`.

As an example of the code the user might run:
```
calibration = Calibration()
calibration.crosstalk_matrix = crosstalk_matrix

calibration.add_crosstalk_history()

# Intra qubit experiments
   calibration.save_crosstalk(experiment_name_1)

# Inter qubit experiments
   calibration.save_crosstalk(experiment_name_2)

calibration.save_history()
```


elygoner (commented):
LGTM.
I left a suggestion bellow.
If It has the functionalities measurements wants and has been End to End tested you can merge.

elygoner (approved):

elygoner (approved):
The code looks good but I would like more about how exactly is supposed to be used and what some specific things work in practice. But again, the code looks good and if it has been reviewed by the measurement team members that have to use it, then you can go ahead and merge.

PR #1082, opened 2026-03-27
qhc-1334-empty-qblox-acquisition-matrix

flavie-lebars:
For Qblox, the acquisition matrix needs to be reset at each `initial_setup` and `acquire_qprogram_results` call for QRM and QRM-RF instruments. The reset is performed by uploading an empty sequence. This ensures that users do not encounter limitations on the number of available bins due to previously run experiments.

The empty sequence uploaded is:

```
empty_sequence = {
            "waveforms": {},
                "weights": {},
                "acquisitions": {},
                "program": "",
        }
```

### DIFF
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -31,6 +31,19 @@
 
 ### Improvements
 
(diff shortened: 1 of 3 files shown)


fedonman (approved):
💯 💯

PR #1088, opened 2026-04-02
qhc-1340-bug-fix-stale-sequence-cache-causing-skipped-uploads-after-bin-allocation-reset

flavie-lebars:
Qblox: After each acquisition, an empty sequence is re-uploaded to ensure that bin allocation is properly reset (introduced in [this PR](https://github.com/qilimanjaro-tech/qililab/pull/1082)). Previously, the upload cache compared sequence hashes and skipped re-uploading if no changes were detected. However, since the acquisition portion of the sequence had been removed, this led to an inconsistency: the cached sequence was not re-uploaded, resulting in an error during execution.

The cache has now been removed so that the sequence is always re-uploaded, ensuring correct bin allocation and preventing this error before each run.

This is a temporary workaround. Once firmware version 0.13 is deployed across all clusters, we will transition to using the newly introduced `_update_sequence()` method from qblox-instruments and reintroducing the hashing.


jjmartinezQT (approved):
looks good to me, I can test it on the chip next week

PR #1091, opened 2026-04-13
Fix qcodes version

flavie-lebars:
Fix `qcodes` version to `0.54.3` to avoid breaking changes introduced in later releases.


jjmartinezQT (approved):
Look goods to me, could we do it as well for bsc branch please?

PR #1097, opened 2026-04-14
Qhc 1346 changes to db manager to support live plot

jordivallsq:
Modified database manager's `load_by_id` to allow a list of ids to return a list of the measurements with said ids. Also added function `db_manager.get_dc_offsets(id)`, for recent addition to the measurements database, `dc_offsets`.


src/qililab/result/database/database_manager.py
```diff
@@ -204,22 +204,24 @@ def add_calibration_run(self, calibration_tree: dict, sample_name: str, cooldown
...
-
-            if measurement_by_id is not None:
-                path = measurement_by_id.result_path
-                if not os.path.isfile(path):
-                    new_path = path.replace(self.base_path_local, self.base_path_share)
-                    measurement_by_id.result_path = new_path
-
-            return measurement_by_id
+            if not isinstance(id, list):
+                id = [id]
```
DZPM, line 215: rename "id" to "measurement_id" (or similar), as "id" is a reserved function in python
https://docs.python.org/3/library/functions.html#id

jordivallsq: changed

src/qililab/result/database/database_manager.py
```diff
@@ -204,22 +204,28 @@ def add_calibration_run(self, calibration_tree: dict, sample_name: str, cooldown
                 running_session.rollback()
                 raise e
 
-    def load_by_id(self, id: int) -> Measurement | None:
-        """Load measurement by its measurement_id.
+    def load_by_id(self, id: int | list[int]) -> list[Measurement] | Measurement | None:
```
elygoner, line 223: Might be worth to add an overload to guard for future typing issues when used in qilitools

elygoner (approved):
LGTM. I have left A comment. Approved with the precondition that has been properly tested. Also resolve conflicts.

src/qililab/result/database/database_manager.py
jordivallsq: Good suggestion, added it

DZPM (approved):

jordivallsq:
tested and working

PR #1098, opened 2026-04-17
qhc-1356-bug-bin-aquisition-index-not-incrementing-qblox

flavie-lebars:
[BUG] Fixed a bug in the Qblox compiler where the bin acquisition index was not incrementing correctly when multiple `measure` calls are used sequentially inside an `average` block with an outer sweep loop.  Each sequential acquire now gets its own bin register initialised to its position offset, and the bin register is advanced by the total number of acquires per sweep step (instead of always 1), so that consecutive acquires write to consecutive bins and the full acquisition matrix is filled correctly.


DZPM (approved):
Looks good to me 👍

jordivallsq:
If the tests pass anyway means that this case was not registered in the tests for qblox compiler. Could you add an example of a qprogram where this is the case and with a resulting Q1ASM with the adds in the right place?

flavie-lebars:
@jordivallsq I am still working on this PR and currently adding the tests, I am also waiting for Ely to test in the hardware. I will let the team know once the PR is ready to be reviewed. Thanks.

jordivallsq:
No problem! Good catch anyway

fedonman (approved):
Great catch! 💯 I am amazed how this was left unnoticed. Does this mean that inner averaging was not actively used until now?

flavie-lebars:
@fedonman Thanks!
Fluxonium use inner average but they only have one measure, which worked fine.
Ely tried to use an inner average with two measures, I think this was the first time this was attempted. Transmons experiments can have more than one measure but they tend to have the average as an outer loop.

PR #1099, opened 2026-04-17
Qhc 1358 [BUG-FIX] Wait trigger does not create right durations and creates unnecessary wait_sync

jordivallsq:
- Fixed a bug for `wait_trigger` where for large durations the waiting time was not correctly implemented. Now it uses `_handle_add_waits` like `wait`.
  These are some examples of uses:
  - Duration of 4 creates a simple Q1ASM with the wait trigger with the minimal wait duration, not defining a port defaults to port 15:
    ```
        qp.wait_trigger(bus="drive", duration=4)
      ---
        wait_trigger     15, 4
    ```
  - Duration of 70,000 (and any duration bigger than 8) will execute `_handle_add_waits` and create waits with duration minus 4 ns (from the wait trigger):
    ```
        qp.wait_trigger(bus="drive", duration=70_000, port=1)
      ---
        wait_trigger     1, 4
        wait             65532
        wait             4464
    ```
  - If the wait_trigger is set before any parameter update qprogram defines an `upd_param` of 4 ns (adding this value to the total time):
    ```
        qp.set_frequency(bus="readout", frequency=1e6)
        qp.wait_trigger(bus="drive", duration=1_000, port=1)
      ---
        upd_param        4
        wait_trigger     1, 4
        wait             992
    ```
- Removed unnecessary wait_syncs created by `qp.wait_trigger` when the instrument has only one sequencer. 
  Now for one sequencer it will behave like this:

    ```
        qp.set_frequency(bus="drive", frequency=1e6)
        qp.wait_trigger(bus="drive", duration=1_000, port=1)
      ---            
        set_freq         4000000
        set_freq         4000000
        upd_param        4
        wait_trigger     1, 4
        wait             992
    ```
  Whereas with two buses there will be an extra wait?sync in the Q1ASM (also compensated by duration):
    ```
        qp.set_frequency(bus="drive", frequency=1e6)
        qp.set_frequency(bus="readout", frequency=1e6)
        qp.wait_trigger(bus="drive", duration=1_000, port=1)
      ---
      drive sequencer:
        set_freq         4000000
        set_freq         4000000
        upd_param        4
        wait_trigger     1, 4
        wait             988
        wait_sync        4

      readout sequencer:
        set_freq         4000000        
        set_freq         4000000        
        wait_sync        4 
    ```


src/qililab/qprogram/qblox_compiler.py
```diff
@@ -1057,74 +1057,48 @@ def _handle_wait_trigger(self, element: WaitTrigger):
 
         duration: QPyProgram.Register | int
         convert = QbloxCompiler._convert_value(element)
-        duration = convert(element.duration)
+        duration = convert(element.duration if len(self._buses) == 1 else element.duration - 4)
```
DZPM, line 1060: Please add a comment to explain this line, it's not obvious without the changelog in context

DZPM:
:+1:

docs/releases/changelog-dev.md
```diff
@@ -4,10 +4,71 @@
...
+        qp.set_frequency(bus="drive", frequency=1e6)
+        qp.wait_trigger(bus="drive", duration=1_000, port=1)
+      ---            
+        set_freq         4000000
+        set_freq         4000000
+        upd_param        4
+        wait_trigger     1, 4
+        wait             992
+    ```
+  Whereas with two buses there will be an extra wait?sync in the Q1ASM (also compensated by duration):
```
flavie-lebars: typo on wait_sync

flavie-lebars: this is not correct, you cannot compensate the `wait_sync` by duration bc the `wait_sync` is not deterministic:
`wait_sync 4` does not take 4 ns.
https://www.notion.so/qilimanjaro/wait_sync-34c7eec14c5380a1be00fec52cccae0b?source=copy_link

Can you please make sure the experimentalists are aware of this, and also since this will chnage the way their current experiment are run, this needs to be tested before merging.

docs/releases/changelog-dev.md
```diff
@@ -4,10 +4,71 @@
 
 ### Improvements
 
+- Removed unnecessary wait_syncs created by `qp.wait_trigger` when the instrument has only one sequencer.
```
flavie-lebars: We need to be a lot more careful with PR naming in the future, the name only refers to half of the change, I would have two different PRs, they are not related.

src/qililab/qprogram/qblox_compiler.py
flavie-lebars: if you have more than one bus and there is a variable, you do not reach this line. please fix this and add a test, the variable check needs to be done before

```
if isinstance(element.duration, Variable):
            raise ValueError("Wait trigger duration cannot be a Variable, it must be an int.")
```

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -1057,74 +1057,48 @@ def _handle_wait_trigger(self, element: WaitTrigger):
...
-        if self._buses[bus].upd_param_instruction_pending:
-            if (
-                4 <= duration and duration <= 8
-            ):  # you cannot play an update param and then a wait bc both have a minimum of 4
-                self._buses[bus].qpy_block_stack[-1].append_component(component=QPyInstructions.UpdParam(duration))
-                self._buses[bus].qpy_block_stack[-1].append_component(
-                    component=QPyInstructions.WaitTrigger(address=port, wait_time=4)
+        if self._buses[element.bus].upd_param_instruction_pending:
+            self._buses[element.bus].qpy_block_stack[-1].append_component(
+                component=QPyInstructions.UpdParam(duration if 4 < duration < 8 else 4)
```
flavie-lebars: pls replace all the appropriates 4 by the cst `INST_MIN_WAIT`

flavie-lebars (changes requested):
- The logic of the long wait is the same between `wait` and `wait_trigger`, but what would be cleaner is to have the `upd_param` in the `_handle_wait` and the `_handle_wait_trigger` functions directly and then call a shared long wait function. Because at the moment the `upd_param` part in `_handle_add_waits` only concerns the regular wait and not the wait_trigger.

- The sync is false, the duration with the upd_param is not tracked correctly (this is a prexisiting bug, please add it as such in the changelog)
```
qp = ql.QProgram()

qp.set_frequency("drive",100)
qp.wait_trigger("drive",2)
qp.wait("readout",2)
qp.sync()
compiler = ql.QbloxCompiler()
seq = compiler.compile(qp)
seq.sequences["drive"]._program
```

I have other general comments regarding the wait_trigger:
- Why is the below check done in the compiler? I think it should be in the platform. This means we cannot check the compilation without the runcard, we cannot have this sort of things, we need to be careful with software layers.
```
if not self._ext_trigger:
        #     raise AttributeError("External trigger has not been set as True inside runcard's instrument controllers.")
```
- Why has the `wait_trigger` not been implemented in qblox draw? I understand it is not deterministic but I think we can treat it as a regular wait and document the behaviour. For example, we can do `set_cond `in the draw, which is also non deterministic.

Please all these points need to be adressed, but not in one single PR.
I would suggest having one PR for the wait_sync, and another for the bugs with wait_trigger and another for the two points above (draw & moving the self._ext_trigger error in the platform)

src/qililab/qprogram/qblox_compiler.py
jordivallsq: done

src/qililab/qprogram/qblox_compiler.py
jordivallsq: Nice catch! changed

docs/releases/changelog-dev.md
jordivallsq: fixed

src/qililab/qprogram/qblox_compiler.py
jordivallsq: This condition no longer exists

docs/releases/changelog-dev.md
jordivallsq: I used the automatic naming system from the task on linear, changing that.
Despite that I believe that the important place here we must include all changes are the changelog and PR description

docs/releases/changelog-dev.md
jordivallsq: Thank you for the information, as you could see, I was not aware that `wait_sync` was nondeterministic. I fixed any implementation that uses wait sync to use update parameter after the wait sync and to do not count either wait trigger nor wait sync to the full wait duration (also defined in the changelog

```
        qp.set_frequency(bus="drive", frequency=1e6)
        qp.set_frequency(bus="readout", frequency=1e6)
        qp.wait_trigger(bus="drive", duration=1_000, port=1)
      ---
      drive sequencer:
        set_freq         4000000        
        set_freq         4000000        
        wait_trigger     1, 4           
        wait_sync        4              
        upd_param        4              
        wait             996            

      readout sequencer:
        set_freq         4000000        
        set_freq         4000000        
        wait_sync        4 
```

regarding the notification I was considering of documenting it in the crosstalk notion page also attaching teh link you sent as it is informative enough rgarding the situation

jordivallsq (commented):
Adressed all the comments.

@flavie-lebars, regarding your comments:
- For the upd_parameter I would leave it inside `_handle_add_waits` as right now both `_handle_wait_trigger` and `_handle_wait` use upd_parameter through that function (with only one exception where there is only one sequencer with wait_trigger. But since this is an exception I would say that adding that on `_handle_wait` is unnecessary.
- Regarding the sentence about the sync being false it was not clear enough for me. Is ithis comment related to `wait_sync` because if it is the case I imlemented the suggested solution in the comment. Otherwise please elaborate and we can create a task to fix this.

- The check done in the compiler is because it cannot be done in the platform level as without the qprogram information on whether we use qp.wait_trigger AND platform's information on the qblox external trigger is connected. In the compile without the runcard we can always state `ext_trigger = True`.
- I honestly didn't think about it I will create a PR to add itand treat it as a regular wait. Maybe with a disclaimer that is not deterministic?

jordivallsq:
Regarding the issues:
- Moved the error raising from within the compiler to `_execute_qblox_compilation_output`
- Added wait trigger inside the drawer as another wait

docs/releases/changelog-dev.md
```diff
@@ -119,6 +119,43 @@ With `execute_qprogram(..., crosstalk= True / False)` the parameter introduced i
...
```
flavie-lebars, line 7: please include all changes done

tests/qprogram/test_qblox_compiler.py
```diff
@@ -637,39 +664,39 @@ def fixture_single_bin_different_depth_qp() -> QProgram:
...
-            waveform=square_wf,
-            weights=weights_shape)
-
-    qp.measure(bus=f"readout",
-                waveform=square_wf,
-                weights=weights_shape)
-    qp.measure(bus=f"readout",
-                waveform=square_wf,
-                weights=weights_shape)
+        qp.measure(bus=f"readout", waveform=square_wf, weights=weights_shape)
```
flavie-lebars: Please fix all the sonarclouds issues regarding the f string

src/qililab/platform/platform.py
```diff
@@ -356,6 +356,8 @@ def __init__(self, runcard: Runcard):
 
         self.trigger_runs: int = 0
 
+        self.has_ext_trigger: bool = False
```
flavie-lebars: docstring

flavie-lebars (commented):
As previously suggested this should have been done in three different PRs, having many changes like this make the review process very difficult. Could you please create separate PRs:

1. Bug fix: fix wait_trigger duration handling + drawer
2. Skip wait_sync for single-bus
3. Move ext_trigger check to platform

Also, the wait_trigger duration still has bugs, please review the implementation:

For example:
```
qp = ql.QProgram()
qp.set_frequency(bus="drive", frequency=1000)
qp.set_frequency(bus="readout", frequency=1000)
qp.wait_trigger(bus="drive", duration=1_000)
compiler = ql.QbloxCompiler()
seq = compiler.compile(qp)
seq.sequences["drive"]._program
```
```
setup:
                wait_sync        4              
                set_mrk          0              
                upd_param        4              

main:
                set_freq         4000           
                set_freq         4000           
                wait_trigger     15, 4          
                wait_sync        4              
                upd_param        4              
                wait             996            
                set_mrk          0              
                upd_param        4              
                stop                            
```

src/qililab/platform/platform.py
jordivallsq: fixed here in the ext_trigger error PR
https://github.com/qilimanjaro-tech/qililab/pull/1112

docs/releases/changelog-dev.md
jordivallsq: All changes added

tests/qprogram/test_qblox_compiler.py
jordivallsq: Fixed all f strings

jordivallsq:
@flavie-lebars As suggested, I created new PRs:

3: Created a new PR for this topic: Move ext_trigger check to platform. github.com/qilimanjaro-tech/qililab/pull/1112

1 and 2 remain in this PR, since skipping wait_sync for single-bus is a small change and the testing requires this PR anyway for it to make sense (added testing for single buses though)

I do not understand the bug that you mention, as I see the generated Q1ASM that you show in the example is the expected one. Please state what is wrong.

flavie-lebars:
@jordivallsq the sum of the duration is wrong: 996 + 4 + 4

flavie-lebars:
This PR has been cancelled. This PR replaces it: [(https://github.com/qilimanjaro-tech/qililab/pull/1201)]

PR #1100, opened 2026-04-23
[QHC-1364] Save Bus Distortions Using save_platform

elygoner:


jordivallsq:
Can we add a new test for this error? basically checks a full platform (the default runcard saved in tests) saves it as a platform and then reads it again and compares the two platforms. In test_platform.py

flavie-lebars (changes requested):
Nice catch Ely!
Be careful with the changelog you have modified the final one used for releases, please modify the changelog-dev one

tests/platform/test_platform.py
```diff
@@ -344,6 +344,69 @@ def fixture_qp_quantum_machine() -> QProgram:
...
+            differences.append(f"{path}: {obj1!r} != {obj2!r}")
+        return obj1 == obj2, differences
+
+    # Lists and tuples
+    if isinstance(obj1, (list, tuple)):
+        if len(obj1) != len(obj2):
+            differences.append(f"{path}: length mismatch ({len(obj1)} vs {len(obj2)})")
+            # Still compare up to the shorter length
+        for i, (a, b) in enumerate(zip(obj1, obj2)):
+            eq, diffs = compare_platforms(a, b, path=f"{path}[{i}]", depth=depth + 1, max_depth=max_depth)
```
jjmartinezQT: eq is not used later on, so better replacing it with an underscore
```suggestion
            _, diffs = compare_platforms(a, b, path=f"{path}[{i}]", depth=depth + 1, max_depth=max_depth)
```

jjmartinezQT (dismissed):
well spotted this one!

tests/platform/test_platform.py
```diff
@@ -806,24 +869,13 @@ def test_serialization(self, platform: Platform):
...
-        assert isinstance(new_runcard_dict, dict)
-        assert new_runcard_dict == runcard_dict
-
-        newest_platform = Platform(runcard=Runcard(**new_runcard_dict))
-        assert isinstance(newest_platform, Platform)
-        assert str(newest_platform) == str(new_platform)
-        assert str(newest_platform.name) == str(new_platform.name)
-        assert str(newest_platform.buses) == str(new_platform.buses)
-        assert str(newest_platform.instruments) == str(new_platform.instruments)
-        assert str(newest_platform.instrument_controllers) == str(new_platform.instrument_controllers)
```
jordivallsq, line 826: The compare platform is good, but I don't understand why you had to remove these assertions

jordivallsq (approved):
Well done! Left a comment on the tests

tests/platform/test_platform.py
elygoner: all of that is already compared with the new code, the only difference is that with the old test all are explicit. if any of those things change it will be cached on the function.

tests/platform/test_platform.py
```diff
@@ -344,6 +344,69 @@ def fixture_qp_quantum_machine() -> QProgram:
...
+
+    # Dicts
+    if isinstance(obj1, dict):
+        keys1, keys2 = set(obj1.keys()), set(obj2.keys())
+        for k in keys1 - keys2:
+            differences.append(f"{path}[{k!r}]: key only in first object")
+        for k in keys2 - keys1:
+            differences.append(f"{path}[{k!r}]: key only in second object")
+        for k in keys1 & keys2:
+            eq, diffs = compare_platforms(obj1[k], obj2[k], path=f"{path}[{k!r}]", depth=depth + 1, max_depth=max_depth)
```
jordivallsq:
```suggestion
            _, diffs = compare_platforms(obj1[k], obj2[k], path=f"{path}[{k!r}]", depth=depth + 1, max_depth=max_depth)
```
Here you had the same issue

flavie-lebars:
very minor, but can you please add the line of the changelog at the top of the bugs, we tend to follow this convention.

flavie-lebars (approved):
Image

PR #1101, opened 2026-04-23
qhc-1356-bug-bin-aquisition-index-not-incrementing-qblox-bsc

flavie-lebars:
[BUG] Fixed a bug in the Qblox compiler where the bin acquisition index was not incrementing correctly when multiple `measure` calls are used sequentially inside an `average` block with an outer sweep loop.  Each sequential acquire now gets its own bin register initialised to its position offset, and the bin register is advanced by the total number of acquires per sweep step (instead of always 1), so that consecutive acquires write to consecutive bins and the full acquisition matrix is filled correctly.


fedonman (approved):
💯

PR #1103, opened 2026-05-04
[QHC-1371] DEBUG Transpilation Differences in BSC Branches

elygoner:
In #1085 empty PulseSchedule-s for flux buses were eliminated because they would cause an error if using an instrument like S4g, since it doesn't support pulses. But the empty pulses were initially created to set the offsets.

This Issue can be solved just eliminating the buses of the instruments that don't support pulses from digital compilation settings since they aren't needed there. 
The changes in that commit will be walked back.


elygoner:
No estoy segura de que poner en el changelog.

jjmartinezQT (approved):
good job!

PR #1104, opened 2026-05-05
Qhc 1375 bugfix instrument controller features not propagated to to_dict

jordivallsq:
Fixed a bug where the qblox instrument controller parameter `ext_trigger` and the qdac instrument controller parameter `reference_clock` where not correctly translated to dictionary from the runcard and therefore not saved with `ql.save_platform(platform)`.


src/qililab/instrument_controllers/instrument_controller.py
```diff
@@ -297,4 +297,5 @@ def to_dict(self):
             INSTRUMENTCONTROLLER.CONNECTION: self.connection.to_dict(),
             INSTRUMENTCONTROLLER.MODULES: [module.to_dict() for module in self.settings.modules],
             INSTRUMENTCONTROLLER.RESET: self.settings.reset,
+            INSTRUMENTCONTROLLER.REFERENCE_CLOCK: self.settings.reference_clock,
```
elygoner, line 300: We are still missing the ext_trigger field in the to_dict, no?
It is a field in the Setting that is lost.
Also if you add the ext_trigger here you don't need to define an explicit to_dict in the qblox_instrument_controller.

elygoner (changes requested):

src/qililab/instrument_controllers/instrument_controller.py
jordivallsq: The `ext_trigger` is defined in the `qblox_instrument_controller.py` as it is exclusive to this instrument. It is not useful on any other machine. In the tests, the ' ext_trigger ' has been added in the comparison of runcards before and after saving (using todict and it does not crash meaning that works (I also tested myself)

elygoner: Then why is it in the abc settings. That would suggest that is a somewhat common setting. 
Right now there is a `ext_trigger` in the class that is not being propagated to the` to_dict`, so we have to put it on the `to_dict` or eliminate it from the `InstrumentControllerSettings`.

elygoner (approved):
All good now.

PR #1106, opened 2026-05-06
Qhc 1362 missing triggers handling

jordivallsq:
Added `timeout_repetitions` parameter for QRM and QRM-RF instruments sequencers inside the runcard. This parameter controls how many (if any) executions of the same qblox qprogram execution must be done after an acquisition `TimeoutError`. Defaults to no repetitions.
In the runcard this parameter is located inside the instruments sequencer for QRM and QRM-RF modules.

  ```
    - name: QRM-RF
    alias: QRM-RF1
    ...
    awg_sequencers:
    - identifier: 0
      ...
      acquisition_timeout: 1  # In minutes
      timeout_repetitions: 3  # Optional parameter, defaults to None
      ...
  ```


jordivallsq:
Tested in HW

src/qililab/instruments/qblox/qblox_qrm.py
```diff
@@ -446,6 +449,18 @@ def _set_acquisition_timeout(self, value: int | float | str | bool, sequencer_id
         """
         cast("QbloxADCSequencer", self.get_sequencer(sequencer_id)).acquisition_timeout = int(value)
 
+    def _set_timeout_repetitions(self, value: int | float | str | bool, sequencer_id: int):
```
flavie-lebars: can it be a float or bool, or str?

src/qililab/platform/platform.py
```diff
@@ -1456,16 +1455,18 @@ def _execute_qblox_compilation_output(self, output: QProgramCompilationOutput, d
...
-
                 # Reset instrument settings
                 for bus_alias in sequences:
                     for instrument, channel in zip(buses[bus_alias].instruments, buses[bus_alias].channels):
                         if isinstance(instrument, QbloxModule):
                             instrument.desync_sequencer(sequencer_id=int(channel))
                 self.trigger_runs += 1
 
-                if self.trigger_runs <= 3:
+                timeout_repetitions = bus.check_recurrent_timeout()
```
flavie-lebars, line 1465: There is a bug here, `bus` is whatever was the last bus defined line 1435. It is not guaranteed to be the QRM bus with the `wait_trigger` and the timeout repeitions in the runcard. If it isn't, `check_recurrent_timeout()` returns None, the retry condition is never met, and the timeout is re-raised without retrying.

src/qililab/platform/platform.py
```diff
@@ -1456,16 +1455,18 @@ def _execute_qblox_compilation_output(self, output: QProgramCompilationOutput, d
...
+                if timeout_repetitions and self.trigger_runs <= timeout_repetitions:
+                    logger.info(
```
flavie-lebars: is there a reason we are not both raising and logging the warning?

docs/releases/changelog-dev.md
```diff
@@ -81,6 +81,23 @@ With `execute_qprogram(..., crosstalk= True / False)` the parameter introduced i
 
 ### Improvements
 
+- Added `timeout_repetitions` parameter for QRM and QRM-RF instruments sequencers inside the runcard. This parameter controls how many (if any) executions of the same qblox qprogram execution must be done after an acquisition `TimeoutError`. Defaults to no repetitions.
```
flavie-lebars: I would add in breaking change the fact that now the retry of 3 times is no longer done by default.

flavie-lebars (changes requested):

docs/releases/changelog-dev.md
jordivallsq: Good point, added

src/qililab/platform/platform.py
jordivallsq: With `logger.warning` (I changed it as it is more fitting), it will be saved as a warning and raised for the user. `warnings.warn(...)` will be redundant, and because it will report the information twice

src/qililab/instruments/qblox/qblox_qrm.py
jordivallsq: No, removed other options other than int

src/qililab/platform/platform.py
jordivallsq: The timeout error can only be raised when running `instrument.acquire_qprogram_results` inside `bus.acquire_qprogram_results`. This depends on the bus being a QRM type module (line 274 of `bus.py`: `if isinstance(instrument, QbloxQRM):`), so in principle it is only being raised when bus is a `QbloxQRM` module. 
Despite this, this implementation is weak and prone to error so I am finding a solution to make it more clear or at least to make sure that the bus taken is guaranteed to be a QRM module

jordivallsq: added extra step in case bus instrument is not a QbloxQRM

src/qililab/instruments/qblox/qblox_adc_sequencer.py
```diff
@@ -33,6 +33,7 @@ class QbloxADCSequencer(QbloxSequencer):
     threshold: float
     threshold_rotation: float
     time_of_flight: int  # nanoseconds
+    timeout_repetitions: int | None = None
```
fedonman: Question: Why do we allow `None` as an option and not use zero as the default value? Like
```suggestion
    timeout_repetitions: int = 0
```

flavie-lebars (commented):
@jordivallsq Has this been tested on hardware?

src/qililab/instruments/qblox/qblox_qrm.py
```diff
@@ -446,6 +449,18 @@ def _set_acquisition_timeout(self, value: int | float | str | bool, sequencer_id
...
         cast("QbloxADCSequencer", self.get_sequencer(sequencer_id)).acquisition_timeout = int(value)
 
+    def _set_timeout_repetitions(self, value: int, sequencer_id: int):
+        """set timeout_repetitions for the specific channel
+
+        Args:
+            value (int): value to update
+            sequencer_id (int): sequencer to update the value
+
+        Raises:
```
flavie-lebars: it doesn't raise a ValueError

src/qililab/instruments/qblox/qblox_adc_sequencer.py
jordivallsq: Good point, changed to only ints default to 0

src/qililab/instruments/qblox/qblox_qrm.py
jordivallsq: This docstring is designed after the rest of `_set_...` functions which all of them contain this raise in the docstring. It is not a raise that appears in the function but a ValueError that is risen by `int()` itself when the value introduced cannot be converted to int.
Nonetherless, as it is not a raise from the function itself I removed it

jordivallsq:
This feature has been tested on hardware

flavie-lebars (approved):
thank you Jordi!

PR #1111, opened 2026-05-18
Qhc 1375 Returned default spi reset to false

jordivallsq:
Returned the default value of the QbloxSPIRackControllerSettings reset to False


elygoner (approved):
Yup

PR #1112, opened 2026-05-18
Qhc 1358 Remove ext_trigger check from platform

jordivallsq:
Moved qblox external trigger check from compiler to platform. This avoids raising errors related to the Runcard inside the compiler.


src/qililab/qprogram/structured_program.py
```diff
@@ -90,6 +91,15 @@ def buses(self) -> set[str]:
...
 
+    @property
+    def ext_trigger(self) -> bool:
+        """Get the external trigger flag.
+
+        Returns:
+            bool: Flag remarking the external trigger being used
+        """
+        return self._ext_trigger
+
```
fedonman: I wouldn't add this as a attribute / property of QProgram as a) it has nothing to do with the QProgram per se, b) it is Qblox-specific and c) I foresee it will create a breaking change with the QProgram refactor.

What we want is a way to say that the compiled qprogram requires the platform to have an external trigger, right?

Why not add a property to the `QbloxCompilationOutput`, e.g. `requires_external_trigger: bool`?

fedonman (changes requested):
Made a comment on the location of the flag.

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -1169,6 +1167,8 @@ def _handle_wait_trigger(self, element: WaitTrigger):
         if element.bus not in self._qblox_buses:
             return
 
+        self._qprogram._ext_trigger = True
```
flavie-lebars: I don't understand this line of code, why is there still a reference to the ext_trigger in the compiler? This is known from the qprogram directly now

src/qililab/platform/platform.py
```diff
@@ -1284,7 +1288,7 @@ def compile_qprogram(
                 for controller in self.instrument_controllers.elements
                 if isinstance(controller, QbloxClusterController)
             ]
-            ext_trigger = any(controller.ext_trigger for controller in instrument_controllers)
+            self.has_ext_trigger = any(controller.ext_trigger for controller in instrument_controllers)
```
flavie-lebars: this only runs if all the controllers are qblox, I think it's a bug

flavie-lebars (changes requested):
Why do we even need the flag in the runcard? Another architecture could be that as soon as a qprogram has a wait_trigger we set up whatever needs to be set up between the compilation and execution. The downisde would be that this would need to be done at runtime (but this would be negligible) but this would avoid the user needing to put this flag and make things simpler in my opinion.

Also, I completely agree with Vyron's comment.

src/qililab/qprogram/structured_program.py
flavie-lebars: To answer the first question, yes exactly. Right now, you cannot compile a qprogram that has a `wait_trigger` without a runcard.

And yes, I completely agree with you. I had done something similar for the `measure_reset`, but I had this flag on the `_QbloxInterface` instead (`latch_enabled` and `trigger_network_required`).

src/qililab/platform/platform.py
jordivallsq: from line 1240:
```
instruments = {
            instrument
            for bus in buses
            for instrument in bus.instruments
            if isinstance(instrument, (QbloxModule, QuantumMachinesCluster))
        }
```
Instruments only constitute Qblox or QM instruments, the rest (such as qdac or others are ignored), This is a check to select Qblox or QM as they are both mutally exclusive

I can change the name of instruments to cluster_instruments to make it clearer

src/qililab/qprogram/qblox_compiler.py
jordivallsq: It was code not deleted, removed

src/qililab/qprogram/structured_program.py
jordivallsq: I changed the external trigger to be carried by `QbloxCompilationOutput`. Avoiding using the runcard, so now that parameter has been removed from the runcard and whenever a `wait_trigger` is used it creates the external trigger inside the trigger network

jordivallsq (commented):
@flavie-lebars @fedonman Following your advice, the external_trigger flag in the runcard has been removed, now the device creates this external trigger inside the trigger network when wait_trigger is used.

> The downisde would be that this would need to be done at runtime (but this would be negligible) but this would avoid the user needing to put this flag and make things simpler in my opinion.

I need to test this on hardware to see if the qblox controller has any reaction to run 
```
self.device.ext_trigger_input_trigger_en(True)
self.device.ext_trigger_input_trigger_address(15)
self.device.ext_trigger_input_delay(0)
```
many consecutive times (I don't expect anything but just in case).


new changelog

Removed `external_trigger` parameter from within the runcard's qblox controller instrument. Now the function `QbloxClusterController.set_ext_trigger` is risen internally every time a qprogram contains a `wait_trigger` using the trigger channel 15 (last one).

fedonman (approved):
LGTM 💯 Thank you Jordi! 🚀

docs/releases/changelog-dev.md
```diff
@@ -4,6 +4,9 @@
 
 ### Improvements
 
+- Removed `external_trigger` parameter from within the runcard's qblox controller instrument. Now the function `QbloxClusterController.set_ext_trigger` is risen internally every time a qprogram contains a `wait_trigger` using the trigger channel 15 (last one).
```
flavie-lebars: this should be marked as a breaking change. If they keep their runcard as is, it will fail. My suggestion is to have a future warning. I have a pending PR on the integration length where I will also remove a field from the runcard, you can check how I did it here https://github.com/qilimanjaro-tech/qililab/tree/qhc-1187-integration-length-fix

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -346,10 +348,12 @@ def traverse(block: Block):
 
         self._sync_counter = 0
         self._buses = self._populate_buses()
-        self._ext_trigger = ext_trigger
         self._single_channel = single_channel if single_channel is not None else []
         self._acquisition_metadata = {}
 
+        # Pre-processing: Set initial external trigger buses to an empty list
```
flavie-lebars: the dosctring is wrong, not a list

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -87,10 +87,13 @@ class QbloxCompilationOutput:
...
     """
 
-    def __init__(self, qprogram: QProgram, sequences: Sequences, acquisitions: Acquisitions):
+    def __init__(
+        self, qprogram: QProgram, sequences: Sequences, acquisitions: Acquisitions, external_trigger: bool = False
+    ):
         self.qprogram = qprogram
         self.sequences = sequences
         self.acquisitions = acquisitions
+        self.external_trigger = external_trigger
```
flavie-lebars: in my opinion, the flag shouldn't be inferred in the compiler (_handle_wait_trigger), it's known at qprogram-authoring time. We already have this pattern: trigger_network_required lives on QProgram._QbloxInterface, set inside measure_reset(). external_trigger should follow this. But I think @fedonman's comment suggested to move it here in the compiler. Happy to discuss this at a sprint/daily.

src/qililab/platform/platform.py
```diff
@@ -1388,6 +1382,10 @@ def execute_compilation_output(
 
     def _execute_qblox_compilation_output(self, output: QProgramCompilationOutput, debug: bool = False):
         try:
+            if output.qblox.external_trigger:  # type: ignore[union-attr]
+                for controller in self.instrument_controllers.elements:
```
flavie-lebars, line 1302: this would apply it to every cluster on the runcard. this should not be the case, only the ones which need the triggers should set the ext trigger

flavie-lebars (changes requested):

src/qililab/qprogram/qblox_compiler.py
jordivallsq: fixed

src/qililab/platform/platform.py
jordivallsq: Fixed, now the flag is a string list that returns the buses to find out which controller needs to be set as external

src/qililab/qprogram/qblox_compiler.py
jordivallsq: I have implemented these changes in a single commit ([Implemented requested changes](https://github.com/qilimanjaro-tech/qililab/pull/1112/commits/d6db43736a29fe1aa58d53fabf82597069d79c1c)) so the changes are reversible. Let's see what @fedonman thinks of the changes.

docs/releases/changelog-dev.md
jordivallsq: docstring moved to towncrier removal, closest to breaking change

PR #1113, opened 2026-05-18
qhc-1387-bug-set-offset-negative-move-register

flavie-lebars:
Fixed a bug in `set_offset` where using a `Variable` on one path and a negative static value on the other would generate a `move` instruction with a negative immediate, which is invalid Q1ASM.

### DIFF
--- a/src/qililab/qprogram/qblox_compiler.py
+++ b/src/qililab/qprogram/qblox_compiler.py
@@ -832,7 +832,7 @@ class QbloxCompiler:
                     component=QPyInstructions.Move(var=abs(value), register=register), bot_position=insert_position
                 )
 
-                if value < 0:
+                if value < 0: # take the two's complement
                     self._buses[element.bus].qpy_block_stack[block_index_for_move_instruction].append_component(
                         component=QPyInstructions.Nop(), bot_position=insert_position
                     )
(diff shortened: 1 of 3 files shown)


src/qililab/qprogram/qblox_compiler.py
```diff
@@ -824,12 +824,27 @@ def _handle_set_offset(self, element: SetOffset):
...
+                        component=QPyInstructions.Nop(), bot_position=insert_position
+                    )
+                    self._buses[element.bus].qpy_block_stack[block_index_for_move_instruction].append_component(
+                        component=QPyInstructions.Not(register, register), bot_position=insert_position
+                    )
+                    self._buses[element.bus].qpy_block_stack[block_index_for_move_instruction].append_component(
+                        component=QPyInstructions.Nop(), bot_position=insert_position
+                    )
+                    self._buses[element.bus].qpy_block_stack[block_index_for_move_instruction].append_component(
+                        component=QPyInstructions.Add(register, 1, register), bot_position=insert_position
```
DZPM, line 846: Why this hardcoded `1`? Should have a comment explaining it

DZPM (approved):

src/qililab/qprogram/qblox_compiler.py
flavie-lebars: It is to take the two's complement, I have added a comment.

fedonman (approved):
LGTM 💯

PR #1114, opened 2026-05-18
qhc-1388-bug-qpqbloxplay-with-0-duration

flavie-lebars:
Fixed a bug where `qp.qblox.play` with `wait_time=0` was treated as no `wait_time` provided, producing incorrect Q1ASM. The wait time is now correctly clamped to the minimum valid value of 4 ns.

### DIFF
--- a/src/qililab/qprogram/qblox_compiler.py
+++ b/src/qililab/qprogram/qblox_compiler.py
@@ -1680,7 +1680,7 @@ class QbloxCompiler:
         if waveform_variables:
             logger.error("Variables in waveforms are not supported in Qblox.")
             return
-        if element.wait_time:
+        if element.wait_time is not None:
             # The qp.qblox.play() was used. Don't apply optimizations
             index_I, index_Q, _ = self._append_to_waveforms_of_bus(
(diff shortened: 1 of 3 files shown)


fedonman (approved):
LGTM 💯

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -1680,7 +1680,7 @@ def _handle_play(self, element: Play):
         if waveform_variables:
             logger.error("Variables in waveforms are not supported in Qblox.")
             return
-        if element.wait_time:
+        if element.wait_time is not None:
```
DZPM, line 1683: Suggestion: adding a test where wait_time=None will make this more resilient

DZPM (approved):

src/qililab/qprogram/qblox_compiler.py
flavie-lebars: I'll add it. Note there's already a test for wait_time=0 which covers the actual behavioral change. This one is just verifying that None is not None evaluates to False.

PR #1118, opened 2026-05-26
Qhc 1380 non linear crosstalk implementation for with crosstalk qblox

jordivallsq:
Added Qblox automatic non-linear crosstalk compensation inside qprogram. This feature requires a `NonLinearCrosstalkMatrix` as the input crosstalk matrix and it makes use of the features given by `NonLinearFluxVector`. 

It functions much like the linear crosstalk compensation as it modifies the qprogram to adjust the flux buses into their correct bias, in this case the compensation is non-linear.

As the conversion to the new bias is non-linear, the loops given to the qblox Q1ASM cannot be linear and are therefore unpacked limiting the amount of instructions to be given to the Q1ASM sequencer. As well as the `qp.play` waveforms given, if those are arbitrary and inside a loop, it can lead to a waveform memory limitation. Square pulses and `qp.set_offset` does not have this limitation.

The changes involving `NonLinearFluxVector` have to be fixed in its respective PR and therefore I will not mention them in the changelog.


docs/releases/changelog-dev.md
```diff
@@ -141,43 +186,6 @@ With `execute_qprogram(..., crosstalk= True / False)` the parameter introduced i
 
   [#1030](https://github.com/qilimanjaro-tech/qililab/pull/1030)
 
-- Added `NonLinearFluxVector` class for managing per-bus flux offsets and gain values with crosstalk compensation across multi-loop sweeps.
```
flavie-lebars: I assume it is an error that you are deleting the changelog for flux vector.

src/qililab/qprogram/qprogram.py
```diff
@@ -482,19 +491,51 @@ def traverse(
...
-                    )
-                    crosstalk_elements.add_element(element, i)
+                    if non_lin_flux_vector is not None:
+                        if isinstance(element, (SetOffset, SetGain)):
+                            if isinstance(element, SetOffset) and element.bus in non_lin_offset_list:
+                                non_lin_offsets.append(non_lin_flux_vector.get_corrected_offsets())
+                                non_lin_offset_list = []
+                            elif isinstance(element, SetOffset):
+                                non_lin_offset_list.append(element.bus)
+                            value = element.gain if isinstance(element, SetGain) else element.offset_path0
```
flavie-lebars, line 509: Can they give a second offset (offset_path1)?

src/qililab/qprogram/qprogram.py
```diff
@@ -469,8 +469,17 @@ def with_crosstalk_qblox(self, crosstalk: CrosstalkMatrix):
         self._bus_variable_map: dict[tuple[Variable | float | None, str], Variable] = {}
         self._block_variables: dict[Variable | float | None, list[Variable]] = {}
         self._parallel_loops: dict[Variable, list[ForLoop]] = {}
-        self._active_loops: list[ForLoop] = []
+        self._active_loops: list[ForLoop | Parallel] = []
         self._loop_depths: list[int] = []
+        self._index_dim: int = 0
```
flavie-lebars: index_dim is dead code.

src/qililab/qprogram/qprogram.py
```diff
@@ -839,9 +901,119 @@ def make_for_loop(variable: Variable, bus: str, bias_vector: Sequence[float]) ->
 
             return ForLoop(variable=self._bus_variable_map[variable, bus], start=start, stop=stop, step=step)
 
+        def handle_non_linear(
```
flavie-lebars, line 901: Can you add all the typings and a docstring please

src/qililab/qprogram/qprogram.py
```diff
@@ -533,10 +577,28 @@ def traverse(
                         parallel_loop = Parallel(loops=loop_list)
                         parallel_loop.elements = element.elements
                         block.elements[i] = parallel_loop
-                        self._active_loops = self._active_loops[:-1]
-                        self._loop_depths = self._loop_depths[:-1]
+                    self._active_loops = self._active_loops[:-1]
```
flavie-lebars: is it normal that now this is no longer conditional?

src/qililab/qprogram/qprogram.py
```diff
@@ -839,9 +901,119 @@ def make_for_loop(variable: Variable, bus: str, bias_vector: Sequence[float]) ->
...
+        ):
+            corrected_elements: list[Block | Operation] = []
+            offsets_index = offsets_0
+            last_appended_offset = -1
+            plays_index = plays_0
+            play_bus_list: list[str] = []
+
+            for element in elements:
+                if isinstance(element, SetGain) and element.bus in flux_vector.buses:
+                    continue
```
flavie-lebars, line 954: is it normal that SetGain are ignored?

flavie-lebars (dismissed):
Great job!

- The compiler has not been modified but qprogram has, I do agree on adding tests to the compiler just in case, but in my opinion, the bulk of the tests should be checking that the input qprogram with crosstalk generates the good qprogram. The change from qprogram to Q1ASM is not what is being implemented here. It does test that the second qprogram is good, but it is not the right place in my opinion.

- Has the limitations regarding the memory been assessed? both number of lines and waveform memory.

- Also should the restart_flux_vector also be applied to the non linear?

src/qililab/qprogram/qprogram.py
jordivallsq: Yes, it is sort of merged with the play as the actual flux sent is `flux = envelope*gain + offset` and the non-linear function depends on the total flux, the offset is not much of a problem but gain is, because since the voltage is no longer linear this multiplication doesn't make any more sense.

Therefore the gain modifies the envelope and we no longer need to execute a SetGain

docs/releases/changelog-dev.md
jordivallsq: yes let me fix it, it must've been deleted in a merge

src/qililab/qprogram/qprogram.py
jordivallsq: removing it

src/qililab/qprogram/qprogram.py
jordivallsq: It doesn't make sense, as the flux lines have only one output

src/qililab/qprogram/qprogram.py
jordivallsq: added

src/qililab/qprogram/qprogram.py
jordivallsq: it has been reverted, those shouldn't be outside

jordivallsq (commented):
To Answer your questions @flavie-lebars:
- That is a good point, I will create the tests inside qprogram tests, but they will be very similar to the ones already there as the Q1ASM may be not what is creating the handle crosstalk but it is the "final" step.
- I have been testing the memory limitations but I need to sumarize and explain all of them in a notion page. As it was discussed, it was defined a final step after this PR. And thus I have treat them as such
- restart flux is useful for linear corrections, the `NonLinearFluxVector.exit_loop() `is doing essentially the same (closing the loop and getting the flux vector ready for the next loop.

src/qililab/qprogram/qprogram.py
```diff
@@ -839,9 +898,150 @@ def make_for_loop(variable: Variable, bus: str, bias_vector: Sequence[float]) ->
...
+                        element_copy.elements = corrected_loop
+                        corrected_elements.append(element_copy)
+                    if offset_defined:
+                        offsets_index += 1
+                        offset_defined = False
+                else:
+                    corrected_elements.append(element)
+
+            # Needs to sync at the end of every loop for the unpack to work with non-flux buses
+            if not isinstance(corrected_elements[-1], Sync):
```
DZPM: `corrected_elements` could be empty, `corrected_elements[-1]` would crash

jordivallsq: fixed

flavie-lebars (commented):
as I am off from today, I will remove my blocker to approve, @jjmartinezQT can you please check when it is ready?

docs/releases/changelog-dev.md
elygoner, line 137: Should this line be modified?

tests/qprogram/test_qprogram.py
```diff
@@ -439,6 +440,330 @@ def test_with_crosstalk_multi_variable_offset(self):
...
+        assert isinstance(new_qp.body.elements[5], SetGain)
+        assert math.isclose(new_qp.body.elements[5].gain, 0.3471177904070214)
+        assert new_qp.body.elements[5].bus == "flux1"
+        assert isinstance(new_qp.body.elements[6], SetOffset)
+        assert math.isclose(new_qp.body.elements[6].offset_path0, 0.0)
+        assert new_qp.body.elements[6].bus == "flux1"
+        assert isinstance(new_qp.body.elements[7], Play)
+        assert isinstance(new_qp.body.elements[7].waveform, Square)
+        assert new_qp.body.elements[7].bus == "flux1"
+        assert new_qp.body.elements[7].waveform.amplitude == 1.0  #Check that Square pulses are normalized
```
elygoner: SonarClaud: Do not perform equality checks with floating point values.

tests/qprogram/test_qprogram.py
```diff
@@ -439,6 +440,330 @@ def test_with_crosstalk_multi_variable_offset(self):
...
+        assert isinstance(new_qp.body.elements[8], SetGain)
+        assert new_qp.body.elements[8].bus == "flux2"
+        assert math.isclose(new_qp.body.elements[8].gain, 0.5442355808140428)
+        assert isinstance(new_qp.body.elements[9], SetOffset)
+        assert new_qp.body.elements[9].bus == "flux2"
+        assert new_qp.body.elements[9].offset_path0 == 0.0
```
elygoner: SonarClaud: Do not perform equality checks with floating point values.

tests/qprogram/test_qprogram.py
```diff
@@ -439,6 +440,330 @@ def test_with_crosstalk_multi_variable_offset(self):
...
+        assert isinstance(new_qp.body.elements[20], SetGain)
+        assert math.isclose(new_qp.body.elements[20].gain, 0.09888972435411081)
+        assert new_qp.body.elements[20].bus == "flux1"
+        assert isinstance(new_qp.body.elements[21], SetOffset)
+        assert math.isclose(new_qp.body.elements[21].offset_path0, 0.31843980232472535)
+        assert new_qp.body.elements[21].bus == "flux1"
+        assert isinstance(new_qp.body.elements[22], Play)
+        assert isinstance(new_qp.body.elements[22].waveform, Square)
+        assert new_qp.body.elements[22].bus == "flux1"
+        assert new_qp.body.elements[22].waveform.amplitude == 1.0  #Check that Square pulses are normalized
```
elygoner: SonarClaud: Do not perform equality checks with floating point values.

tests/qprogram/test_qprogram.py
```diff
@@ -439,6 +440,330 @@ def test_with_crosstalk_multi_variable_offset(self):
...
+        assert isinstance(new_qp.body.elements[10], SetGain)
+        assert math.isclose(new_qp.body.elements[10].gain, 0.08143146795214351)
+        assert new_qp.body.elements[10].bus == "flux1"
+        assert isinstance(new_qp.body.elements[11], SetOffset)
+        assert math.isclose(new_qp.body.elements[11].offset_path0, 0.3471177904070214)
+        assert new_qp.body.elements[11].bus == "flux1"
+        assert isinstance(new_qp.body.elements[12], Play)
+        assert isinstance(new_qp.body.elements[12].waveform, Square)
+        assert new_qp.body.elements[12].bus == "flux1"
+        assert new_qp.body.elements[12].waveform.amplitude == 1.0  #Check that Square pulses are normalized
```
elygoner: SonarClaud: Do not perform equality checks with floating point values.

elygoner (commented):
Good Job. I have left some comments.

docs/releases/changelog-dev.md
jordivallsq: this is related to another PR, in the PR for variable expression using Qblox I already reference this not implemented error. In this Changes this line is not related as while unpacking we do not do any Variable expression (there are no loops in the Q1ASM)

jordivallsq:
Tested non-linear crosstalk with Conscience WF02 3.6 chip in Mimas. Yifei helped me with the measurements and checked the results for junction asymmetry compensation.

This is the sample with only linear crosstalk compensation
imagen

(also with Lower resolution)
imagen

And we can see how applying the Junction asymmetry non-linear compensation (inside `NonLienearFluxVector`):
`phi_d = np.arctan(d * np.tan(np.pi * flux_x))`
`-phi_d / (2 * np.pi)`
for a d of -0.1:
imagen

a d of 0.5
imagen

and a d of 0.7
imagen

According to experimentalists, this rotation is the expected behavior. (Although d seems to be larger than expected. This should be addressed in `NonLienearFluxVector`)

tests/qprogram/test_qprogram.py
jordivallsq: solved

tests/qprogram/test_qprogram.py
jordivallsq: solved

tests/qprogram/test_qprogram.py
jordivallsq: solved

tests/qprogram/test_qprogram.py
jordivallsq: solved

elygoner (approved):
You have changed all what Flavie asked of you and Flavie is on vacation

PR #1119, opened 2026-05-26
Dependabot updates for idna

DZPM:
uv lock --upgrade-package "idna>=3.15"


elygoner (approved):
Checked imports and light execution

PR #1122, opened 2026-05-28
[QHC-1398][BUG] FlatTop envelope length mismatch.

elygoner:


tests/qprogram/test_qblox_compiler.py
```diff
@@ -949,7 +949,7 @@ def fixture_crosstalk_qprogram_parallel() -> QProgram:
 @pytest.fixture(name="crosstalk_qprogram_plays")
 def fixture_crosstalk_qprogram_plays() -> QProgram:
     square_wf = Square(amplitude=0.1, duration=50)
-    flattop_iq = FlatTop(amplitude=0.1, duration=49, smooth_duration=10)
+    flattop_iq = FlatTop(amplitude=0.1, duration=50, smooth_duration=10)
```
jordivallsq, line 952: Don't the rest of the tests with flattop fail as well? Those are tests decomposing the flattop envelop int an initial ramp, a set of identical square pulses and the final decreasing ramp.

These tests should also change in duration length

jordivallsq (commented):
The code implementation looks good to me, fix the coverage and test issues and test this on a qblox and we can merge it

src/qililab/waveforms/flat_top.py
```diff
@@ -52,13 +52,17 @@ def envelope(self, resolution: int = 1) -> np.ndarray:
         Returns:
             np.ndarray: Height of the envelope for each time step.
         """
-        x = np.arange(-self.duration / 2, self.duration / 2 + 1, resolution)
+        x = np.linspace(-self.duration / 2, self.duration / 2, self.duration // resolution)
```
DZPM: `linspace` distorts sample spacing. It includes both endpoints, so the step is `d/(N-1)`, not `resolution` (1.034 ns at dur=30, res=1; ~14% off at res=4). Every sample value shifts.

docs/releases/changelog-dev.md
```diff
@@ -11,3 +11,6 @@
 ### Documentation
 
 ### Bug fixes
+
+- Fixed a mismatch between `FlatTop`'s duration and `envelope`, were the waveform's `envelope` had length duration + 1.
```
DZPM, line 18:
```suggestion
- Fixed a mismatch between `FlatTop`'s duration and `envelope`, where the waveform's `envelope` had length duration + 1.
```

tests/waveforms/test_flat_top.py
```diff
@@ -20,20 +33,10 @@ def test_init(self, flat_top):
...
-        x = np.arange(-flat_top.duration / 2, flat_top.duration / 2 + 1, 2)
-        A = flat_top.amplitude
-        sigma = flat_top.smooth_duration
-        buf = flat_top.buffer
-        dur = flat_top.duration
-        envelope = (
-            0.5 * A * np.real((erf(4 * (x + (dur / 2 - buf)) / sigma - 2) - erf(4 * (x - (dur / 2 - buf)) / sigma + 2)))
-        )
-
-        assert np.allclose(flat_top.envelope(resolution=2), envelope)
```
DZPM, line 36: The old test did `resolution=2`. Not anymore, we're doing only `resolution=1`, why?

Does it make sense to add more resolutions?

DZPM:
Given that the title is "FlatTop envelope length mismatch", should we have a specific test on envelope length?

tests/qprogram/test_qblox_compiler.py
```diff
@@ -949,7 +949,7 @@ def fixture_crosstalk_qprogram_parallel() -> QProgram:
 @pytest.fixture(name="crosstalk_qprogram_plays")
 def fixture_crosstalk_qprogram_plays() -> QProgram:
     square_wf = Square(amplitude=0.1, duration=50)
-    flattop_iq = FlatTop(amplitude=0.1, duration=49, smooth_duration=10)
```
DZPM, line 952: Why the duration change, from `49` to `50`?

tests/qprogram/test_qblox_compiler.py
elygoner: the resulting compilation output it compares again has.

src/qililab/waveforms/flat_top.py
elygoner: thanks for the tip, I have gone back to the arrange, I geniuently didn't know that

tests/waveforms/test_flat_top.py
elygoner: QBlox and Qdac both use resolution 1 so I don't think so

src/qililab/waveforms/flat_top.py
DZPM: It was sus, asked Claude for the differences, and it did the rest :+1:

tests/qprogram/test_qblox_compiler.py
elygoner: If you look at the rest of the waveforms, they are 50. FlatTop was 49 because of the missmatch in length, since before the bugfix it was 1 longer. This indirectly checks the length of the FlatTop.

elygoner: There isn't a specific test on the envelope length, but there is a check on the envelope output on resolution 1 that not explicitly checks for that. But I'll add a check thare.

jordivallsq (approved):
Well done! Resolve David's comments and share the Oscilloscope image of the correct behavior and we can merge

elygoner:
The pr has been checked on the hardware.
image
RigolDS2ctflattop0

DZPM (approved):
Fix the "where" typo and we're good

PR #1123, opened 2026-06-01
Update qilisdk dependency

elygoner:
I updated the qilisdk dependency to a more recent version.


pyproject.toml
```diff
@@ -79,7 +79,7 @@ prerelease = "if-necessary"
 override-dependencies = ["marshmallow==3.26.2"]
 
 [tool.uv.sources]
-qilisdk = { git = "https://github.com/qilimanjaro-tech/qilisdk.git", branch = "main" }
+qilisdk = { git = "https://github.com/qilimanjaro-tech/qilisdk.git", tag = "0.2.0" }
```
DZPM: This means we'll always use this tag in the future. If QiliSDK main breaks, won't notice until we update the tag 👍

DZPM (approved):

fedonman:
@elygoner You probably need to check how Rmw was patched for latest QiliSDK in qili-runtime and make the same changes in your PR: https://github.com/qilimanjaro-tech/qili-runtime/blob/main/src/qili_runtime/digital/gates/gates.py

PR #1125, opened 2026-06-02
Qhc 1403 bug fix crosstalk matrix to array fails to fill with zeroes

jordivallsq:
Fixed a bug where `CrosstalkMatrix.to_array()` crashed instead of creating the right array when some field of the matrix dictionary where missing (missing values of the matrix), this should default to the identity (values of 1 in the diagonal and 0 outside the diagonal). Now it creates an array even if there are missing elements of the matrix and defaults them to 0 or 1 if those elements are in the diagonal of the matrix.
Also fixed the str representation of the same crosstalk matrix as the same missing values were represented with ones instead of zeroes.

### DIFF
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -11,3 +11,7 @@
 ### Documentation
 
 ### Bug fixes
+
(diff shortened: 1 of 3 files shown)


elygoner (approved):
I don't know if there is a circumstance where eye could cause problems. Appart from that LGTM as long as you have tested it in hardware

jordivallsq:
The eye emulates the mutual inductance, where the default is 1 instead of 0. This has been consulted to Mireia.

PR #1132, opened 2026-06-08
Bug fix load sequence

OscarQili:
fixed ordering of returned measurements when calling load_sequence_by_id(). It previously had an arbitrary order


DZPM (approved):

src/qililab/result/database/database_manager.py
```diff
@@ -252,7 +252,12 @@ def load_sequence_by_id(self, id: int | list[int]) -> list[Measurement] | None:
...
         """
         with self.session() as running_session:
-            measurement_by_id_list = running_session.query(Measurement).where(Measurement.sequence_id == id).all()
+            measurement_by_id_list = (
+                running_session.query(Measurement)
+                .where(Measurement.sequence_id == id)
+                .order_by(Measurement.measurement_id)
+                .all()
+            )
             if measurement_by_id_list is not None:
```
DZPM:
```suggestion
            if measurement_by_id_list != []:
```

The `measurement_by_id_list` will never be `None`.
The `.all()` could return an empty list `[]`, but never `None`.

DZPM: Let's better remove the line: it does not make sense, if it's empty, the `for meas in measurement_by_id_list:` will just get skipped

DZPM (approved):
Approved, but consider fixing the "dead" `if` line

PR #1134, opened 2026-06-09
 upgrade qblox-instruments dependency

flavie-lebars:
Upgraded `qblox-instruments` dependency from `0.16.0` to `1.0.3`.

### DIFF
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -4,6 +4,9 @@
 
 ### Improvements
 
+- Upgraded `qblox-instruments` dependency from `0.16.0` to `1.0.3`.
+  [#1134](https://github.com/qilimanjaro-tech/qililab/pull/1134)
+
 ### Breaking changes
 
 ### Deprecations / Removals
--- a/uv.lock
+++ b/uv.lock
@@ -586,6 +586,15 @@ wheels = [
 ]
 
 [[package]]
+name = "configparser"
+version = "7.2.0"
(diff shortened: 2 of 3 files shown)


fedonman (approved):
🚀 🚀

uv.lock
```diff
@@ -1481,7 +1490,7 @@ name = "intel-cmplr-lib-ur"
 version = "2025.3.1"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "umf", marker = "platform_machine != 'aarch64' and sys_platform != 'darwin'" },
```
DZPM: Was aarch64 dropped on purpose? Assuming it's qblox-instruments.

Is it ok to drop it?

docs/releases/changelog-dev.md
```diff
@@ -4,6 +4,9 @@
 
 ### Improvements
 
+- Upgraded `qblox-instruments` dependency from `0.16.0` to `1.0.3`.
```
DZPM, line 7: Should make explicit that it's a breaking change

DZPM (approved):

docs/releases/changelog-dev.md
flavie-lebars: I don't think it's a breaking change, it is just a requirement update.

uv.lock
flavie-lebars: yes I assume it's qblox, it was done automatically when doing uv lock

PR #1135, opened 2026-06-10
[QHC-1393] Pin Spirack version to avoid reported install/import issues

elygoner:
The pined version is 0.2.12 because is compatible with all the versions of qblox-instruments we are using. Also It is the version pined currently in the latest version of qblox-intruments: https://gitlab.com/qblox/packages/software/qblox_instruments/-/blob/main/pyproject.toml?ref_type=heads

### DIFF
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -14,3 +14,6 @@
 
 - Fixed a bug in `load_sequence_by_id` where measurements were returned in an undefined order, causing incorrect results when processing sequences. Measurements are now ordered by `measurement_id`.
   [#1132](https://github.com/qilimanjaro-tech/qililab/pull/1132)
+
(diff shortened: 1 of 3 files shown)


DZPM (approved):

fedonman (approved):
Tip: If you don't want to explicitly ask for a dependency (`qililab` doesn't need `spirack`, `qblox-instruments` does) and just want to force a version, you can use `override-dependencies` setting of `uv`: https://docs.astral.sh/uv/reference/settings/#override-dependencies

elygoner:
> Tip: If you don't want to explicitly ask for a dependency (`qililab` doesn't need `spirack`, `qblox-instruments` does) and just want to force a version, you can use `override-dependencies` setting of `uv`: https://docs.astral.sh/uv/reference/settings/#override-dependencies

Good tip. I have applied it and checked everything again.

fedonman (approved):
💯 💯

PR #1136, opened 2026-06-10
QHC 1312 Added bus mapping to measurement database columns for live plot drawing

jordivallsq:
Added bus_mapping to measurement database table `Measurement`. Bus mapping is necessary for live plot drawing of the qprogram and it has been information missing in the database. StreamArray already has the bus_mapping as an input, this input is the dictionary that will be saved in the database.


elygoner (approved):
LGTM, but the PR name is not descriptive of the changes, please correct it.

jordivallsq:
Added bus mapping inside database columns as an optional parameter and tested that it works
imagen

And that main branch still works while leaving the field empty
imagen

Tests for this PR can be found in QHC_hardware tests: https://github.com/qilimanjaro-tech/QHC_hardware_tests/blob/main/tests/qililab/test_PR_1136.ipynb

PR #1139, opened 2026-06-11
[QHC-1413] Silence QCoDeSDeprecationWarning on import of QiliLab

elygoner:


flavie-lebars (changes requested):
thanks for the fix! Please add a changelog.

src/qililab/_suppress_warnings.py
```diff
@@ -0,0 +1,19 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Ignore QCodes deprecation warnings from the spirack library
+import warnings
+
+_ = warnings.filterwarnings("ignore", module=r"qblox_instruments\.qcodes_drivers\.spi_rack")
```
flavie-lebars: Please fix the sonarcloud issue here

src/qililab/__init__.py
```diff
@@ -82,6 +83,7 @@
     "SuddenNetZero",
     "Waveform",
     "__version__",
+    "_suppress_warnings",
```
flavie-lebars, line 86: does it need to be here?

flavie-lebars (changes requested):

src/qililab/__init__.py
elygoner: yes if not ruff will complain

elygoner:
Newer versions of qblox_intruments stopped having this issue so we decided to wait until we can just use that newer version.

PR #1140, opened 2026-06-12
Changed to version 0.33.2

jordivallsq:


elygoner (approved):
It sure does look like a version change.

PR #1116, opened 2026-05-19
[SDK-284] Removing QiliSDK dependency 

amirazzam2000:

PR #1142, opened 2026-06-15
fix(security): QILILAB-01/02 safe-by-default YAML deserialization

DZPM:
## Summary

Closes **QILILAB-01** and **QILILAB-02** (High, CWE-502) - the qilisdk unsafe-loader root cause, reached locally in qililab.

Findings: [QILILAB-01](https://app.notion.com/p/37d7eec14c538123aac7ff474db1b0f9) · [QILILAB-02](https://app.notion.com/p/37d7eec14c5381c984d4e2ac9c696d7d)

qililab does its own serialization on `from qilisdk.yaml import yaml` (the `typ="unsafe"` instance) and calls `yaml.load` directly in `deserialize` / `deserialize_from` (`utils/serialization.py:91,121`), so loading an untrusted calibration file or runcard executes arbitrary code (`!!python/object/apply`, `!function` / `!lambda` via dill, `!type` / `!PydanticModel` / `!defaultdict` via `__import__`); the `isinstance(result, cls)` guard runs *after* execution. **QILILAB-02** reaches the same sink through `GateEvent._load_waveform` / `_load_weights` during `build_platform`. qililab can't just wait for the upstream QSDK-01 fix (unreleased, and qililab bypasses qilisdk's safe path).

## Fix

New `src/qililab/utils/safe_yaml.py` - a **registry-isolated, data-only** loader built lazily after qililab's `@register_class` decorators populate the registry: it gives its constructor fresh per-instance `yaml_constructors` / `yaml_multi_constructors` (ruamel's `add_constructor` mutates a shared class registry, so a plain second instance is *not* isolated), seeds only side-effect-free data/class tags, **rejects `!function` / `!lambda`**, gates `!type` / `!PydanticModel` / `!defaultdict` + `!!python/*` behind an (empty) FQN allow-list, and **raises on any unknown tag**. `deserialize` / `deserialize_from` route through it by default, with an additive default-off `trust_code=True` escape hatch. Closes QILILAB-02 at the same layer.

Verified closed 3/3; targeted tests pass, ruff + mypy clean. The higher-risk direct-construction rewrite of `_from_external` was deliberately deferred as optional defense-in-depth.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,11 +25,6 @@
 
 ## 0.33.1
 
-### Security fixes
-
-- Made `deserialize` / `deserialize_from` safe by default. They now load through a registry-isolated, data-only YAML loader that rejects code-execution tags (`!!python/object/apply`, `!!python/name`, `!function`, `!lambda`, and the gated `!type` / `!PydanticModel` / `!defaultdict`), so deserializing an untrusted string or file can no longer execute arbitrary code. Every legitimate round-trip (waveforms, `QProgram`, `Experiment`, `Calibration`, numpy arrays, tuples, complex numbers, UUIDs, enums) keeps working, and a `trust_code=True` opt-in restores the previous behavior for fully trusted input. This also closes the `build_platform` runcard remote-code-execution path, where a gate-event waveform/weights string reached the unsafe loader via the `GateEvent` validators.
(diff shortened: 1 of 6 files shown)


flavie-lebars (changes requested):
There is a pending PR to remove the qilisdk dependency on qililab, I would recommend not touching the yaml files.
https://github.com/qilimanjaro-tech/qililab/pull/1116

PR #1143, opened 2026-06-15
fix(security): QILILAB-03 bind live-plot Dash server to loopback by default

DZPM:
## Summary

Closes **QILILAB-03** (Medium, CWE-1327).

Finding: https://app.notion.com/p/37d7eec14c5381fe9c0dd73d9e3bfb15

`ExperimentLivePlot` started its Dash server with a hardcoded `host="0.0.0.0"` and no auth, and put the **absolute HDF5 result-file path** in the page title. Anyone on the lab LAN reaching the port (default 8050) saw in-progress measurements + an internal path. Gated behind two opt-in settings (both default `False`), hence Medium.

## Fix

Mirrors the existing `experiment_live_plot_port` plumbing end-to-end:

- **Loopback by default** - new `experiment_live_plot_host` setting defaults to `127.0.0.1`; `QILILAB_EXPERIMENT_LIVE_PLOT_HOST` is the explicit opt-in for the SLURM cross-node case (which is why the old "127.0.0.1 does not work" TODO existed). Its description documents the no-auth caveat + reverse-proxy requirement for any routable bind.
- Threaded through the constructor to the bind site `self._dash_app.run(debug=False, host=self._host, ...)`.
- **Removed `# noqa: S104`** so ruff's bind-all-interfaces lint re-arms against a future regression.
- **Path disclosure**: the title now uses `os.path.basename(self.path)` (no absolute directory tree).

Verified closed 3/3; 12 tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
 
 ### Bug fixes
 
+- Hardened the live-plot Dash server: it now binds to loopback (`127.0.0.1`) by default instead of all interfaces. Routable exposure (for cross-node SLURM viewing) is opt-in via `QILILAB_EXPERIMENT_LIVE_PLOT_HOST` and must be fronted by an authenticated reverse proxy, since the server has no authentication of its own. The absolute result-file path was also removed from the live-plot page title.
+  [#1141](https://github.com/qilimanjaro-tech/qililab/pull/1141)
+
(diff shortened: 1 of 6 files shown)


flavie-lebars (dismissed):
We should never modify the actual changelog, it should be the dev one.
I will add @jordivallsq as a reviewer because he is the most familiar with the live plot.

DZPM:
Good catch, thanks @flavie-lebars - fixed. Moved the entry to `docs/releases/changelog-dev.md` and restored `CHANGELOG.md` untouched (also corrected the PR reference to #1143). Applied the same correction to the sibling security PR #1142.

PR #1146, opened 2026-06-17
[QHC-1415] Create cache system for QBlox execution to avoid full re-uplode of the sequencer

elygoner:
The workflow in `Platform` proposed is:
- Hash Sequence
- Compare with platform cache
  - If no sequence previously uploaded -> all cache updated and sequencer fully uploaded.
  - If the program, weights or waveforms changed -> only the changed get updated on the cache and them along with acquisitions get updated in the sequencer.
  - If nothing different -> only the acquisitions get updated.

On `Bus.upload_qpysequence` added an additional parameter called "components_to_update". If the parameter is not None/empty dict, It re-routes it to `QbloxModule.update_sequencer`. It is worth mentioning that the `Bus` and  `QbloxModule` are general (in the sense that do not have any special logic), in case we want to reuse it in the future.

`QbloxModule.update_sequencer` gets the "components_to_update" that are True and updates them using ``qblox_instruments.qcodes_drivers.sequencer.Sequencer.update_sequence``.

On the tests side, I have moved all the tests for `QbloxModule` to test_qblox_module.py. And I have changed some tests In the QRM.


docs/releases/changelog-dev.md
```diff
@@ -4,6 +4,13 @@
 
 ### Improvements
 
+- Upgraded `qblox-instruments` dependency from `0.16.0` to `1.0.3`.
```
flavie-lebars: this bump should not appear in this changelog

docs/releases/changelog-dev.md
```diff
@@ -4,6 +4,13 @@
...
+  [#1134](https://github.com/qilimanjaro-tech/qililab/pull/1134)
+
+- Optimized sequence uploads in `Platform.execute_qprogram` and `Platform.execute_qprograms_parallel`. When any components of the `QProgram` uploaded to a Qblox sequencer are unchanged from the previous execution, only the components that actually changed are re-uploaded: program, waveforms and weights only when they differ, and acquisitions always (to reset the bins). On the first upload for a bus, the full sequence is uploaded. This avoids redundant re-uploads and speeds up software loops that repeatedly run the same program while only changing an instrument setting (`Platform.set_parameter`) or waveform data.
```
flavie-lebars: I would be careful mentioning waveform data. If the duration changes, the Q1ASM changes.

flavie-lebars: Since we do still have some clusters at BSC 0.11, could you please sepcify that this requires qblox firmware >= 0.13

src/qililab/instruments/qblox/qblox_module.py
```diff
@@ -201,12 +201,14 @@ def module_type(self):
         """returns the qblox module type. Options: QCM or QRM"""
         return self.device.module_type()
 
-    def run(self, channel_id: ChannelID):
+    def _have_sequencer(self, channel_id: ChannelID):
```
flavie-lebars: return type is missing, please always include this. I am only commenting this line, but please fix this throughout the PR.

src/qililab/instruments/qblox/qblox_module.py
```diff
@@ -647,32 +649,45 @@ def reset(self):
...
         self.device.reset()
 
-    def upload_qpysequence(self, qpysequence: QpySequence, channel_id: ChannelID):
+    def upload_qpysequence(self, qpysequence: QpySequence, channel_id: int):
         """Upload the qpysequence to its corresponding sequencer.
 
         Args:
             qpysequence (QpySequence): The qpysequence to upload.
-            port (str): The port of the sequencer to upload to.
+            channel_id (int, str): Identifier for the channel and it's corresponding sequencer.
```
flavie-lebars: typo: its
You have it in a few places, I haven't commented all of them.

src/qililab/instruments/qblox/qblox_module.py
```diff
@@ -647,32 +649,45 @@ def reset(self):
...
-            self.sequences[sequencer.identifier] = qpysequence
+            self.device.sequencers[channel_id].sequence(qpysequence.todict())
+            self.sequences[channel_id] = qpysequence
+
+    def update_sequencer(self, qpysequence: QpySequence, channel_id: int, **components_to_update: bool):
+        """Updates the last sequencer with the current qpysequence if the program hasn't changed.
+
+        Args:
+            qpysequence (QpySequence): The qpysequence to update.
+            channel_id (int, str): Identifier for the channel and it's corresponding sequencer.
```
flavie-lebars: in the docstring, it says channel_id can be a str but not in the fn declaration

src/qililab/instruments/qblox/qblox_module.py
```diff
@@ -647,32 +649,45 @@ def reset(self):
...
+            **components_to_update (bool): Flags describing which components are getting updated. Accepts any of:
```
flavie-lebars: the docstring and typings are wrong for components_to_update

flavie-lebars (changes requested):
- If the program changes, evthg needs to be reuploaded. Becausee the order in which the weights/acquisitions appear in the Q1ASM may change.

- I think it could be nice to quantify how much time this PR saves. Could you please create a test file (in the new repo Juanjo created); and compare how much time qililab takes for in main and with this PR. And please also link the test file to this PR in some way, so that when I review I can check that no hw test is missing.

- Also on QBlox documentation I just saw they have `Cluster.update_sequences` (update multiple sequences at once), this could be another way of speeding up software loops & general execution, I have created a Linear task to explore this https://linear.app/qilimanjaro/issue/QHC-1486/explore-the-use-of-clusterupdate-sequences

- I have asked Claude to review the code, I recommend always doing this before submitting a PR for approval. I have not checked all of them in depth, feel free to ignore what you think is irrelevant.

```
{
  "level": "xhigh",
  "findings": [
    {
      "file": "src/qililab/platform/platform.py",
      "line": 1835,
      "category": "correctness",
      "summary": "_compute_components_to_update() writes the new hash into self._qpy_sequence_cache before the corresponding upload to hardware is confirmed to succeed.",
      "failure_scenario": "buses[bus_alias].upload_qpysequence(...) raises (comms error, validation error, non-TimeoutError exception) right after the cache was already updated with the new hash. A retry or the next execute_qprogram() call with the same sequence sees cached_hashes == new_hashes, concludes nothing changed, and skips re-uploading program/waveforms/weights that never actually reached the instrument — the device silently keeps running stale/absent content.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/qililab/platform/platform.py",
      "line": 391,
      "category": "correctness",
      "summary": "Platform._qpy_sequence_cache is only ever cleared in disconnect(); initial_setup(), and QbloxModule.reset()/clear_cache(), never invalidate it even though QbloxQRM.initial_setup() wipes the real device sequencer content to empty.",
      "failure_scenario": "connect() -> execute_qprogram(qp) (full upload, cache populated) -> platform.initial_setup() (device docstring recommends doing this 'every time'; QbloxQRM.initial_setup() pushes {program:'', waveforms:{}, weights:{}, acquisitions:{}} to every ADC sequencer) -> execute_qprogram(qp) again with the identical program. The stale cache reports no change, so only 'acquisitions' gets pushed via update_sequencer — the program/waveforms/weights just wiped by initial_setup are never re-uploaded, and the sequencer silently runs empty.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/qililab/instruments/qblox/qblox_module.py",
      "line": 664,
      "category": "correctness",
      "summary": "update_sequencer() (the new partial-upload path) never writes back into self.sequences[channel_id], unlike upload_qpysequence(), which does.",
      "failure_scenario": "After any call that takes the partial-update path, self.sequences[channel_id] keeps pointing at old/absent content. run() gates arming/starting on `channel_id in self.sequences` (line 209) and can silently no-op; the legacy upload()/Bus.upload() path reads self.sequences[channel_id] and re-pushes stale content to hardware; QbloxQRM.acquire_qprogram_results() gates on `sequencer.identifier in self.sequences` and can silently drop acquisition results.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/qililab/instruments/qblox/qblox_qrm.py",
      "line": 177,
      "category": "correctness",
      "summary": "acquire_qprogram_results() references the loop variable `sequencer` after a for-loop over acquisitions.items(), with no assignment to `sequencer` anywhere before the loop.",
      "failure_scenario": "If `acquisitions` is an empty dict for a given bus/channel in a compiled run, the for-loop body (line 153-154, the only place `sequencer` is assigned) never executes. The post-loop `if sequencer is not None:` at line 177 then raises UnboundLocalError, crashing execute_qprogram() instead of skipping the intended cleanup call.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/qililab/platform/platform.py",
      "line": 1836,
      "category": "correctness",
      "summary": "Commit 8b64beac removed the branch that forced a full re-upload (return None) whenever the program hash changed; the surviving test's docstring still claims this safety behavior while its own assertion checks a partial-update dict.",
      "failure_scenario": "Previously any program change forced a full re-upload as a safety measure against waveform/weight index remapping. Now a program-only change just sets program=True and still goes through the partial update_sequencer path. tests/platform/test_platform.py:1220-1221 still says 'must force a full re-upload (None)' in its docstring but asserts a partial dict — so no test in the suite actually guards the scenario the removed code was protecting against.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/qililab/platform/platform.py",
      "line": 1822,
      "category": "correctness",
      "summary": "Platform._qpy_sequence_cache is keyed by logical bus_alias, not by the physical (instrument, channel_id) pair; no validation prevents two bus aliases from pointing at the same physical target.",
      "failure_scenario": "BusSettings/Runcard perform no uniqueness check on (instrument, channel) across bus aliases. If two aliases share a physical sequencer, executing via alias A caches hash(seq_A) under key 'A' and uploads; executing via alias B overwrites the actual hardware but caches under key 'B'. Re-running alias A later compares against its own stale entry, wrongly reports 'no change', and skips a needed re-upload — leaving the device running B's content while qililab believes A's is still active.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "tests/platform/test_platform.py",
      "line": 2220,
      "category": "test-coverage",
      "summary": "A test patches Platform._compute_components_to_update via a bare pytest.MonkeyPatch() instance instead of the monkeypatch fixture, and never calls .undo().",
      "failure_scenario": "Because the patch targets the Platform class itself and is never reverted, every Platform instance created for the rest of the pytest process silently gets _compute_components_to_update forced to return None (full-upload behavior). Any later test asserting on the real hash-based caching/components_to_update logic added by this PR would be masked and could pass even if that logic regresses.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "tests/instruments/qblox/test_qblox_module.py",
      "line": 21,
      "category": "test-coverage",
      "summary": "A fixture mutates the shared, class-level InstrumentFactory.handlers registry (adding \"QBloxModule\") and never removes the entry afterward.",
      "failure_scenario": "The fixture returns directly with no yield/teardown, so the registration leaks for the rest of the pytest session. Currently benign since no other test uses that name, but any future test checking that an unregistered instrument name raises a lookup error would silently stop failing if it happened to use \"QBloxModule\", with behavior depending on test execution order.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/qililab/instruments/qblox/qblox_qrm.py",
      "line": 178,
      "category": "reuse",
      "summary": "acquire_qprogram_results() calls the low-level device driver directly (device.sequencers[...].update_sequence(acquisitions={}, erase_existing=True)) instead of routing through the new QbloxModule.update_sequencer() wrapper added in this same PR.",
      "failure_scenario": "Two divergent code paths now exist for 'push a partial sequence update to hardware' within the same class hierarchy, added in the same change. Any future fix or safety check added to update_sequencer() (e.g. around self.sequences bookkeeping) won't apply to this call site.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/qililab/utils/hashing.py",
      "line": 63,
      "category": "altitude",
      "summary": "hashing.py reaches into private attributes of the third-party qpysequence package (_program, _waveforms._waveforms, _weights._weights) instead of the public todict() API used elsewhere in the codebase.",
      "failure_scenario": "pyproject.toml pins qpysequence as an open-ended \">=0.10.8\" with no upper bound. If a future qpysequence release renames or restructures these private attributes, hashing.py breaks (AttributeError or silently wrong hash) while upload_qpysequence/update_sequencer keep working since they use .todict(). Note: qpysequence itself is a private package qililab currently depends on via an SSH deploy-key stopgap, which increases exposure to unreviewed internal changes.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/qililab/instruments/qblox/qblox_module.py",
      "line": 675,
      "category": "efficiency",
      "summary": "update_sequencer() calls qpysequence.todict() (full serialization) unconditionally before filtering down to the components flagged as changed.",
      "failure_scenario": "During a parameter sweep where only e.g. gain changes (program/waveforms/weights unchanged per the hash cache), every iteration still pays the full CPU cost of serializing the entire sequence (stringifying the program, building full waveform/weight dicts) before discarding the unchanged parts — the cache only ends up saving device I/O, not the local work it was meant to avoid.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/qililab/platform/platform.py",
      "line": 1415,
      "category": "simplification",
      "summary": "The compute-then-upload-then-sync sequence (_compute_components_to_update + upload_qpysequence + sync_sequencer loop) is duplicated near-verbatim in two places: _execute_qblox_compilation_output and _upload_qblox_parallel_sequences.",
      "failure_scenario": "Both call sites were changed in this same PR to add identical caching/sync logic, differing only in how the Bus object is looked up. Any future fix to this logic (e.g. the cache-poisoning-on-failure bug above) must be found and applied in two places, and future edits risk drifting out of sync.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/qililab/utils/hashing.py",
      "line": 44,
      "category": "reuse",
      "summary": "_hash_str() reimplements the exact md5-hex-digest idiom already used by hash_qua_program() in quantum_machines_cluster.py.",
      "failure_scenario": "Both do hashlib.md5(<str>.encode('utf-8'), usedforsecurity=False).hexdigest() for the same conceptual purpose (change-detection hash of a serialized program string). hash_qua_program could delegate to _hash_str once it has its program string, but instead reimplements the hashing call, leaving the same convention defined twice with no shared source of truth.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "src/qililab/instruments/qblox/qblox_module.py",
      "line": 665,
      "category": "simplification",
      "summary": "update_sequencer()'s docstring (\"Updates the last sequencer with the current qpysequence if the program hasn't changed\") is stale now that commit 8b64beac removed the branch that guaranteed this.",
      "failure_scenario": "update_sequencer is now reachable via Bus.upload_qpysequence whenever components_to_update is not None, including when program=True (program did change). A reader trusting the docstring would wrongly assume program changes always take the full-upload path.",
      "verdict": "CONFIRMED"
    },
    {
      "file": "tests/test.yml",
      "line": 1,
      "category": "hygiene",
      "summary": "A 786-line generated YAML file appears to have been accidentally committed as a test-output artifact rather than an intentional fixture.",
      "failure_scenario": "tests/test_data_management.py:38-39 runs save_platform(path=\"tests/test.yml\", ...) as part of a test, which writes exactly this kind of file to disk. The committed file's content (a runcard named 'galadriel') matches that test's output, suggesting a dev ran the test locally and `git add`ed the resulting artifact by mistake — it serves no purpose checked into source control.",
      "verdict": "CONFIRMED"
    }
  ]
}
```

flavie-lebars:
@elygoner Any bugs you fix from now in this implementation, make sure to add a regression test please.

docs/releases/changelog-dev.md
elygoner: So if we merge this, you won't be able to use main with those clusters. Should be fine since they use releases, but if there is a release after this, the clusters at BSC have to be upgraded.

PR #1148, opened 2026-06-24
[Bugfix] ExperimentExecutor with Acquire and MeasureReset

fedonman:
Fixed `ExperimentExecutor` not allocating result datasets for `Acquire` (`qp.qblox.acquire`) and `MeasureReset` operations. Previously only `Measure` operations were counted as measurements, so a QProgram that read out via `qp.qblox.acquire` produced no result datasets and `ExperimentResults.get()` raised `KeyError`. The executor now counts the same `(Acquire, Measure, MeasureReset)` set as the `QbloxCompiler`.

### DIFF
--- a/tests/qprogram/test_experiment_executor.py
+++ b/tests/qprogram/test_experiment_executor.py
@@ -144,6 +144,18 @@ def fixture_experiment(qprogram: QProgram, crosstalk: CrosstalkMatrix):
     return experiment
 
 
+def make_platform_returning(qprogram_results: QProgramResults):
+    """Build a mock Platform whose ``execute_qprogram`` returns the given results."""
+    platform = create_autospec(Platform)
+    platform.set_parameter = Mock()
+    platform.get_parameter = Mock(return_value=1.23)
+    platform.execute_qprogram = Mock(return_value=qprogram_results)
+    platform.to_dict = Mock(return_value={"name": "platform"})
+    platform.set_crosstalk = Mock()
+    platform.db_manager = Mock()
+    return platform
+
+
 class TestExperimentExecutor:
     """Test ExperimentExecutor class"""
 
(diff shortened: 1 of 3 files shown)


tests/qprogram/test_experiment_executor.py
```diff
@@ -282,6 +294,77 @@ def test_execute(self, platform, experiment, qprogram, crosstalk, override_setti
...
+        frequency = qp.variable(label="frequency", domain=Domain.Frequency)
+        with qp.for_loop(frequency, 0, 10, 1):  # 11 points
+            qp.qblox.acquire(bus="readout_bus", weights=IQPair(Square(1.0, 100), Square(1.0, 100)))
+
+        experiment = Experiment(label="acquire_experiment")
+        experiment.execute_qprogram(qp)
+
+        qprogram_results = QProgramResults()
+        qprogram_results.append_result(
+            "readout_bus", QuantumMachinesMeasurementResult(bus="readout", I=np.arange(0, 11), Q=np.arange(100, 111))
```
jjmartinezQT: I have picky troubles reconciling the fact that the bug fix is about qblox but the results block is QuantumMachinesMeasurementResult. Very confusing.

jjmartinezQT (changes requested):
Hardware test is needed as the experiment this attempt to fix involves complex operations with triggers.

tests/qprogram/test_experiment_executor.py
fedonman: It does not really matter, since both `QbloxMeasurementResult` and `QuantumMachinesMeasurementResult` are children of `MeasurementResult`, and they are not under test; they are fixtures. The difference is that the QM one is easier to construct, because for Qblox you need to provide the raw dictionary. Nevertheless, I changed it to use Qblox results.

fedonman:
> Hardware test is needed as the experiment this attempt to fix involves complex operations with triggers.

Triggers are not touched in this PR. We should not confuse the two issues. I would like to merge this, propagate the merge commit to the EuroQCS branch, and update `qili-runtime` to use it, so we can test it in the experiment.

jjmartinezQT:
> > Hardware test is needed as the experiment this attempt to fix involves complex operations with triggers.
> 
> Triggers are not touched in this PR. We should not confuse the two issues. I would like to merge this, propagate the merge commit to the EuroQCS branch, and update `qili-runtime` to use it, so we can test it in the experiment.

I know triggers are not touched, but it is one of the couple of experiments we should be testing in hardware, along active reset. If we are in a hurry because this needs to go in this weeks delivery I can merge it, if not there should not be such a rush no? I will ask Natalia for confirmation about the priority.

jjmartinezQT:
Tested on real hardware using the following notebook, please have a look and if you think the tests covered the use cases we are good to go with this PR: 

https://github.com/qilimanjaro-tech/LabScripts/blob/main/Personal_folders/Juanjo/PRs_testing/test_PR_1148.ipynb

fedonman:
> Tested on real hardware using the following notebook, please have a look and if you think the tests covered the use cases we are good to go with this PR:
> 
> https://github.com/qilimanjaro-tech/LabScripts/blob/main/Personal_folders/Juanjo/PRs_testing/test_PR_1148.ipynb

I think the tests covered all the cases. 👏 I really like this format for hardware testing, by the way. Congrats!

jjmartinezQT (approved):
Fast and great work, approved!

PR #1151, opened 2026-06-30
qhc-1187-integration-length-fix

flavie-lebars:
[QHC-1187](https://linear.app/qilimanjaro/issue/QHC-1187)


src/qililab/instruments/qblox/qblox_qrm.py
```diff
@@ -77,7 +79,9 @@ def is_adc(self) -> bool:
 
     @check_device_initialized
     def initial_setup(self):
-        """Initial setup"""
+        """Initial setup
+        Device threshold is not set here: the integration_length is only known
+        once a QProgram is compiled. It is applied in the platform execution flow."""
```
elygoner: unnecessary addition to the docstring.

fedonman (approved):
Sorry for the long wait :) Looks good, thank you Flavie!

src/qililab/platform/platform.py
```diff
@@ -985,14 +987,18 @@ def session(self):
...
+            self.initial_setup()
 
             self.turn_on_instruments()
-            cleanup_methods.append(self.turn_off_instruments)  # Store turn_off_instruments for cleanup
+            # Store turn_off_instruments for cleanup
+            cleanup_methods.append(self.turn_off_instruments)
 
-            yield  # Experiment logic goes here
+            # Experiment logic goes here
+            yield
```
elygoner: There are a lot of comment on a different line, is this a new ruff rule that I am not aware of. I saw it flagged by SonarCloude

src/qililab/instruments/qblox/qblox_adc_sequencer.py
```diff
@@ -23,17 +24,26 @@
...
     threshold: float
     threshold_rotation: float
     time_of_flight: int  # nanoseconds
+    integration_length: int | None = None
     timeout_repetitions: int = 0
 
     def __post_init__(self):
         cast_enum_fields(obj=self)
+        if self.integration_length is not None:
+            warnings.warn(
```
jordivallsq, line 44: Why not using logger.warning?

jordivallsq: We currently have the code split between logger and warnings.warn so I propose to change all of them to logger.warning in a separate task (without much priority)

flavie-lebars: yes I agree

src/qililab/instruments/qblox/qblox_adc_sequencer.py
```diff
@@ -23,17 +24,26 @@
 class QbloxADCSequencer(QbloxSequencer):
     scope_acquire_trigger_mode: AcquireTriggerMode
     scope_hardware_averaging: bool
-    sampling_rate: float  # default sampling rate for Qblox is 1.e+09
-    hardware_demodulation: bool  # demodulation flag
-    integration_length: int
+    # default sampling rate for Qblox is 1.e+09
+    sampling_rate: float
+    # demodulation flag
+    hardware_demodulation: bool
```
jordivallsq, line 30: I agree that SonarCloud messages are annoying but these changes make the PR more confusing to review and I do not agree with these changes for example. @iamtxena Can we make it so SonarCloud stops complaining for comments at the side?

fedonman: @jordivallsq I manage sonlarcloud rules.

fedonman: This rule makes sense. Explanatory comments should appear above the code they refer to. First, this allows for better formatting. For example, longer comments can be split across multiple lines. Second, it leaves inline comments available for tool-specific directives, such as `# ruff: ignore[...]`. Third, in many cases, a docstring is more appropriate than a comment. Here, for example, `sampling_rate` and `hardware_demodulation` should be documented in the docstring rather than explained through inline comments.

PR #1152, opened 2026-06-30
Qhc 1431 rename database folder locations to use microsecond and report missing database.ini parameters

jordivallsq:
- Added a `ValueError` while creating the `DatabaseManager` (for example with `get_db_manager`) checking for `user`, `passwd`, `host`, `port` or `database` inside the database.ini config file, if any of these parameters is missing an error is thrown.
- Fixed the folder shape at `add_measurement` and `add_results` to take into account us intervals of time. This will solve any issue with parallelization while creating more than one folder in less than a second.


jjmartinezQT (approved):
good work

PR #1153, opened 2026-07-01
Bug fix experiment executor creates incorrect qprogram loop arrays

jordivallsq:
Fixed `ExperimentExecutor` not generating the loop shape correctly for data coming from a `QProgram` hardware loop, breaking the execution whenever the data does not fit with .h5 loop shape. This has been fixed by setting the same data size from `ExperimentExecutor` class axis generation as the output shape from the Qblox.


src/qililab/qprogram/experiment_executor.py
```diff
@@ -518,24 +519,51 @@ def _inclusive_range(self, start: int | float, stop: int | float, step: int | fl
...
+                    self._inclusive_range_qprogram(block.start, block.stop, block.step)
+                    if isinstance(block, ForLoop)
+                    else block.values
+                )
+            else:
+                values = (
+                    self._inclusive_range(block.start, block.stop, block.step)
+                    if isinstance(block, ForLoop)
+                    else block.values
+                )
```
fedonman: Why do we have two different interpretations of inclusive range? The `for_loop` should produce the same ranges regardless if it is defined within qprogram or within experiment aka if it is executed in hardware or software.

jordivallsq: The problem is that QBlox output using Hardware loops is not a simple arange, it follows the structure defined in `_inclusive_range_qprogram` therefore if `_inclusive_range` is used, in some cases (depending on the start, stop and step values of the for_loop) the range defined by `_inclusive_range` and the output from the QBlox may mismatch, crashing the h5 writing because the shapes are incorrect.

If you want one single `_inclusive_range` structure, we can use the `_inclusive_range_qprogram` for SW loops as well. I didn't change it because I thought it was the desired structure for SW loops

fedonman: @jordivallsq Either SW loops should behave like HW loops or HW loops should behave like SW loops. But the fact is that both loops should produce the same ranges no matter what. What is the current difference? The original spec was that `for_loop(A, B, C)` produces `[A, B]` both inclusive, e.g. `for_loop(0.0, 1.0, 0.1)` produces `[0.0, 0.1, 0.2, ..., 0.9, 1.0]`. What are the cases that ranges are different between HW and SW? Also, is this difference specific to Qblox? Then probably this is where the bug lies.

fedonman: I experimented a bit with different values and I think the correct version is a mixture of both.

In some cases `_inclusive_range_qprogram` produces better results:

```
In [12]: _inclusive_range_qprogram(0.0, 10.0, 0.33)
Out[12]:
array([ 0.        ,  0.33333333,  0.66666667,  1.        ,  1.33333333,
        1.66666667,  2.        ,  2.33333333,  2.66666667,  3.        ,
        3.33333333,  3.66666667,  4.        ,  4.33333333,  4.66666667,
        5.        ,  5.33333333,  5.66666667,  6.        ,  6.33333333,
        6.66666667,  7.        ,  7.33333333,  7.66666667,  8.        ,
        8.33333333,  8.66666667,  9.        ,  9.33333333,  9.66666667,
       10.        ])

In [13]: _inclusive_range(0.0, 10.0, 0.33)
Out[13]:
array([ 0. ,  0.3,  0.7,  1. ,  1.3,  1.7,  2. ,  2.3,  2.7,  3. ,  3.3,
        3.7,  4. ,  4.3,  4.7,  5. ,  5.3,  5.7,  6. ,  6.3,  6.7,  7. ,
        7.3,  7.7,  8. ,  8.3,  8.7,  9. ,  9.3,  9.7, 10. ])
```

In other cases, `_inclusive_range` produces better results:

```
In [17]: _inclusive_range_qprogram(0, 10, 0.3)
Out[17]:
array([ 0.        ,  0.3030303 ,  0.60606061,  0.90909091,  1.21212121,
        1.51515152,  1.81818182,  2.12121212,  2.42424242,  2.72727273,
        3.03030303,  3.33333333,  3.63636364,  3.93939394,  4.24242424,
        4.54545455,  4.84848485,  5.15151515,  5.45454545,  5.75757576,
        6.06060606,  6.36363636,  6.66666667,  6.96969697,  7.27272727,
        7.57575758,  7.87878788,  8.18181818,  8.48484848,  8.78787879,
        9.09090909,  9.39393939,  9.6969697 , 10.        ])

In [18]: _inclusive_range(0, 10, 0.3)
Out[18]:
array([ 0. ,  0.3,  0.6,  0.9,  1.2,  1.5,  1.8,  2.1,  2.4,  2.7,  3. ,
        3.3,  3.6,  3.9,  4.2,  4.5,  4.8,  5.2,  5.5,  5.8,  6.1,  6.4,
        6.7,  7. ,  7.3,  7.6,  7.9,  8.2,  8.5,  8.8,  9.1,  9.4,  9.7,
       10. ])
```

jordivallsq: Thank for the reply @fedonman, There is no set way of measuring software loops, so it can be any structure that we want. I just left it as it was because I thought it was the desired outcome. But we can use `_inclusive_range_qprogram`.

Sadly, there is no possibility of using `_inclusive_range` and the Qblox output fitting these loops. Even in the loop that you said looked better, there is a jump that is a 0.4 (instead of 0.3) between `4.8,  5.2`; this is constantly happening with `_inclusive_range`. This is impossible for the Qblox to process, as all jumps must be evenly spaced, so the only solution is to set all loops to `_inclusive_range_qprogram` (which should be fine even for SW loops).

fedonman: > Sadly, there is no possibility of using _inclusive_range and the Qblox output fitting these loops. Even in the loop that you said looked better, there is a jump that is a 0.2 (instead of 0.3) between 5.8,  6.1; this is constantly happening with _inclusive_range. This is impossible for the Qblox to process, as all jumps must be evenly spaced

You are right, this isn't possible.

> so the only solution is to set all loops to _inclusive_range_qprogram (which should be fine even for SW loops).

I wouldn't prefer this currently.

Let's go ahead with your solution. Thanks a lot for the fix and explaining it to me. 💯

fedonman (approved):
Thanks a lot Jordi! 💯 💯

jordivallsq:
Implementation tested in HW:
For the example that you have given for a range between 0-10 with a step of 0.3 (In my example is 0-1 with a step of 0.03 because of Qblox units). The hardware loop is:
```
offset = qp.variable(label="Offset", domain=ql.Domain.Voltage)
with qp.for_loop(variable=offset, start=0, stop=1, step=0.03):
```
and it outputs:
```
DimensionInfo(labels=['Offset'], values=[array([0.        , 0.03030303, 0.06060606, 0.09090909, 0.12121212,
         0.15151515, 0.18181818, 0.21212121, 0.24242424, 0.27272727,
         0.3030303 , 0.33333333, 0.36363636, 0.39393939, 0.42424242,
         0.45454545, 0.48484848, 0.51515152, 0.54545455, 0.57575758,
         0.60606061, 0.63636364, 0.66666667, 0.6969697 , 0.72727273,
         0.75757576, 0.78787879, 0.81818182, 0.84848485, 0.87878788,
         0.90909091, 0.93939394, 0.96969697, 1.        ])])
```

and the software loop is:
```
bins = experiment.variable(label="Bins", domain=ql.Domain.Voltage)
with experiment.for_loop(variable=bins, start=0, stop=1, step=0.03):
```
and it outputs:
```
DimensionInfo(labels=['Bins'], values=[array([0.  , 0.03, 0.06, 0.09, 0.12, 0.15, 0.18, 0.21, 0.24, 0.27, 0.3 ,
         0.33, 0.36, 0.39, 0.42, 0.45, 0.48, 0.52, 0.55, 0.58, 0.61, 0.64,
         0.67, 0.7 , 0.73, 0.76, 0.79, 0.82, 0.85, 0.88, 0.91, 0.94, 0.97,
         1.  ])]),
```
Just as we observed using the functions independently.

Regarding the bug fix: despite using a step size that does not fit between the stop and step, the execution is a success.
image

PR #1154, opened 2026-07-06
Qhc 1433 qdac multiple users sharing internal trigger network

jordivallsq:
Fixed the QDAC-II trigger network breaking when multiple users are connected to the same instrument. Each qililab instance allocated internal trigger numbers from its own driver-side pool, with no way of knowing which numbers other Python processes were already using. When two instances picked the same number, their triggers fired through each other's networks, interfering with both experiments.

  Internal triggers in use are now read directly from the instrument's marker registers before every allocation, so a new trigger always takes the lowest number that is actually free on the hardware. Related behavior changes:
  - `QDevilQDac2.start()` now starts only the DC lists and AWG waveforms uploaded by this instance, instead of `start_all()`, which also fired generators armed by other users.
  - `QDevilQDac2.clear_trigger()` now frees the triggers created by this instance from the instrument (previously `free_all_triggers()` released the triggers set by `QCodes`, not affecting the instrument).
  - `Platform.execute_qprogram` now clears each QDAC-II's trigger network and generator caches after every execution, so consecutive executions always start from a clean trigger state.
  - Allocating a trigger when all 14 internal triggers are busy raises a `ValueError` including other users triggers.

### DIFF
--- a/src/qililab/instruments/qdevil/qdevil_qdac2.py
+++ b/src/qililab/instruments/qdevil/qdevil_qdac2.py
@@ -579,9 +579,7 @@ class QDevilQDac2(VoltageSource):
     @check_device_initialized
     def reset(self):
         """Reset instrument. This will affect all channels."""
-        if self._triggers:
-            for trigger_name in list(self._triggers.keys()):
-                self.clear_trigger(trigger_name)
+        self.clear_trigger()
         self.clear_cache()
         self.device.reset()
 
--- a/src/qililab/instruments/qdevil/qdevil_qdac2.py
+++ b/src/qililab/instruments/qdevil/qdevil_qdac2.py
@@ -15,9 +15,11 @@
 """QDevil QDAC-II Instrument"""
 
 from dataclasses import dataclass
+from itertools import product
 from typing import TYPE_CHECKING
 
 import numpy as np
+from qcodes_contrib_drivers.drivers.QDevil.QDAC2 import QDac2Trigger_Context
 
 from qililab.instruments import InstrumentFactory, ParameterNotFound, check_device_initialized, log_set_parameter
 from qililab.instruments.voltage_source import VoltageSource
@@ -26,7 +28,7 @@ from qililab.typings import QDevilQDac2 as QDevilQDac2Driver
 from qililab.waveforms import Waveform
 
 if TYPE_CHECKING:
-    from qcodes_contrib_drivers.drivers.QDevil.QDAC2 import List_Context, QDac2Trigger_Context
+    from qcodes_contrib_drivers.drivers.QDevil.QDAC2 import List_Context
 
 
 @InstrumentFactory.register
@@ -42,6 +44,11 @@ class QDevilQDac2(VoltageSource):
     _MIN_RAMPING_RATE: float = 0.01
     _MAX_RAMPING_RATE: float = 2e7
     name = InstrumentName.QDEVIL_QDAC2
+
+    _N_DACS: int = 24
+    _N_INT_TRIGGERS: int = 14
+    _GENERATOR_LIST: tuple[str, str, str, str, str] = ("DC", "SINe", "SQUare", "TRIangle", "AWG")
+    _MARKER_LOCATION: tuple[str, str, str, str, str, str] = ("STARt", "END", "PSTart", "PEND", "SSTart", "SEND")
 
     @dataclass
     class QDevilQDac2Settings(VoltageSource.VoltageSourceSettings):
@@ -61,6 +68,8 @@ class QDevilQDac2(VoltageSource):
         self._cache_awg: dict[int | str, bool] = {}
         self._cache_dc: dict[int | str, List_Context] = {}
         self._triggers: dict[str, QDac2Trigger_Context] = {}
+
+        self._internal_triggers: dict = {}
 
     @property
     def low_pass_filter(self):
@@ -187,7 +196,7 @@ class QDevilQDac2(VoltageSource):
         self._validate_channel(channel_id=channel_id)
 
         envelope = waveform.envelope()
-        values = list(envelope)  # TODO: does np array work?
+        values = list(envelope)
         if channel_id in self._cache_awg:
             raise ValueError(
                 f"Device {self.name} already has a waveform allocated to channel {channel_id}. Clear the cache before allocating a new waveform"
@@ -289,7 +298,12 @@ class QDevilQDac2(VoltageSource):
         if str(trigger) in self._triggers.keys():
             self.clear_trigger(trigger)
 
-        self._triggers[str(trigger)] = self._cache_dc[f"{self.device.name}_{channel_id}"].start_marker()
+        self._triggers[str(trigger)] = self.allocate_internal_trigger(
+            self._cache_dc[f"{self.device.name}_{channel_id}"]
+        )
+
+        channel = self.device.channel(channel_id)
+        channel.write_channel(f"sour{'{0}'}:dc:mark:star {self._triggers[str(trigger)].value}")
 
         self.device.connect_external_trigger(port=out_port, trigger=self._triggers[str(trigger)], width_s=width_s)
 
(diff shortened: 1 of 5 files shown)


src/qililab/instruments/qdevil/qdevil_qdac2.py
```diff
@@ -455,23 +424,87 @@ def play_awg(self, channel_id: ChannelID | None = None, clear_after: bool = True
...
+        if trigger in self._triggers:
+            self.clear_trigger(trigger)
+
+        self._triggers[trigger] = self.allocate_internal_trigger(self._cache_dc[f"{self.device.name}_{channel_id}"])
+
+        channel = self.device.channel(channel_id)
+        channel.write_channel(f"SOUR{'{0}'}:DC:MARK:{marker_location} {self._triggers[trigger].value}")
+        self._marker_registers[trigger] = (channel_id, "DC", marker_location)
+
+    def _check_internal_triggers(self):
```
flavie-lebars: missing return type

src/qililab/instruments/qdevil/qdevil_qdac2.py
```diff
@@ -455,23 +424,87 @@ def play_awg(self, channel_id: ChannelID | None = None, clear_after: bool = True
 
     def start(self):
         """All generators, that have not been explicitly set to trigger on an internal or external trigger, will be started."""
-        self.device.start_all()
+        for channel_id, awg in self._cache_awg.items():
```
flavie-lebars: for the start it seems that the baheviour has changed beyond what is expected.
Comment from claude"start() (qdevil_qdac2.py:425-434): old version (start_all() → *TRG) only started channels still in default "bus" trigger mode, correctly skipping any channel armed via set_in_internal_trigger/set_in_external_trigger (those wait for their own trigger). New version calls .start() on every cached dc_list/awg unconditionally, which force-fires immediately regardless of trigger config — so an armed channel now starts right away instead of waiting for its real trigger. Docstring above start() no longer matches behavior. Is set_in_*_trigger + start() used together anywhere? If so this needs a hardware check."

src/qililab/instruments/qdevil/qdevil_qdac2.py
```diff
@@ -455,23 +424,87 @@ def play_awg(self, channel_id: ChannelID | None = None, clear_after: bool = True
...
+            query = ";:".join(
+                f"SOUR{channel_id}:{generator}:MARK:{marker_location}?" for generator, marker_location in registers
+            )
+            response = self.device.ask(query)
+            for (generator, marker_location), value in zip(registers, response.split(";")):
+                internal_trigger = int(value)
+                if internal_trigger != 0:
+                    self._internal_triggers[internal_trigger] = (channel_id, generator, marker_location)
+
+    def allocate_internal_trigger(self, qdac_generator: "QDevilQDac2Driver") -> QDac2Trigger_Context:
```
flavie-lebars: qdac_generator is a List_Context

src/qililab/instruments/qdevil/qdevil_qdac2.py
```diff
@@ -455,23 +424,87 @@ def play_awg(self, channel_id: ChannelID | None = None, clear_after: bool = True
...
+            ValueError: no free triggers
+
+        Returns:
+            QDac2Trigger_Context: Context manager
+        """
+        self._check_internal_triggers()
+        free_int_trigger = set(range(1, self._N_INT_TRIGGERS + 1)) - self._internal_triggers.keys()
+        if not free_int_trigger:
+            raise ValueError("No free internal triggers")
+        return QDac2Trigger_Context(qdac_generator, min(free_int_trigger))
```
flavie-lebars: should not be the qdac_generator here but self.device

flavie-lebars (changes requested):

src/qililab/instruments/qdevil/qdevil_qdac2.py
jordivallsq: set return to None

src/qililab/instruments/qdevil/qdevil_qdac2.py
jordivallsq: Very nice catch! I created a cache for dc lists `_armed_on_trigger` that saves all dc lists set to be triggered by a trigger and for those I omit the start

src/qililab/instruments/qdevil/qdevil_qdac2.py
jordivallsq: since I modified the QDac2Trigger_Context input this parameter is no longer needed and I removed it

src/qililab/instruments/qdevil/qdevil_qdac2.py
jordivallsq: I was looking into the wrong allocate_trigger for reference but now I changed the input to self.device (the right one)

jordivallsq:
I also modified `_MARKER_LOCATION` (the possible location of the trigger) to exclude "SSTart", "SEND" (send a trigger at the **ST**art or **END** of every **S**tep) because it only applies to DC lists, thus I created
```
_DC_MARKER_LOCATION: tuple[str, str] = ("SSTart", "SEND")
```
and modified the registers to contain less combinations

jordivallsq:
This PR has been tested successfully on hardware.
The communication between users with both setting triggers using qdac drivers has been tested between [User A](https://github.com/qilimanjaro-tech/QHC_hardware_tests/blob/main/tests/qililab/qdac_ii/test_PR_1154_userA.ipynb) and [User B](https://github.com/qilimanjaro-tech/QHC_hardware_tests/blob/main/tests/qililab/qdac_ii/test_PR_1154_userB.ipynb). And the interaction with `platform.execute_qprogram(...)` has been tested in [this notebook](https://github.com/qilimanjaro-tech/QHC_hardware_tests/blob/main/tests/qililab/qdac_ii/test_PR_1154_qdac_qblox.ipynb).

src/qililab/instruments/qdevil/qdevil_qdac2.py
```diff
@@ -199,13 +213,13 @@ def upload_awg_waveform(self, waveform: Waveform, channel_id: ChannelID):
...
         trace.waveform(values)
-        self._cache_awg[channel_id] = True
+        self._cache_awg[channel_id] = self.get_dac(channel_id).arbitrary_wave(trace.name)
 
     def upload_voltage_list(
         self,
         waveform: Waveform,
         channel_id: ChannelID,
-        dwell_us: int = 1,
+        dwell_us: int = 2,
```
flavie-lebars, line 222: why has the default changed?

src/qililab/platform/platform.py
```diff
@@ -1257,6 +1256,9 @@ def compile_qprogram(
                     if isinstance(instrument, QDevilQDac2)
                 }
             )
+            # Start from a clean instrument state before the compiler re-populates it
+            for qdac_instrument in self.qdac_instruments:
```
flavie-lebars, line 1260: why do we need to clear the cache at both the compilation and execution stage?

src/qililab/instruments/qdevil/qdevil_qdac2.py
```diff
@@ -326,7 +344,7 @@ def set_end_marker_external_trigger(
             channel_id (ChannelID): Channel id of the qdac
             out_port (int): Trigger output port.
             trigger (str): Name of the trigger.
-            width_s (float, optional): duration in seconds of the trigger pulse. Defaults to 1e-6.
+            width_s (float, optional): duration in seconds of the trigger pulse. Defaults to 2e-6 s.
```
flavie-lebars: the value in the docstring changed from 1e-6 to 2e-6 but the default value remained 1e-6

src/qililab/platform/platform.py
jordivallsq: The cache cleaning before compilation is to make sure that previously interrupted executions get removed (for example if the measurement crashes due to an incorrect parameter and the triggers have already been set, or even if someone has been working with QDAC driver commands without using Qprogram)

At the end is to make sure to leave the triggers and local registered lists and input triggers clean for the next execution.

Whenever a clear_cache is executed in an empty cache, nothing is run in the qdac and the parameters are set as empty in the same way as they were before

src/qililab/instruments/qdevil/qdevil_qdac2.py
jordivallsq: Because it was the wrong docstring, apologies

src/qililab/instruments/qdevil/qdevil_qdac2.py
jordivallsq: Because the right minimum for voltage lists dwell time is 2 us: https://qm.quantum-machines.co/hubfs/QDAC%20II%20-%20User%20manual%20v2.2%20(2024-01-17).pdf

Image

I have corrected the docstrings and added a changelog entry on bug fixes.

flavie-lebars (approved):
Very nice feature!

PR #1158, opened 2026-07-10
[QHC-1482] Fix Windows-only test failures (data_management, slurm)

flavie-lebars:
- test_data_management.py: compare Path objects instead of raw strings, since save_platform() returns OS-native separators.
- test_slurm.py: skip TestSubmitJob on Windows, since submitit's local executor relies on POSIX-only signal.SIGKILL.

### DIFF
--- a/docs/releases/changelog-dev.md
+++ b/docs/releases/changelog-dev.md
@@ -4,6 +4,9 @@
 
 ### Improvements
 
+- Fixed `test_data_management.py` and `test_slurm.py` failing on local Windows dev environments: `save_platform()`'s return path is now compared as a `Path` instead of a raw string (Windows uses backslash separators), and `TestSubmitJob` (which relies on `submitit`'s POSIX-only local executor) is now skipped on Windows.
+  [#1158](https://github.com/qilimanjaro-tech/qililab/pull/1158)
+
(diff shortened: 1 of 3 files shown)


docs/releases/changelog-dev.md
```diff
@@ -4,6 +4,9 @@
 
 ### Improvements
 
+- Fixed `test_data_management.py` and `test_slurm.py` failing on local Windows dev environments: `save_platform()`'s return path is now compared as a `Path` instead of a raw string (Windows uses backslash separators), and `TestSubmitJob` (which relies on `submitit`'s POSIX-only local executor) is now skipped on Windows.
```
DZPM, line 7: Question: should the test catch the exact exception, instead of just skipping it? It would be more resilient

flavie-lebars: I don't think so, we skip by platform because Windows lacks signal.SIGKILL, which submitit's local executor needs. Catching the exception wouldn't help anyway, since the real damage is a leaked file handle that breaks other tests afterward, not just this one.

jordivallsq (approved):
Great catch! 
Tested in Linux (server) and Windows (my local) and tests now pass for both.

PR #1161, opened 2026-07-13
[QHC-1467] Natural bus sorting for CrosstalkMatrix + fix array/label ordering inconsistency

elygoner:
### Summary
Introduces a natural bus-name sorter (`q2` before `q10`) and, in doing so, makes `CrosstalkMatrix`'s array↔label ordering consistent end-to-end. The consistency work fixes a latent bug where any matrix whose buses weren't stored in natural order produced a **mislabeled inverse and wrong bias values**.

### Motivation
The original goal was cosmetic: sort buses so multi-digit names read naturally in `to_array`/`__str__` instead of lexicographically (`flux q10` before `flux q2`). But `to_array` feeds `inverse` → `flux_to_bias` and the calibration matrix products, and those paths didn't share a single ordering, which surfaced a correctness bug.

### The bug
`to_array()` orders rows/columns with the sorter, but `inverse()`/`from_array()` labeled the array by **raw insertion order** (`list(self.matrix.keys())`), and `calibration` built offset vectors / relabeled products the same way. When a matrix's key order differs from the sorted order — e.g. any system with ≥10 buses stored alphabetically (`flux q0, flux q1, flux q10, flux q2`), which YAML runcards preserve verbatim — the array and its labels desynced:

- `A @ inverse().to_array()` was **not** the identity;
- `flux_to_bias` returned wrong bias (~30% off in a 4-bus example), propagating to `FluxVector`, the hardware bias vector, and the annealing waveform path;
- `Calibration.add_intra_crosstalk` / `add_inter_crosstalk` **persisted** a corrupted matrix, and `add_intra_crosstalk` additionally swapped offsets (matrix in sorted order × offset vector in insertion order).

Single-digit systems were unaffected (`sorted == sort_buses`), which is why it went unnoticed.

### Changes
- **`utils/bus_sorter.py`**: new `sort_buses` (stable order: fewer indices first → numeric index → bus type readout/drive/flux → loop type x/z → alphabetical tiebreak) and `argsort_buses`, which also returns the sort permutation so an array and its labels can be reordered together (`array[np.ix_(order, order)]`).
- **`crosstalk_matrix.py`**: added a private `_sorted_buses()` as the single source of truth; `to_array`/`__str__` use it, and **`inverse()` now labels the inverse with the same order `to_array` produced**.
- **`calibration.py`**: `add_intra_crosstalk`/`add_inter_crosstalk` do the matrix relabeling, offset vectors, and `from_array` calls in one consistent `sort_buses` order.
- **Tests**: parametrized regression tests (natural / alphabetical / scrambled / mixed insertion orders) asserting `A @ inv == I` and `flux_to_bias` matches an independent inversion; per-criterion sorter tests + an `argsort_buses` test.
- **Changelog**: entries under New features, Improvements, and Bug fixes.


docs/releases/changelog-dev.md
```diff
@@ -7,8 +7,14 @@
 - Added bus_mapping to measurement database table `Measurement`. Bus mapping is necessary for live plot drawing of the qprogram and it has been information missing in the database. StreamArray already has the bus_mapping as an input, this input is the dictionary that will be saved in the database.
   [#1136](https://github.com/qilimanjaro-tech/qililab/pull/1136)
 
+- Added `sort_buses` and `argsort_buses` to `qililab.utils`, utilities that order bus identifiers into a stable, easy to read order: fewer-index buses before couplers, indices compared numerically (so `q2` precedes `q10`), then by bus type (readout, drive, flux) and loop type. `argsort_buses` also returns the sort permutation, so a matrix and its bus labels can be reordered together.
+  [#XXXX](https://github.com/qilimanjaro-tech/qililab/pull/XXXX)
```
jordivallsq: You forgot to change the number of the PR

docs/releases/changelog-dev.md
```diff
@@ -7,8 +7,14 @@
...
+
 ### Improvements
 
+- `CrosstalkMatrix.to_array` and its `__str__` representation now order buses with `sort_buses`, so multi-digit bus names are shown in natural order (`flux q2` before `flux q10`) instead of lexicographically.
+  [#XXXX](https://github.com/qilimanjaro-tech/qililab/pull/XXXX)
```
jordivallsq: Here too

docs/releases/changelog-dev.md
```diff
@@ -7,8 +7,14 @@
 - Added bus_mapping to measurement database table `Measurement`. Bus mapping is necessary for live plot drawing of the qprogram and it has been information missing in the database. StreamArray already has the bus_mapping as an input, this input is the dictionary that will be saved in the database.
   [#1136](https://github.com/qilimanjaro-tech/qililab/pull/1136)
 
+- Added `sort_buses` and `argsort_buses` to `qililab.utils`, utilities that order bus identifiers into a stable, easy to read order: fewer-index buses before couplers, indices compared numerically (so `q2` precedes `q10`), then by bus type (readout, drive, flux) and loop type. `argsort_buses` also returns the sort permutation, so a matrix and its bus labels can be reordered together.
```
jordivallsq: `fewer-index buses before couplers, indices compared numerically (so `q2` precedes `q10`), then by bus type (readout, drive, flux) and loop type`

This sentence is not very clear, what are the priorities, first qubit number, then line type and then if it is a qbit or coupler? Shouldn't this order be reversed?

jordivallsq: The order described in the docstring for sort_buses is clearer in my opinion

src/qililab/utils/bus_sorter.py
```diff
@@ -0,0 +1,70 @@
...
+    (r"flux|(?<![a-z])(?:f)(?![a-z])", 2),
+)
+
+
+def _bus_sort_key(bus: str):
+    """Sort key for a single bus name. See :func:`sort_buses` for the ordering criteria."""
+    ids = sorted(int(i) for i in re.findall(r"\d+", bus))
+    bus_type = next((order for pattern, order in _BUS_TYPE_ORDER if re.search(pattern, bus, flags=re.IGNORECASE)), 3)
+    loop_type = _LOOP_TYPE_ORDER.get(
+        m.group().lower() if (m := re.search(r"(?<![a-z])[xz](?![a-z])", bus, flags=re.IGNORECASE)) else "", 3
```
jordivallsq: please replace this m with a recognizable name. it might be anoying for generic variables such as these but makes the code way more understandable in the future

tests/qprogram/test_crosstalk_matrix.py
```diff
@@ -264,6 +288,21 @@ def test_flux_to_bias_with_array(self, crosstalk_matrix):
             for bus in flux_dict:
                 assert float(bias[bus][i]) == pytest.approx(scalar_bias[bus], rel=1e-6)
 
+    # bus-ordering regression: array and labels must stay consistent for any key insertion order
```
jordivallsq: remove this comment or put it inside the the test

src/qililab/utils/bus_sorter.py
jordivallsq: I would say that the exceptions are the `i`'s used in this file as they are a standard for loop iterations

src/qililab/utils/bus_sorter.py
```diff
@@ -0,0 +1,70 @@
...
+    Args:
+        bus_sequence (Iterable[str]): Bus identifiers to sort.
+
+    Returns:
+        tuple[list[str], list[int]]: (sorted_buses, order) where
+            sorted_buses[k] == list(bus_sequence)[order[k]].
+    """
+    items = list(bus_sequence)
+    order = sorted(range(len(items)), key=lambda i: _bus_sort_key(items[i]))
+    return [items[i] for i in order], order
```
jordivallsq: What is this `order` and why sort_buses grabs only order 0. What are the rest of orders not used in sort_buses? are we generating several times the data necessary every time we sort?

Please contextualize the description and returns of `argsort_buses` and add a comment on why [0] in line 70

jordivallsq (commented):
Good work!! The implementation is very solid, I left some comments focusing on descriptions and docstrings.

Additionally there is a `sorted` in [FluxVector](https://github.com/qilimanjaro-tech/qililab/blob/4af3978d65d479cf477fabd1768ddb946d3d8a75/src/qililab/qprogram/flux_vector.py#L473) does it need the same treatment?

docs/releases/changelog-dev.md
elygoner: no I didn't it is a placeholder because I hadn't made the PR yet, I have to add it.

docs/releases/changelog-dev.md
elygoner: Ok then I can paste it here

src/qililab/utils/bus_sorter.py
elygoner: An argsort returns the sorted list and then gives you the order of the original indices. Like if you put [6,2,3] (0,1,2) you are going to get back ([2,3,6],[1,2,0]). Order is the second one. 
To return this, what we do is, sort the indices using as sorting order the sorting key of the value they represent (`lambda i: _bus_sort_key(items[i]))`). This gives you the indices of the sorted bus list, then you do `[items[i] for i in order]` to get the actual sorted list.

Sort_buses only uses 0 because it only cares about the list of sorted buses and not the ordered index list.

elygoner: I can add this as a comment but I think it shouldn'0t be in the docstring because you shouldn't inform the user of the low level.

jordivallsq: I agree that the heavy explanation should be in the comments not in the description but the return should give context on what is it that your sorter is giving to the user, so a small point regarding the two lists generated is necessary instead of what it is right now

tests/qprogram/test_crosstalk_matrix.py
elygoner: This comment is a separator that tell you what the tests below are for. I added extra `#` before and after to elucidate its function.

jordivallsq (approved):
Well done!

PR #1162, opened 2026-07-14
Qhc 1478 database changes for calibration gui

jordivallsq:
Added `fitting_path` and `fitting_parameters` columns to `Measurements` database. These optional fields add the location folder of the resulting fit plots and the outcome parameters of those fittings respectively. The path is introduced as a string and the parameters as a dictionary with the parameter name as a string. Those can be added at three different levels:
  - Using the `Measurement` database table with `meas.add_fitting(db_manager.session, path, parameters)`. The only inconvenience is that the user needs to add the database manager session.
  - Using `DatabaseManager` with `db_manager.add_fitting(id, path, parameters)`. Using an ID like `load_by_id`. This method is ideal to add the fitting results in a later session (for completed experiments or if the post processing of the data has been done in a separate Python instance).
  - Using `StreamArray` with `stream_array.add_fitting(path, parameters)` after a measurement that used `StreamArray`. This is the simplest way of adding the fit results as it requires no ID nor session but it requires the instance of `StreamArray` to still be loaded. The ideal scenario is for fits done in sequence after measurements (such as a `qilitools` implementation).


jordivallsq:
Adding @jjmartinezQT and @OscarQili as reviewers just to check and approve the structure of the new fitting database parameters.

src/qililab/result/database/database_manager.py
```diff
@@ -252,6 +252,23 @@ def load_by_id(self, id: int | list[int]) -> list[Measurement] | Measurement | N
                         meas.result_path = new_path
             return measurement_by_id_list if len(measurement_by_id_list) > 1 else measurement_by_id_list[0]
 
+    def add_fitting(self, id: int, path: str, parameters: dict[str, Any] | None = None) -> Measurement:
+        """Attach fitting results to an existing measurement, loaded by its measurement_id.
+
+        This is a thin wrapper over ``Measurement.add_fitting`` that loads the measurement the same
+        way ``load_by_id`` does and injects the database session, so the user does not need to handle
+        it.
```
elygoner: I don't know about the mechanism explanation.

src/qililab/result/database/database_measurements.py
```diff
@@ -253,6 +256,19 @@ def read_numpy(self):
         axis_labels = {dim: da.coords[dim].values for dim in da.dims[1:]}
         return arr, axis_labels
 
+    def add_fitting(self, session: sessionmaker[Session], path: str, parameters: dict[str, Any] | None = None):
```
elygoner: add `DatabaseManager` as a possible input since `DatabaseManager.session` isn't really used by the users in most cases.

elygoner (approved):
Looks pretty good, I left two comments.
I'll also add here AI code review that I think are worth at least considering:
## Code Review — PR #1162

Scope (`origin/main...HEAD`): `database_manager.py`, `database_measurements.py`, `stream_results.py` + tests. (QDAC changes were already in `origin/main`, not part of this PR.)

---

### 1. `DatabaseManager.add_fitting` silently rewrites `result_path` in the DB — HIGH
`database_manager.py:267`

`load_by_id` mutates `meas.result_path` to the **share** path when the local file is absent (lines 250-252), then returns that detached instance. `Measurement.add_fitting` does `session.merge(self)` + `commit()`, so `merge` treats the in-memory share-path value as canonical and the commit **UPDATEs `result_path` in the database**.

This fires in exactly the documented use case — "add fitting results in a later session for completed experiments" — where the local file usually won't exist.

### 2. Missing measurement raises `IndexError`, not `ValueError` — HIGH
`database_manager.py:268`

The `if measurement is None:` guard is dead code. `load_by_id` never returns `None` for a missing id — line 253 does `... else measurement_by_id_list[0]`, and an empty result is `[]`, so `[][0]` → `IndexError: list index out of range`.

`db_manager.add_fitting(9999, "/p")` crashes with `IndexError` instead of the promised `ValueError("Measurement entry '9999' does not exist.")`. The test passes only because it mocks `load_by_id` to return `None`, masking the real contract.

### 3. `if parameters:` silently drops an empty dict — LOW
`database_measurements.py:263`

Passing `parameters={}` skips the assignment, so `fitting_parameters` stays `None`, and you can never clear previously-stored parameters via `add_fitting(..., parameters={})`. Prefer `if parameters is not None:`.

### 4. `Measurement.add_fitting` duplicates `end_experiment` boilerplate — LOW
`database_measurements.py:259`

The `with session() ... merge → set fields → try commit/return / except rollback/raise` block is copy-pasted verbatim from `end_experiment` (lines 196-209). A shared private helper would remove the duplication and keep the two write paths consistent.

src/qililab/result/database/database_manager.py
jordivallsq: fixed docstrings

src/qililab/result/database/database_measurements.py
jordivallsq: Good point! changed to DB Manager
also added docstrings

jordivallsq (commented):
@elygoner Thank you for your comments, I implemented all the requested changes. Regarding the commented points:
1. The result path is modified (from local to shared/NAS in case the local is not found) as it is intended whenever someone uses load_by_id. Here we are doing the same as loading by id but with extra features.
2. Modified the error into IndexError as it makes more sense.
3. This is the intended behavior, as these parameters is not something one should erase (maybe modify in case of a mistake, but this is completely possible).
4. I didn't want to create a new internal function just for this (only two functions with similar behaviour while running session but updating completely different fields in the database).

OscarQili:
The main functionality looks all good to me

jordivallsq:
The PR has been tested with the database. It took a while because the database policies were missing permits to update the database, while the experiment had finished (issues with the StreamArray), so I modified the policies with this:
```
ALTER POLICY update_measurements_policy ON measurements
    USING (NOT experiment_completed OR fitting_path IS NULL);
```
Now adding the fitting information updates the database (loading the entry on another instance allows for free modification, this constraint only affects any changes during the StreamArray execution)
And now all the tests pass:
https://github.com/qilimanjaro-tech/QHC_hardware_tests/blob/main/tests/qililab/test_PR_1162.ipynb

PR #1164, opened 2026-07-15
Fix QDAC trigger race: defer clear_cache() until after acquisition

jjmartinezQT:
## Summary

Fixes a regression from #1154 where triggered `Qblox` executions using a `QDAC-II` could hang forever, stuck waiting for a trigger that never arrives.

- `qdac.start()` issues a non-blocking SCPI command (`sour{n}:dc:init`); the QDAC's actual trigger pulse to Qblox fires later, on the instrument's own internal timing (dwell/delay).
- The post-execution `qdac.clear_cache()` added in #1154 ran immediately after `start()` and calls `clear_trigger()`, which unconditionally zeroes the exact SCPI marker register (`SOUR{ch}:{gen}:MARK:{loc}`) that routes that pulse to the external port wired into Qblox.
- If that register got zeroed before the instrument's internal clock reached the marker event, the pulse never fired. Qblox stayed armed on `wait_trigger` indefinitely — exactly the symptom reported by an experimentalist on Mimas after pulling the latest `qililab`/`qilitools` main.
- `trigger_position == "front"` is the worst case: `qdac.start()` and the trailing `qdac.clear_cache()` run back-to-back with nothing in between. `trigger_position == "back"` has a loop of `bus.run()` calls in between, giving more elapsed time, so it's less exposed but not immune.

This is invisible to unit tests, since they mock `QDevilQDac2.start()`/`clear_cache()` with no real timing, and invisible to the existing `QHC_hardware_tests` QDAC notebooks, since none of them put a `wait_trigger` on the Qblox side (they validate #1154's actual goal — trigger-register allocation not colliding across users — extremely thoroughly, but never exercise a Qblox sequence that blocks waiting for the QDAC's physical pulse).

## Fix

Move the trailing `qdac.clear_cache()` from immediately after `qdac.start()` to right after the acquisition block, instead of removing it. `bus.acquire_qprogram_results(...)` can only return once Qblox's `wait_trigger` has already unblocked, which can only happen after the real trigger pulse arrived — so by that point, clearing the marker register is provably safe, not just "probably enough time has passed."

This was deliberately **not** just deleted: per #1154's own description, the whole point of that PR was letting multiple users share the QDAC's 14-slot internal trigger pool, freeing this execution's registers as promptly as possible rather than waiting for this process's own next compile. Deferring within the same call preserves that goal while removing the race.

One caveat for reviewers: if a qprogram drives a QDAC trigger but the resulting bus set has no ADC bus at all (nothing to acquire), the code never reaches a blocking wait and the same race could theoretically resurface. Didn't find an existing "wait for DC-list completion" primitive in `qdevil_qdac2.py` to guard that case explicitly — flagging for discussion rather than blocking this fix on it, since the reported failure mode does involve acquisition.

## Test plan

- [x] `py_compile` clean
- [x] **Could not run `tests/platform/test_platform.py -k qdac` locally** — this machine's Xcode Command Line Tools C++ headers are broken (missing `cstdint` under `/Library/Developer/CommandLineTools/usr/include/c++/v1/`, unrelated to this change), which breaks building the `qilisdk` git dependency needed to import `qililab` at all. Needs CI or a working local toolchain to confirm the existing QDAC platform tests still pass.
- [x] Recommend validating against real hardware (Mimas) before merging, given this touches the exact code path that broke — ideally with a new `QHC_hardware_tests` notebook that adds an actual `wait_trigger` on the Qblox side (none of the existing three QDAC notebooks exercise that), to both confirm the fix and close the coverage gap that let this regression through.
- [x] Given this touches trigger-sharing logic Jordi designed in #1154, would like his review before merging.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


flavie-lebars (changes requested):
- Very minor note, but make sure to tell the AI the format we have to name branch/PR `[QHC-XXX]` issue (we should add this to the CLAUDE.md)
- One note on the implementation, it means that the cache is not cleared if it fails and goes into `TimeOutError`.

jjmartinezQT:
Good catch, thanks — fixed in 02c231ac.

`clear_cache()` is now also called in the `except TimeoutError` branch, before deciding whether to retry or give up. It's safe there for the same reason it's safe after acquisition on the success path: reaching that branch means Qblox's own wait/acquisition polling already gave up, so there's no pulse still in flight to race against.

While adding a regression test for this I found the two existing timeout tests had a latent bug: they set `.qdac` on the `QdacCompilationOutput` mock, but the real class only has `.qdacs` (a list). That was invisible before because the old code never touched `.qdacs` before the mocked `run()` raised `TimeoutError` — my new `clear_cache()` call is the first thing on that path that does. Fixed both mocks and added `clear_cache.call_count` assertions so this is now actually covered.

Full local suite (1597 tests) passes after this change.

jjmartinezQT:
Fixed the SonarQube duplication gate failure too (971b9c40).

The two QDAC timeout-error tests already shared ~25 near-identical lines of mock setup before this PR touched them — my clear_cache() assertion just tipped both copies over the "new code" duplication threshold together. Extracted the shared setup into a \`_build_qdac_timeout_mocks()\` helper; each test now only configures what actually differs between them (\`check_recurrent_timeout\`). Full suite still passes (1597 tests).

flavie-lebars (commented):
- I find the comments very wordy😅, they are both 6 lines each, but up to you.
- If a program doesn't have any acquisitions, it behaves the same as before the bug fix, it goes straight from start to clearing the cache. This is related in a way to [this linear task](https://linear.app/qilimanjaro/issue/QHC-1488/bug-allow-for-long-time-loops) where in general we are not considering use cases without acquisitions. But I guess when using the qdac we always want to acquire with the qblox so as of now it is not a problem.

jjmartinezQT:
> * I find the comments very wordy😅, they are both 6 lines each, but up to you.
> * If a program doesn't have any acquisitions, it behaves the same as before the bug fix, it goes straight from start to clearing the cache. This is related in a way to [this linear task](https://linear.app/qilimanjaro/issue/QHC-1488/bug-allow-for-long-time-loops) where in general we are not considering use cases without acquisitions. But I guess when using the qdac we always want to acquire with the qblox so as of now it is not a problem.

Yeah, I have to revisit my claude settings because this were automatic comments that were posted without supervision :D

src/qililab/platform/platform.py
```diff
@@ -1462,6 +1469,16 @@ def _execute_qblox_compilation_output(self, output: QProgramCompilationOutput, d
...
+
+                # A TimeoutError here means Qblox's own wait/acquisition polling already gave up,
+                # so there is no pulse still in flight to race against (unlike the success path,
+                # clearing here can't zero the marker register before Qblox consumes it). Clear
+                # regardless of whether we're about to retry or give up, so a retry starts from a
+                # clean trigger state and a final failure doesn't leak an allocated trigger until
+                # the next execute_qprogram's pre-compile clear_cache() happens to run.
+                for qdac in output.qdac.qdacs:
+                    qdac.clear_cache()
+
```
jordivallsq: These lines have to go in lines 1945 (same indentation it still needs to be inside the condition `if output.qdac:`) between:
`return self._execute_qblox_compilation_output(output, debug)`
and
`raise timeout`
because otherwise you are restarting the cache before the next run of `_execute_qblox_compilation_output` is executed removing the triggers again.

jjmartinezQT: Do you mind taking over @jordivallsq ? We are trying not to rollback by fixing the issue the ticket about multiple users using the Qdac introduced. I will be for the first half of the week busy in the lab. Thanks.

jordivallsq: Sure! let me address Flavie's comments and prepare some tsts for hardware

jjmartinezQT: Great, check the already existing tests in progress as well in this branch aiming to make the previous ones more completed: [github.com/qilimanjaro-tech/QHC_hardware_tests/pull/1](https://github.com/qilimanjaro-tech/QHC_hardware_tests/pull/1)

jordivallsq:
@flavie-lebars all changes requested have been implemented

jordivallsq:
tested successfully in hardware and updated https://github.com/qilimanjaro-tech/QHC_hardware_tests/pull/1

flavie-lebars (approved):
Image

jordivallsq:
The Code has been modified as Yifei found an error with simultaneously creating triggers in software loops due to the constant clearing of triggers. Now the cache is not restarted in each run, fixing the current issue this PR is fixing as well as making sure that the QDAC uses the right internal triggers.

To prevent any user from reaching the internal trigger limit (in case of a kernel restart without disconnecting the instruments), in the turn ont: he dacs defined in the runcard are whiped from any existing trigger

Since this is a huge change I rerequested PR revision from Flavie

jordivallsq:
The code has been tested on HW and an oscilloscope with these notebooks
https://github.com/qilimanjaro-tech/LabScripts/tree/main/Personal_folders/Jordi/Test_QDAC/test_2users_1qdac

flavie-lebars (approved):
Looks good to me!
One quick note though, I think for tests relating to the qdac/qblox trigger we should always have one where we purposefully leave the qblox hanging and try to reconnect after.
Especially now that we have two users at the same time, I am wondering what happens if one of the qblox misses a trigger.

PR #1165, opened 2026-07-21
Qhc 1474 add calibration to platform before execute qprogram

jordivallsq:
Added `platform.set_calibration`, which stores a `Calibration` (given an instance or file path) on the platform. `execute_qprogram`, `execute_qprograms_parallel`, and database/stream saving now fall back to it when no `calibration` argument is passed; an explicit argument always overrides it.


src/qililab/platform/platform.py
```diff
@@ -335,6 +336,9 @@ def __init__(self, runcard: Runcard):
         self._qpy_sequence_cache: dict[str, str] = {}
         """Dictionary for caching qpysequences."""
 
+        self.calibration: Calibration | None = None
+        """Calibration class, defaults to None (only used on FLUX parameters)"""
```
flavie-lebars: why is it only for FLUX parameters?

jordivallsq: This is a misspell carried from a copypaste from self.crosstalk. It should be:
```suggestion
        """Calibration class introduced to platform, defaults to None"""
```

flavie-lebars (approved):

PR #1168, opened 2026-07-23
Qhc 1455 bug readout bus without acquisition fails

flavie-lebars:
https://github.com/qilimanjaro-tech/QHC_hardware_tests/pull/3


elygoner (approved):
LGTM.

PR #1169, opened 2026-07-27
qhc-1501-fix-codecov-false-positive-failures-and-backfill-missing'

flavie-lebars:


.github/codecov.yml
```diff
@@ -1,40 +1,16 @@
...
-        # branches:
-        #   - "main"
-        flags:
-          - "unittests"
-        paths:
-          - "src"
+        target: 100%
+        threshold: 0%
 
-ignore:
```
jordivallsq, line 36: Does this mean that despite having these files ignored in CodeCov we already have tests for everything?

.github/codecov.yml
```diff
@@ -1,40 +1,16 @@
...
     project:
-      default: false # disable the default status that measures entire project
       src:
-        paths:
-          - "src"
-        target: auto # make sure coverage is not decreased (in comparison with the main branch)
-        flags:
-          - "unittests"
+        target: auto
+        threshold: 1%
```
jordivallsq, line 10: By desabling the default: false, if we remove covered code it can increase this percentage and trigger the codecov error because it exceeds the limit

jordivallsq (commented):
The changes implemented to code coverage look great as well as the new platform tests!

I left a comment on the default for project and a small question

.github/codecov.yml
jordivallsq: nevermind I read your slack reply

.github/codecov.yml
DZPM: 1% is reasonable for a project this big

DZPM (commented):
`after_n_builds` is still unset, the ticket explicitly asked to check it

tests/platform/test_platform.py
```diff
@@ -2044,6 +2111,16 @@ def test_db_real_time_saving(self, platform: Platform):
...
 
+    def test_db_real_time_saving_without_db_manager_raises_error(self, platform: Platform):
+        """Test db_real_time_saving raises a ReferenceError when no db_manager has been loaded."""
+        shape = (2, 2)
+        loops = {"test_amp_loop": np.arange(0, 2)}
+        experiment_name = "test_db_real_time_saving"
+        platform.db_manager = None
+
+        with pytest.raises(ReferenceError, match="Missing db_manager, try using platform.load_db_manager()."):
+            platform.db_real_time_saving(shape, loops, experiment_name)
```
DZPM: This tests `db_real_time_saving`.

`db_real_time_saving()` raises the same error, but it's not tested.

A `parametrize` could close both-

.github/codecov.yml
flavie-lebars: @jordivalls this is correct, I have added the default to False back, thanks!

jordivallsq (approved):
Great work! All comments properly addressed!

PR #1170, opened 2026-07-29
Qhc 1499 fix distortions implementation with optimized square pulses

jordivallsq:
- Added `QProgram.with_distortions`, which returns a copy of the `QProgram` with the bus pulse distortions applied to the played waveforms. Distortions are now applied at compile time inside `QbloxCompiler`, on the optimized waveforms produced by the compiler (e.g. square pulses), instead of being applied to the already-compiled sequences at execution time. This also fixes a bug where distortions were only applied to a single bus instead of every bus that defines them in the runcard.

- Added `PulseDistortion.amplitude_gain`, which returns the peak amplification that a distortion's correction filter applies to a given envelope (before normalization). This is useful when a distorted pulse is normalized back into the sequencer range (e.g. with `auto_norm=True`): the sequencer gain must be multiplied by this factor to reach the same physical amplitude, and the maximum reachable amplitude for a unit-height pulse is `1 / amplitude_gain(unit_pulse)`. The raw correction filter of each distortion is now exposed through a `_filter` method, with `apply` becoming a shared `_filter` + `normalize_envelope` step in the base class.


docs/releases/changelog-dev.md
```diff
@@ -2,6 +2,12 @@
 
 ### New features since last release
 
+- Added `QProgram.with_distortions`, which returns a copy of the `QProgram` with the bus pulse distortions applied to the played waveforms. Distortions are now applied at compile time inside `QbloxCompiler`, on the optimized waveforms produced by the compiler (e.g. square pulses), instead of being applied to the already-compiled sequences at execution time. This also fixes a bug where distortions were only applied to a single bus instead of every bus that defines them in the runcard.
```
flavie-lebars: I find this confusing,
My understanding is that now we apply the distortion and then compile. So the distortions are NOT applied on the optimized waveforms

flavie-lebars: those are bug fixes, not new features

docs/releases/changelog-dev.md
```diff
@@ -2,6 +2,12 @@
...
+  [#1170](https://github.com/qilimanjaro-tech/qililab/pull/1170)
+
+- Added `PulseDistortion.amplitude_gain`, which returns the peak amplification that a distortion's correction filter applies to a given envelope (before normalization). This is useful when a distorted pulse is normalized back into the sequencer range (e.g. with `auto_norm=True`): the sequencer gain must be multiplied by this factor to reach the same physical amplitude, and the maximum reachable amplitude for a unit-height pulse is `1 / amplitude_gain(unit_pulse)`. The raw correction filter of each distortion is now exposed through a `_filter` method, with `apply` becoming a shared `_filter` + `normalize_envelope` step in the base class.
```
flavie-lebars: if we are exposing filter why is it private?

src/qililab/qprogram/qprogram.py
```diff
@@ -1042,6 +1043,50 @@ def handle_crosstalk_element(block: Block, element_list: list[int], flux_vector:
...
+        Returns:
+            QProgram: A new instance of QProgram with the distortions applied to the affected
+            waveforms. The original QProgram is left unchanged.
+        """
+
+        def traverse(block: Block):
+            for index, element in enumerate(block.elements):
+                if isinstance(element, Block):
+                    traverse(element)
+                elif isinstance(element, (Play, Measure, MeasureReset)):
```
flavie-lebars, line 1079: There is a bug for the MeasureReset. MeasureReset is unpacked into a Play etc.. at compile time. So any distortions would not actually be applied to Play.
We don't need distortions for the MeasureReset so I suggest raising a NotImplementedError to not create unecessary complication.

src/qililab/qprogram/qprogram.py
```diff
@@ -1042,6 +1043,50 @@ def handle_crosstalk_element(block: Block, element_list: list[int], flux_vector:
...
+                    if element.bus in bus_distortions.keys():
+                        waveform = element.waveform
+                        for distortion in bus_distortions[element.bus]:
+                            if isinstance(waveform, IQWaveform):
```
flavie-lebars, line 1089: I think the if is a bit weak. I would suggest if iqwaveform, then elif waveform. And maybe raise an error for anything else

tests/qprogram/test_qprogram.py
```diff
@@ -9,6 +9,7 @@
 from qililab.qprogram.blocks import Average
 from qililab.qprogram.calibration import Calibration
 from qililab.qprogram.crosstalk_matrix import CrosstalkMatrix, NonLinearCrosstalkMatrix
+from qililab.pulse_distortion import ExponentialCorrection
```
flavie-lebars, line 13: the distortions are now applied on the compiler there should be a test in test_qblox_compiler to check this

docs/releases/changelog-dev.md
flavie-lebars: the refactor of apply and filter is a breaking change. And this needs to be communicated very clearly to experimentalists.

flavie-lebars (changes requested):
Great work Jordi,
I have left some comments.
I understand we need to merge quickly so I won't ask for the drawer to be included here. But could you please create a linear task to ensure that `qp.draw` and `platform.draw()` can accept distortions. And also that the liveplot feeds the distortions to the drawer.

src/qililab/qprogram/qprogram.py
```diff
@@ -1042,6 +1043,50 @@ def handle_crosstalk_element(block: Block, element_list: list[int], flux_vector:
...
+            QProgram: A new instance of QProgram with the distortions applied to the affected
+            waveforms. The original QProgram is left unchanged.
+        """
+
+        def traverse(block: Block):
+            for index, element in enumerate(block.elements):
+                if isinstance(element, Block):
+                    traverse(element)
+                elif isinstance(element, (Play, Measure, MeasureReset)):
+                    if element.bus in bus_distortions.keys():
```
flavie-lebars: And please fix the sonar cloud issue
Remove this unnecessary "keys()" call.
Check warning: Remove this unnecessary "keys()" call.

src/qililab/qprogram/qprogram.py
flavie-lebars: fixed it

docs/releases/changelog-dev.md
jordivallsq: exactly

flavie-lebars: I have fixed the changelog.

docs/releases/changelog-dev.md
jordivallsq: This definitely is not a good choice of words. The functions exposed are `apply` (works like the previous implementation, therefore not breaking any previous work) and `amplitude_gain`.
Both call for _filter internally but the user doesn't need to be exposed to the user unless you think othewise

src/qililab/qprogram/qprogram.py
jordivallsq: Agree a NotImplementedError should be best

src/qililab/qprogram/qprogram.py
flavie-lebars: fixed.

docs/releases/changelog-dev.md
jordivallsq: the refactor is just to get the `_filter` outside apply only to be called by `apply` and `amplitude_gain`. The only new thing is `amplitude_gain`, apply should work exactly as it did

flavie-lebars (approved):

jjmartinezQT (approved):
Looks good to me, thanks!

PR #1171, opened 2026-07-29
qhc-1518-qililab-release-0333

flavie-lebars:


jjmartinezQT (approved):

elygoner (approved):
change the version in the __init__.py line 23 too.

flavie-lebars:
@elygoner we should not have the release there, I am not sure when this was added.
I have reverted it back to 0.0.0 as I think it should be

PR #1172, opened 2026-07-29
Add publish.yml workflow

fedonman:


flavie-lebars (commented):
Hi @fedonman ,
I have quick questions on some differences with the template provided by Platform:

1. Why did we drop the automatically release trigger?
2. Why has ubuntu-latest been hardcoded instead of using the shared RUNS_ON setting?
3. The deploy-codeartifacts step references the shared action as qilimanjaro-tech/devops/.github/actions/code-artifact-publish@main instead of a pinned version, why? (Sonar Cloud raises this as well)

PR #1173, opened 2026-07-30
[QHC-1525] Add parameter deletion to `NonLinearCrosstalk.set_non_linear_params`

elygoner:
Used a Sentinel to mark the difference between setting parameters to None and not setting anything. This was done by creating a Sentinel.
Because of the type checking, i decided to implement using a `Final` typed variable rather than doing `isinstance` (to increase code legibility).

I have also added a condition so that if you want to set the beta-sin parameters, they have to be both None or both Values. This is to preserve the previous logic.

One thing that is still pending is to define where the sentinel should go, because even if it only used in the NonLinearCrosstalk for now, I don't think it belongs there.


src/qililab/qprogram/crosstalk_matrix.py
```diff
@@ -21,6 +22,21 @@
 from qililab.yaml import yaml
 
 
+class _Sentinel(enum.Enum):
```
jordivallsq: is the sentinel class name necessary or can it be more descriptive?

elygoner: I thought of making it more general, so if we want, in the future, to make another sentinel for Uniniciated parameters or whatever, we don't have to make another class.
If you think we won't really expand on this I can make it more specific.

jordivallsq: This is a very good reason actually, so let's leave it as it is

jordivallsq (approved):
The tests look reasonable (NO changes required on old tests and feature correctly added), and the changes solid. Good work!

docs/releases/changelog-dev.md
```diff
@@ -14,3 +14,6 @@
 ### Documentation
 
 ### Bug fixes
+
+- Make it so if you input None in `NonLinearCrosstalkMatrix.set_non_linear_params` it sets the parameters to None, instead of ignoring them.
```
flavie-lebars: can you please rephrase the changelog. It is not clear, and the Sentinel is not mentioned.

src/qililab/qprogram/crosstalk_matrix.py
```diff
@@ -21,6 +22,21 @@
...
+    on ``is`` / ``is not``. Add a member here plus its value and ``Literal`` alias below.
+    """
+
+    UNSET = enum.auto()
+
+    def __repr__(self) -> str:
+        return f"<{self.name}>"
+
+
+UNSET: Final = _Sentinel.UNSET
```
flavie-lebars: I think we should have `_UNSET` instead, it should not be accessed by a qililab user.

src/qililab/qprogram/crosstalk_matrix.py
```diff
@@ -285,33 +301,39 @@ def set_non_linear_params(
...
-            junction_asym (float): Junction asymmetry, d ∈ [-1, 1].
+            beta_c (float | None): Bessel modulation parameter beta_c. Must be non-zero.
+            amplitude (float | None): Amplitude of the nonlinear correction in flux units.
+            junction_asym (float | None): Junction asymmetry, d ∈ [-1, 1].
 
         Raises:
             ValueError: If either bus is not present in the matrix.
             ValueError: If beta_c is zero, which would cause a division by zero in the
                 Bessel expansion.
+            ValueError: If "amplitude" and "beta_c aren't both; set to a value, set to None or unset.
```
flavie-lebars: please rephrase

src/qililab/qprogram/crosstalk_matrix.py
flavie-lebars: I agree with making it general, but if it is general it should not be in crosstalk_matrix, I would move it to a sentinels file in utils. And we keep _UNSET defined in crosstalk_matrix.

src/qililab/qprogram/crosstalk_matrix.py
```diff
@@ -285,33 +301,39 @@ def set_non_linear_params(
...
         """
         for bus in (bus_i, bus_j):
             if bus not in self.matrix:
                 raise ValueError(f"Bus '{bus}' not present in the crosstalk matrix.")
 
-        if beta_c is not None or amplitude is not None:
-            if not (beta_c is not None and amplitude is not None):
+        if beta_c is not UNSET or amplitude is not UNSET:
+            if not (beta_c is not UNSET and amplitude is not UNSET):
```
flavie-lebars: I am aware this was not implemented in this PR but can you remove the triple negative please:
`if beta_c is UNSET or amplitude is UNSET:` instead of `if not (beta_c is not UNSET and amplitude is not UNSET):`

src/qililab/qprogram/crosstalk_matrix.py
```diff
@@ -285,33 +301,39 @@ def set_non_linear_params(
...
                 raise ValueError(
                     "Both 'amplitude' and 'beta_c' must be provided together — you cannot specify one without the other."
                 )
+            if ((beta_c is None) + (amplitude is None)) == 1:
```
flavie-lebars:
```suggestion
            if (beta_c is None) != (amplitude is None):
```

flavie-lebars (changes requested):

docs/releases/changelog-dev.md
elygoner: I will add the sentinels in Improvements

docs/releases/changelog-dev.md
```diff
@@ -7,6 +7,9 @@
 
 ### Improvements
 
+- Added the `_Sentinels` enum class in `utils/sentinels.py`, currently exposing only `UNSET` (more sentinels can be added later). Sentinels mark uninitialized, unset or otherwise undefined values without relying on `None`, which keeps the logic clearer and makes it possible to distinguish an explicit `None` from an unset default.
```
flavie-lebars: Now that `Sentinel` is meant to be accessed by other modules, it should no longer be private

flavie-lebars: typo: the class is _Sentinel

src/qililab/qprogram/crosstalk_matrix.py
```diff
@@ -285,33 +287,40 @@ def set_non_linear_params(
...
-            junction_asym (float): Junction asymmetry, d ∈ [-1, 1].
+            beta_c (float | None): Bessel modulation parameter beta_c. Must be non-zero.
+            amplitude (float | None): Amplitude of the nonlinear correction in flux units.
+            junction_asym (float | None): Junction asymmetry, d ∈ [-1, 1].
 
         Raises:
             ValueError: If either bus is not present in the matrix.
             ValueError: If beta_c is zero, which would cause a division by zero in the
                 Bessel expansion.
+            ValueError: If both "amplitude" and "beta_c aren't set to the same type of value.
```
flavie-lebars: can you close the " after beta_c this migh cause problem with some documentation

flavie-lebars (approved):
Image

PR #1174, opened 2026-08-05
qhc-1531-add-dependabot

flavie-lebars:


.github/dependabot.yml
```diff
@@ -0,0 +1,46 @@
...
+    groups:
+      python-minor-patch:
+        applies-to: version-updates
+        patterns: ["*"]
+        update-types: ["minor", "patch"]
+    commit-message:
+      prefix: "deps"
+      prefix-development: "deps-dev"
+    open-pull-requests-limit: 10
+    labels: ["dependencies", "python"]
```
jordivallsq, line 46: Pytthon does not exist as a standalone label, you should try with python:uv

```suggestion
    labels: ["dependencies", "python:uv"]
```

.github/dependabot.yml
```diff
@@ -0,0 +1,46 @@
...
+    groups:
+      actions-minor-patch:
+        applies-to: version-updates  # default, stated so the security path stays visible
+        patterns: ["*"]
+        update-types: ["minor", "patch"]
+    commit-message:
+      prefix: "ci"
+    # Majors arrive as one PR each now, so leave headroom above the default of 5.
+    open-pull-requests-limit: 10
+    labels: ["dependencies", "github-actions"]
```
jordivallsq, line 27: Same here for github-actions (the label does not exist in the repo), you should create a github-actions label in the repo and that should fix it

jordivallsq (approved):
The PR looks good! There are two labels that do not exist at https://github.com/qilimanjaro-tech/qililab/labels please create them in the repo (but for `python` label I would change it to the existing `python:uv`

.github/dependabot.yml
flavie-lebars: I will keep python to make sure all libraries at qili have the same config. Instead I will delete the python:uv label and create the python label

jordivallsq: Makes sense, thank you!

flavie-lebars:
This should not be merged until the AWS access is figured out.

PR #1175, opened 2026-08-06
Qhc 1527 crosstalk matrix for dc and fast bias lines

jordivallsq:
Added a second crosstalk matrix, `Calibration.crosstalk_matrix_ac`, so DC and AC/fast-flux lines can be calibrated independently. Fluxoniums can be driven simultaneously through DC lines (Qdac, via the existing `crosstalk_matrix`) and fast-flux lines (QCM), which have different resistances and therefore need separate crosstalk calibrations. The routing now follows the instrument type: `QdacCompiler` keeps using `crosstalk_matrix` (DC), while `QbloxCompiler` uses `crosstalk_matrix_ac` (AC/fast flux). For backward compatibility, when `crosstalk_matrix_ac` is not set the Qblox compiler falls back to `crosstalk_matrix` and logs a warning.


jordivallsq:
Added @jjmartinezQT and @OscarQili to verify if the implementation is what we need

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -343,8 +343,14 @@ def traverse(block: Block):
             self._qprogram = self._qprogram.with_bus_mapping(bus_mapping=bus_mapping)
         if calibration is not None:
             self._qprogram = self._qprogram.with_calibration(calibration=calibration)
-            if calibration.crosstalk_matrix and crosstalk is None:
-                crosstalk = calibration.crosstalk_matrix
+            if crosstalk is None:
```
jjmartinezQT: I believe compile_qprogram in platform needs some changes as for example is only considering DC crosstalk matrix when clearing it and not the new AC.

jjmartinezQT (dismissed):

src/qililab/qprogram/qblox_compiler.py
jordivallsq: You are right; when not using crosstalk, it only removes `crosstalk_matrix` and not `crosstalk_matrix_ac`. Fixed it

jjmartinezQT:
## Code Review

### Confirmed: `crosstalk=False` doesn't clear `crosstalk_matrix_ac`

`Platform.compile_qprogram` (`platform.py:1243-1245`) only clears `calibration.crosstalk_matrix` when `crosstalk=False`:

```python
if not crosstalk and calibration is not None and calibration.crosstalk_matrix is not None:
    calibration = deepcopy(calibration)
    calibration.crosstalk_matrix = None
```

It never touches the new `calibration.crosstalk_matrix_ac`. Since `crosstalk` then resolves to `None` when passed into `QbloxCompiler.compile`, the new fallback logic in this PR picks up `crosstalk_matrix_ac` anyway — directly contradicting `execute_qprogram`'s own docstring for this flag ("ignores all existing crosstalk matrices").

**Repro:** set `calibration.crosstalk_matrix_ac`, call `platform.execute_qprogram(qp, calibration=calibration, crosstalk=False)`. Qblox compilation still applies AC crosstalk compensation.

### A second, broader gap with the same root cause

`Platform.compile_qprogram` passes crosstalk to both compilers as:
```python
crosstalk=self.crosstalk if crosstalk else None,   # line 1290 (qdac) and 1347 (qblox)
```
`self.crosstalk` is a pre-existing, single, DC-oriented platform attribute set via the public `platform.set_crosstalk(matrix)` API (used for `set_flux_to_zero`/`set_bias_to_zero`/`flux_vector` — all DC concepts). It's a heavily-used, well-established API (a dozen+ call sites already in `test_platform.py`).

Whenever `self.crosstalk` is set — which is the case for any platform already using this existing API, and is the *default* path since `crosstalk=True` by default — it gets passed into `QbloxCompiler.compile(..., crosstalk=<explicit matrix>)`. Since that parameter is then non-`None`, the entire new AC/DC fallback block (`if crosstalk is None: ...`) added in this PR is skipped entirely.

**Net effect:** any platform using the existing `set_crosstalk()` API can never benefit from `crosstalk_matrix_ac` — Qblox/fast-flux lines silently get the DC matrix applied instead, which is exactly the "different resistances need separate calibrations" problem this PR sets out to fix.

None of the new tests catch this because they all call `QbloxCompiler().compile(...)` directly (e.g. `test_crosstalk_compensation_uses_ac_matrix`), bypassing `Platform.compile_qprogram` entirely — so the interaction with `self.crosstalk` is untested in either direction.

**Suggested fix:** both issues share the same root cause — `Platform.compile_qprogram` resolves a single combined crosstalk value *before* calling either compiler and passes it explicitly, which defeats the compiler-level AC/DC fallback added here regardless of which path triggers it. Probably cleanest to make `compile_qprogram` itself AC/DC-aware (e.g. resolve DC vs AC crosstalk per-compiler the same way `QbloxCompiler.compile` now does, rather than resolving one value for both), instead of patching each symptom individually.

### Minor / possibly out of scope

`Calibration.add_intra_crosstalk`/`add_inter_crosstalk` (the fitting-refinement methods) only ever read/write `crosstalk_matrix`, never `crosstalk_matrix_ac`. Plausibly intentional — calibration fitting from bias/flux measurements is inherently a DC concept — but worth a one-line confirmation that AC-matrix refinement is deliberately out of scope here rather than an oversight, since it's an easy thing to forget as a follow-up.

---
*Automated review by Claude Code.*

jordivallsq:
Regarding Claude's review:
- The first issue has already been fixed (replied in your comment). Maybe you used the code before the latest change?

- Regarding the second statement, it is wrong; self.crosstalk is a `CrosstalkMatrix` class that does not depend on `Calibration`. As you can see in `calibration.py`, both `crosstalk_matrix` and `crosstalk_matrix_ac` are `CrosstalkMatrix`, so you can give `platform.set_crosstalk(calibration.crosstalk_matrix_ac)` as any other matrix without restriction. The statement for: 
> used for `set_flux_to_zero`/`set_bias_to_zero`/`flux_vector` — all DC concepts

Is straight up nonsensical; all operations related to the flux are DC by nature (not modulated), and therefore there is nothing wrong with these functions (if we use Qblox for the flux, the parameters set are Qblox's parameters selected automatically). In summary, to use the `crosstalk_matrix_ac` in `set_crosstalk`, just use `platform.set_crosstalk(calibration.crosstalk_matrix_ac)`.

- For the final minor point, this one makes more sense. Right now, the crosstalk history (used for crosstalk calibration only uses `crosstalk_matrix`. This is because when we discussed this, we concluded that the calibration was done in the DC lines (due to the complexity of flux v flux in non-linear compensation in Qblox). I would need @OscarQili advice to decide if it is worth adding a second crosstalk history. Keeping in mind that as long as we don't need to calibrate both crosstalks simultaneously, we can use two different calibration files for the DC and AC calibration. (Then, for common usage, we will need only one calibration file with both matrices)

jjmartinezQT:
> Regarding Claude's review:
> 
> * The first issue has already been fixed (replied in your comment). Maybe you used the code before the latest change?
> * Regarding the second statement, it is wrong; self.crosstalk is a `CrosstalkMatrix` class that does not depend on `Calibration`. As you can see in `calibration.py`, both `crosstalk_matrix` and `crosstalk_matrix_ac` are `CrosstalkMatrix`, so you can give `platform.set_crosstalk(calibration.crosstalk_matrix_ac)` as any other matrix without restriction. The statement for:
> 
> > used for `set_flux_to_zero`/`set_bias_to_zero`/`flux_vector` — all DC concepts
> 
> Is straight up nonsensical; all operations related to the flux are DC by nature (not modulated), and therefore there is nothing wrong with these functions (if we use Qblox for the flux, the parameters set are Qblox's parameters selected automatically). In summary, to use the `crosstalk_matrix_ac` in `set_crosstalk`, just use `platform.set_crosstalk(calibration.crosstalk_matrix_ac)`.
> 
> * For the final minor point, this one makes more sense. Right now, the crosstalk history (used for crosstalk calibration only uses `crosstalk_matrix`. This is because when we discussed this, we concluded that the calibration was done in the DC lines (due to the complexity of flux v flux in non-linear compensation in Qblox). I would need @OscarQili advice to decide if it is worth adding a second crosstalk history. Keeping in mind that as long as we don't need to calibrate both crosstalks simultaneously, we can use two different calibration files for the DC and AC calibration. (Then, for common usage, we will need only one calibration file with both matrices)

Makes sense.

flavie-lebars (changes requested):
Claude is giving me this bug saying it is not backward compatible. Have you checked with a previously generated calibration file?
```
"""
Repro: loading a pre-existing calibration.yml (saved before crosstalk_matrix_ac
existed) crashes the Qblox compiler instead of falling back gracefully, despite
the changelog's backward-compatibility claim.
"""

import numpy as np

from qililab.qprogram.calibration import Calibration
from qililab.qprogram.qblox_compiler import QbloxCompiler
from qililab.qprogram.qprogram import QProgram
from qililab.qprogram.crosstalk_matrix import CrosstalkMatrix
from qililab.utils.serialization import serialize_to, deserialize_from
from qililab.waveforms import Square

OLD_CALIBRATION_YML = "old_calibration.yml"

# --- Step 1: build a calibration the way a user would have BEFORE this branch ---
# (only crosstalk_matrix existed back then)
old_style_calibration = Calibration()
old_style_calibration.crosstalk_matrix = CrosstalkMatrix().from_array(
    ["flux1", "flux2"], np.array([[1, 0.1], [0.1, 1]])
)

serialize_to(old_style_calibration, OLD_CALIBRATION_YML)

with open(OLD_CALIBRATION_YML) as f:
    print("--- contents of the 'old' calibration.yml (note: no crosstalk_matrix_ac key) ---")
    print(f.read())

# Manually strip the key to simulate a file genuinely written pre-branch
# (serialize_to on current code always includes it as `null`; a real old file wouldn't have the line at all)
with open(OLD_CALIBRATION_YML) as f:
    content = f.read()
content = "\n".join(line for line in content.splitlines() if "crosstalk_matrix_ac" not in line) + "\n"
with open(OLD_CALIBRATION_YML, "w") as f:
    f.write(content)

# --- Step 2: load it the way platform.set_calibration(path) does ---
loaded_calibration = deserialize_from(OLD_CALIBRATION_YML, Calibration)

print("has crosstalk_matrix_ac attribute?", hasattr(loaded_calibration, "crosstalk_matrix_ac"))

# --- Step 3: compile a QProgram against it, same as a normal experiment run ---
qp = QProgram()
qp.play(bus="flux1", waveform=Square(amplitude=1.0, duration=40))

print("\nCompiling with the loaded (old-format) calibration...")
QbloxCompiler().compile(qprogram=qp, calibration=loaded_calibration)
print("compiled OK (bug is fixed)")

```

flavie-lebars:
One other point, for the inter and intra crosstalk, even if Oscar says that there is no need to support this, please make sure the error is clear, at the moment it says: "ValueError: No crosstalk has been given to the Calibration file", it should say that ac crosstalk is not accepted.

jordivallsq:
Right, there was a bug because old calibration files don't go through the init, now the bug is fixed and the code you suggested runs without issues

src/qililab/platform/platform.py
```diff
@@ -1240,9 +1240,12 @@ def compile_qprogram(
         calibration: Calibration | None = None,
         crosstalk: bool = True,
     ) -> QProgramCompilationOutput:
-        if not crosstalk and calibration is not None and calibration.crosstalk_matrix is not None:
+        if not crosstalk and calibration is not None:
             calibration = deepcopy(calibration)
-            calibration.crosstalk_matrix = None
+            if calibration.crosstalk_matrix is not None:
```
flavie-lebars: is this not equivalent to doing?
```
if not crosstalk and calibration is not None:
    calibration = deepcopy(calibration)
    calibration.crosstalk_matrix = None
    calibration.crosstalk_matrix_ac = None
```

src/qililab/qprogram/qblox_compiler.py
```diff
@@ -343,8 +343,14 @@ def traverse(block: Block):
...
+                if calibration.crosstalk_matrix_ac is not None:
+                    crosstalk = calibration.crosstalk_matrix_ac
+                elif calibration.crosstalk_matrix is not None:
+                    logger.warning(
```
flavie-lebars: This means that for every program compiled (ie: each iteration of a sw loop), the warning is logged. I understand we need to merge quickly but could you please create a task so that it is addressed later. One way around this could be to cache the warning

src/qililab/platform/platform.py
```diff
@@ -1240,9 +1240,12 @@ def compile_qprogram(
         calibration: Calibration | None = None,
         crosstalk: bool = True,
     ) -> QProgramCompilationOutput:
-        if not crosstalk and calibration is not None and calibration.crosstalk_matrix is not None:
+        if not crosstalk and calibration is not None:
```
flavie-lebars: here there is a set back in efficiency. Before, deepcopy would only run if a calibration with crosstalk was given. Now, the deepcopy is done when a calibration is passed regardless of whether a crosstalk is included. This should not the be the case, this could introduce a slow down in software loops.

flavie-lebars (changes requested):
1. Can you please check this comment from Claude:"      "file": "src/qililab/platform/platform.py",
      "line": 1350,
      "summary": "Platform.compile_qprogram forwards the platform-level self.crosstalk (DC) matrix as the explicit crosstalk argument to QbloxCompiler.compile, which bypasses the PR's new calibration.crosstalk_matrix_ac fallback logic entirely.",
      "failure_scenario": "If a user calls platform.set_crosstalk(dc_matrix) (used elsewhere for _process_crosstalk/set_flux_to_zero/set_bias_to_zero) and then compiles with the default crosstalk=True, crosstalk=self.crosstalk evaluates to the non-None dc_matrix, so QbloxCompiler.compile's 'if crosstalk is None' branch never runs and calibration.crosstalk_matrix_ac is silently ignored — defeating the PR's headline AC-crosstalk feature with no warning. Untested by the new tests, which never set self.crosstalk.",
      "category": "correctness",
      "short_summary": "self.crosstalk shadows crosstalk_matrix_ac fallback"
    },
    {"

2. I had left this comment before "One other point, for the inter and intra crosstalk, even if Oscar says that there is no need to support this, please make sure the error is clear, at the moment it says: "ValueError: No crosstalk has been given to the Calibration file", it should say that ac crosstalk is not accepted.", I don't think it has been addressed yet.

src/qililab/platform/platform.py
jordivallsq: agree, simplified the code with this (+ moving the necessary conditions to the if)

src/qililab/platform/platform.py
jordivallsq: This is solved by moving 
`if calibration.crosstalk_matrix is not None`
and 
`if calibration.crosstalk_matrix_ac is not None:`
to the condition, now the deepcopy is only created if a crosstalk is given through calibration

src/qililab/qprogram/qblox_compiler.py
jordivallsq: Now the warning is raised once per run. Added tests as well.
I can roll back this change if you think a new PR is in order but I think that the changes are straightforward enough to be here.

jordivallsq (commented):
@flavie-lebars Implemented the requested changes, regarding your comments:

1. Fixed the shadowing of `crosstalk_matrix_ac` in the Qblox_Compiler compile function (this was because platform crosstalk took priority over `calibration.crosstalk_matrix_ac`, now is the other way around (fixed in the conditions of the compiler: src/qililab/qprogram/qblox_compiler.py line 347).
2. Apologies for that I thought I addressed everything. Added the comments in the error message.

src/qililab/qprogram/calibration.py
```diff
@@ -205,7 +219,7 @@ def add_intra_crosstalk(
             ValueError: If no crosstalk has been given to calibration file.
         """
         if not self.crosstalk_matrix:
-            raise ValueError("No crosstalk has been given to the Calibration file")
+            raise ValueError("No crosstalk has been given to the Calibration file. AC Crosstalk is not accepted.")
```
flavie-lebars: I thought we would give a different error, otherwise I find it confusing. But this is a minor comment, I will not block the merge because of this.

flavie-lebars (approved):
Image

src/qililab/qprogram/calibration.py
jordivallsq: Right. Is the solution implemented better?

flavie-lebars: yes!

flavie-lebars: thank you

flavie-lebars:
I just checked the cache for the warning and atcually there is a problem. It would only reset when restarting the kernel. This is a problem with the way they currently run experiments, they connect once to the platform and then they send multiple experiments. Currently the warning would only be raised for the first iteration of the first experiment.

src/qililab/qprogram/qblox_compiler.py
jordivallsq: Reverted this change and creating PR to fix it

jordivallsq:
The cache implementation has been removed and a PR has been created 
https://linear.app/qilimanjaro/issue/QHC-1547/add-cache-to-warnings-and-message-to-execute-qprogram-for-1-warning

PR #1176, opened 2026-08-07
add scipy

flavie-lebars:


uv.lock
```diff
@@ -4386,7 +4413,7 @@ name = "tbb"
 version = "2022.3.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
-    { name = "tcmlib", marker = "platform_machine != 'aarch64' and sys_platform != 'darwin'" },
```
elygoner: I guess qilisdk droped the arm support?

elygoner (approved):
It sure is a dependency change. LGTM. 
But the instalation seems to be breacking because the uv lock is not updated in some way.

uv.lock
flavie-lebars: the PR has a lot of problems, I am still looking at it

flavie-lebars:
This PR will be reopened once the qilisdk dependency is removed from qililab, then we should be able to do  "scipy>=1.17.1"

flavie-lebars:
@elygoner I have had to modiy the changelog and crosstalk matrix due to a new ruff revision

elygoner (approved):
L(mostly)GTM. 
I see that arm is  back to being supported and that you are indexing numpy-mkl for the scipy source. 
One thing I noticed is that the index is marked as not usable for darwin and aarch64. darwin has an explicit versioning constraint in dependencies but aarch64 doesn't.
Can you clarify if that is intended. I don't thing it is much of an issue since we don't have users using that platform and we are decoupling from qiliSDK so we don't have to support it. But if it is a quick fix It would be better if we can get it done. (or maybe i am understanding it wrong and that is not the case, that could also be a possiblility)

flavie-lebars:
@elygoner nice catch!
I will go ahead and merge now though, since I don't think I can have a different requirement from qilisdk otherwise `uv lock` would fail.
I will let the qilisdk team decides how they want to handle this @amirazzam2000 @fedonman

PR #1184, opened 2026-08-11
release 0.34

flavie-lebars:


elygoner (approved):
Since the AWS for qpysequence was already sdone it makes sense that the pyproject and the uv.lock are only version bumps, so LGTM.

PD the random changelog line in de middle of CHANGELOG.md is crazy

PR #1186, opened 2026-08-11
Qhc 1494 add add fitting and the fitting and fitting path columns to

jordivallsq:
Added `fitting_parameters` to autocalibration `AutocalMeasurement` database table, mirroring the existing `Measurement` implementation. Fitting results can be attached at three levels:
- Using the `AutocalMeasurement` table with `meas.add_fitting(db_manager, path, parameters)`.
- Using `DatabaseManager` with `db_manager.add_calibration_fitting(id, path, parameters)`, loading the measurement by its `measurement_id` (ideal for adding fits in a later session).
- Using `StreamArray` with `stream_array.add_fitting(path, parameters)`, which now supports autocalibration measurements as well.


src/qililab/result/database/database_autocal.py
```diff
@@ -171,5 +175,36 @@ def update_platform(self, session: sessionmaker[Session], platform: "Platform"):
                 running_session.rollback()
                 raise e
 
+    def add_fitting(self, database_manager: "DatabaseManager", path: str, parameters: dict[str, Any] | None = None):
```
elygoner: Can you please make the path optional.

src/qililab/result/database/database_manager.py
```diff
@@ -300,6 +300,21 @@ def load_calibration_by_id(self, id: int) -> AutocalMeasurement | None:
 
             return measurement_by_id
 
+    def add_calibration_fitting(
```
elygoner, line 303: well talk about it more for the homogenization, but is there any way of detecting what is the format of the DB instead of having to use different functions?

elygoner (commented):
I left some comments.

src/qililab/result/database/database_autocal.py
jordivallsq: What is the reason behind this? I have no issue in changing this if needed but rght now is using the same format as the measurements database and see no reason to change it

jordivallsq:
Applied this to the database
```
ALTER TABLE IF EXISTS public.measurements
    ADD COLUMN fitting_parameters jsonb;
```

src/qililab/result/database/database_manager.py
jordivallsq: I can create a linear task for this specific change, but I see this as part of the homogenization changes on databases. I don't see any point in doing a convoluted check right in this PR.

src/qililab/result/database/database_autocal.py
elygoner: the fitting and the plot in seqtante are two different steps, the fit adds the fitting and the plot, if it generates an image it adds the fitting path.

src/qililab/result/database/database_manager.py
elygoner: of course

src/qililab/result/database/database_autocal.py
jordivallsq: OK, now path is optional

tests/result/test_database.py
```diff
@@ -847,6 +906,22 @@ def test_load_calibration_by_id_path_not_found(self, db_manager: DatabaseManager
...
+        mock_measurement.add_fitting.return_value = mock_measurement
+
+        with patch.object(db_manager, "load_calibration_by_id", return_value=mock_measurement) as mock_load:
+            result = db_manager.add_calibration_fitting(123, path="/test/fit.h5", parameters={"a": 1.0})
+
+        mock_load.assert_called_once_with(123)
+        mock_measurement.add_fitting.assert_called_once_with(db_manager, "/test/fit.h5", {"a": 1.0})
+        assert result == mock_measurement
+
+    def test_add_calibration_fitting_raises_exception_measurement_not_found(self, db_manager: DatabaseManager):
```
elygoner, line 920: obviously this doesn't go in this PR but i think `get_by_id`should error if it doesn't find it instead of returning None.
I don't know in what context would one prefer to get None instead of an error when trying to get a measurement from the DB that isn't there.

elygoner (approved):

tests/result/test_database.py
jordivallsq: added to another linear task.
https://linear.app/qilimanjaro/issue/QHC-1599/unify-structure-of-measurement-and-autocalmeasurement-database-classes

jordivallsq:
Test in a notebook similar to https://github.com/qilimanjaro-tech/QHC_hardware_tests/blob/test-qhc-1164-qdac-qblox-wait-trigger/tests/qililab/test_PR_1136.ipynb with some modifications.

imagen

PR #1187, opened 2026-08-12
release 0.35

flavie-lebars:
The `pyproject.toml` has been modified because the default for `towncrier` is to not have the full desciption for misc changes. I have overriden this. All fields must be overriden if one overrides the default.


elygoner (approved):
LGTM

PR #1191, opened 2026-08-17
Qhc 1553 bugfix nonlinearcrosstalk adds offsets to lines without nl

jordivallsq:
Fixed a spurious offset appearing on the non-swept flux buses when sweeping a flux with QDAC ramps under a `NonLinearCrosstalkMatrix`, so QDAC-ramp sweeps no longer disagree with the equivalent software-loop sweeps. `Platform.compile_qprogram`, `QdacCompiler.compile` and `QProgram.with_crosstalk_qdac` now seed the crosstalk parked point from the target fluxes (set via `set_parameter(FLUX)`) instead of reconstructing them from the hardware bias voltage.


src/qililab/platform/platform.py
```diff
@@ -1124,6 +1124,10 @@ def compile_qprogram(
         self.qdac_buses = [
             bus for bus in self.buses if any(isinstance(instrument, QDevilQDac2) for instrument in bus.instruments)
         ]
+        # Known target fluxes (set via set_parameter(FLUX)).
+        flux_vector = getattr(self, "flux_vector", None)
```
flavie-lebars: Problem here is if the user doesn't do platform.set_parameter but sets everything via the runcard. Then, flux_vector is not updated

src/qililab/qprogram/qprogram.py
```diff
@@ -976,7 +985,14 @@ def traverse(block: Block, flux_vector: FluxVector | None = None):
 
             if flux_vector is None:
                 flux_vector = FluxVector()
-                flux_vector.set_crosstalk_from_bias(crosstalk, qdac_buses_offset)  # type: ignore
+                if target_fluxes:
+                    flux_vector.set_crosstalk(crosstalk)
```
flavie-lebars, line 1037: Claude: set_crosstalk(crosstalk) zero-inits every bus in the crosstalk matrix ([flux_vector.py:98-100](vscode-webview://1o6jv1pp0udulbjrbgujci8u7id21ek3036s2qgsse5p3dt7u53h/src/qililab/qprogram/flux_vector.py#L98-L100)), then only buses present in target_fluxes get overwritten. So with partial tracking (some buses set via FLUX, others not), the untracked ones now get forced to flux=0 instead of falling back to their real hardware bias — a regression versus the old behavior for that mix.

src/qililab/platform/platform.py
```diff
@@ -1124,6 +1124,10 @@ def compile_qprogram(
...
+        target_fluxes = dict(flux_vector.flux_vector) if flux_vector is not None else None
```
flavie-lebars, line 1107: Problem here is if the user doesn't do platform.set_parameter but sets everything via the runcard. Then, flux_vector is not updated

flavie-lebars (changes requested):

src/qililab/qprogram/qprogram.py
jordivallsq: I will add a task for this to take into account this behavior and raise the right Exceptions / Warnings. In reality, the user will either set the Flux via set_parameter or set the bias via the runcard (unlikely but possible).

I agree that it has to be accounted for (I will create a task for this), but it is not a behavior that we will see.

src/qililab/platform/platform.py
jordivallsq: That is why I maintained `qdac_offsets = [float(bus.get_parameter(Parameter.VOLTAGE)) for bus in self.qdac_buses]`, This only works whenever no `set_parameter(FLUX)` is used and takes into account the voltage registered to the runcard. The initial problem was that `set_parameter(FLUX)` applied twice the non-linear aspect of crosstalk. For the case where flux_vector is not updated, this is not the case, and the old implementation is correct.

flavie-lebars: that makes sense, thanks!

flavie-lebars (approved):
Image

jordivallsq:
Tested in HW

PR #1192, opened 2026-08-18
[QHC-1557] RF-switch changes channel when setting LO to bus

elygoner:
Added two things:
- if you have created a bus with *rf-switch* as an instrument and you set the LO on that `Bus`, the *rf-switch* channel will change to that of the `Bus`.
- If you set a channel on the *rf-switch* driver and after set the **SAME** channel, the driver will ignore you. This is so that you don't send unnecessary commands to the *rf-switch* specially when doing a software loop over LO frequency.


changes/1192.feature.md
```diff
@@ -0,0 +1,62 @@
...
+```yml
+name: connect_to_switch
+
+buses:
+- alias: drive_q1
+  instruments: [rs_0, rf_switch]
+  channels: [0, 1]
+- alias: drive_q2
+  instruments: [rs_0, rf_switch]
+  channels: [0, 2]
```
jordivallsq, line 14: What happens if a user tries to use a qprogram with these 2 buses at the same time? Physically this is not possible because both channels use the same rs line and therefore they cannot work at the same time.

We should rise an error

jordivallsq (changes requested):
Left a comment regarding the error raising.
Otherwise, the implementation looks good.

@elygoner, @flavie-lebars regarding the current implementation, it looks good and it works for what we need. But I have a small concern, what happens to the dictionary of LO's that the users are using? Right now there is only one LO registered, the one changed with `set_parameter(LO)`. What hapens to the res, must the user have a list/dictionary in their notebook of all the LO's? Mi proposal changed the runcard more significantly but we also had a place where to save all the LO's mapped to each channel. The implementation is the same with the addition that if the user saves the LO's in the runcard they don't need to use `set_parameter(LO)` at every experiment.

Having said that, the current implementation is fine, works and registers to the database the right information. So if you disagree we can keep it as it is.

elygoner:
@jordivallsq for the execution, right now there isn't really a validation step for busses in the execution.
It is true that to make use of this PR you **HAVE** to make buses with instrument/channel overlap which didn't happen before.  I also raised this, but for now it is deferred since problems with plausible instrument overlap exist since before we even made a driver for the rf_switch.
Therefore, adding a bus validation for qprogram execution should be a different PR.

About the LO, it will have to be saved to parameters on the calibration file or elsewhere. We took that desition since adding those parameters would necessitate altering the rf-switch `Instrument` to store frequencies even if they aren't a parameter on the instrument and that goes against driver_wrapper conventions. We can search for a better solution in the future if it is necessary.

changes/1192.feature.md
```diff
@@ -0,0 +1,62 @@
...
+
+- name: rohde_schwarz
+  alias: rohde_schwarz_controller_0
+  connection:
+    name: tcp_ip
+    address: 192.168.2.148
+  modules:
+  - alias: rs_0
+    slot_id: 0
+  reference_clock: ext
```
flavie-lebars, line 44: out of curiosity, do they ever use ext clock for the RS?

jjmartinezQT:
> Left a comment regarding the error raising. Otherwise, the implementation looks good.
> 
> @elygoner, @flavie-lebars regarding the current implementation, it looks good and it works for what we need. But I have a small concern, what happens to the dictionary of LO's that the users are using? Right now there is only one LO registered, the one changed with `set_parameter(LO)`. What hapens to the res, must the user have a list/dictionary in their notebook of all the LO's? Mi proposal changed the runcard more significantly but we also had a place where to save all the LO's mapped to each channel. The implementation is the same with the addition that if the user saves the LO's in the runcard they don't need to use `set_parameter(LO)` at every experiment.
> 
> Having said that, the current implementation is fine, works and registers to the database the right information. So if you disagree we can keep it as it is.

RF switch as an instrument do not understand frequencies. It is a simple instrument that just open or close channels. That is it, adding extra logic on top complicates it and blurries its purpose. We put it in the LO because they will always set the LO, otherwise there is no mixing. They also have the possibility of changing the channel manually in their notebooks at their will. If you are concerned about a specific case please provide an example so we can evaluate it.

jordivallsq:
@jjmartinezQT One plausible scenario is a notebook where the user needs to change the RF switch constantly, and this will be mapped using a dictionary that maps every bus with an LO frequency (as every drive will require a different LO):

```
lo_mapping = {"bus_a": 6e9, "bus_b": 7e9}

set_parameter(LO, lo_mapping[bus])
...
```

I propose ignoring that requirement by adding this in the runcard and to ease the mental correlation I proposed to add it to the RF switch instrument part.
You are right that the instrument does not work with frequencies but we still attach a frequency to every channel that we use. An honestly I cannot find another place to put this dictionary other than in the user's notebook (and probably in a lot of new qilitools experiments) or the runcard under RF switch instruments.

changes/1192.feature.md
jordivallsq: Please create a task to implement this (leave it in To Be Refined, and someone else can take ownership)

jordivallsq (approved):
Since the comments have been addressed and the error raising will be handled in another PR I am approving these changes as everything else looks good.
Great job!

changes/1192.feature.md
flavie-lebars: I have seen your reply below Ely, but here I very much agree with Jordi. This is new behaviour and I think we should raise an error. It can bve specific to this, and later in another PR we can have a more general function. This is very error prone for the user.

src/qililab/instruments/becker/rswu_sp16tr.py
```diff
@@ -66,6 +66,8 @@ def route(self, channel: str):
                 f"Invalid channel {channel}, valid: Numbers 1 to 16 preceded by one of the following [RF, rf] "
                 "or nothing or X followed by a number from 101 to 116, the two last numbers denoting the channel."
             )
+        if self.settings.active_channel == channel:
+            return
```
flavie-lebars: route() caches the target channel before the hardware call confirms it, so if that call fails, a retry with the same channel now silently no-ops (bc of the cache) instead of re-sending the command — the switch can get stuck un-set with no error.

flavie-lebars (changes requested):
Has this been tested in hardware? Ideally, this would be tested with both Qblox and RS.

jordivallsq:
> @jjmartinezQT One plausible scenario is a notebook where the user needs to change the RF switch constantly, and this will be mapped using a dictionary that maps every bus with an LO frequency (as every drive will require a different LO):
> 
> ```
> lo_mapping = {"bus_a": 6e9, "bus_b": 7e9}
> 
> set_parameter(LO, lo_mapping[bus])
> ...
> ```
> 
> I propose ignoring that requirement by adding this in the runcard and to ease the mental correlation I proposed to add it to the RF switch instrument part.
> You are right that the instrument does not work with frequencies but we still attach a frequency to every channel that we use. An honestly I cannot find another place to put this dictionary other than in the user's notebook (and probably in a lot of new qilitools experiments) or the runcard under RF switch instruments.

Understood, that makes sense. Either way, we have the "MVP" right now. If there is any new need in the future, we can make changes or not. Probably we will see changes in qilitools, or we can even add these `lo_mapping` in other places such as Calibration file. If needed as I said.

changes/1192.feature.md
elygoner: I don't believe so

elygoner:
@flavie-lebars About the validation on the execution I also raised this in the informal reunion we had with @jjmartinezQT and I also agree with Jordi. The explanation I gave is what we ended up deciding for in the reunion. I will implement it.

src/qililab/instruments/becker/rswu_sp16tr.py
flavie-lebars: @elygoner there is still a problem here:
If the device is disconnected, the cache is updated but never set on the instrument at any point. If you have a second call on the instrument with the cached version it does not get updated with a second set_parameter whereas other instruments would have updated.


Adding optimisation like this on drivers, is always more complex than it seems. Is it really needed here?

src/qililab/platform/components/bus.py
```diff
@@ -186,6 +186,24 @@ def set_parameter(
                 raise Exception(f"OutputID {output_id} is not linked to bus with alias {self.alias}")
             self.instruments[0].set_parameter(parameter=parameter, value=value, output_id=bus_output_id)
             return
+
+        if (
```
flavie-lebars, line 190: The logic to find the RSWU_SP16TR instrument+channel on a bus is duplicated here and in the platform. Please refactor to have a shared logic.

src/qililab/platform/platform.py
```diff
@@ -1048,6 +1048,49 @@ def execute_experiment(
...
+        """
+
+        def validate_rswu_sp16tr(
+            match_1: tuple[Instrument, ChannelID | None], match_2: tuple[Instrument, ChannelID | None]
+        ) -> bool:
+            return match_1[0].alias == match_2[0].alias and match_1[1] != match_2[1]
+
+        _instrument_handlers: dict[InstrumentName, Callable] = {InstrumentName.RSWU_SP16TR: validate_rswu_sp16tr}
+
+        def shared_instruments(bus_1: Bus, bus_2: Bus) -> frozenset[str]:
```
flavie-lebars, line 1089: the docstring should say if an error can be returned.

src/qililab/platform/platform.py
```diff
@@ -1048,6 +1048,49 @@ def execute_experiment(
...
+            """
+            return frozenset(
+                instrument_1.alias
+                for instrument_1, channel_1 in zip(bus_1.instruments, bus_1.channels)
+                if instrument_1.name in _instrument_handlers
+                for instrument_2, channel_2 in zip(bus_2.instruments, bus_2.channels)
+                if _instrument_handlers[instrument_1.name]((instrument_1, channel_1), (instrument_2, channel_2))
+            )
+
+        buses_obj: list[Bus] = [self.get_element(bus) for bus in buses]
```
flavie-lebars: instead of `get_element` here, you should use `_get_bus_by_alias`

flavie-lebars (changes requested):
1. For the next instrumentation PR you do please add a test on QHC_hardware_test, it's super convenient.
2. Please fix all the Sonar Cloud trailing comments warnings.

flavie-lebars (approved):
Image

PR #1193, opened 2026-08-19
removal: sequence timeout

flavie-lebars:


src/qililab/instruments/qblox/qblox_adc_sequencer.py
```diff
@@ -47,3 +47,10 @@ def __post_init__(self):
                 FutureWarning,
                 stacklevel=1,
             )
+        if self.sequence_timeout is not None:
+            warnings.warn(
+                "sequence_timeout in the runcard is deprecated and will be removed in a future release. "
+                "It has no effect on the instrument's behavior.",
+                FutureWarning,
+                stacklevel=1,
+            )
```
jordivallsq, line 56: Same comment about the logger.warning (wasn't clear from previous reply), is it a `warnings.warn` because is the only one that can handle `FutureWarning`?

flavie-lebars: i agree, but will do it in a separate PR to unify all of them.

jordivallsq: Ok that was the unclear part, thanks

jordivallsq (approved):
Nice PR, and completely happy with the changes!

elygoner (approved):
If the `Parameter` wasn't used looks good, I find weird the fact that there wasn't a getter for it on the driver but makes sense if it was a dead parameter.

PR #1194, opened 2026-08-19
chore: fix trailing comments

flavie-lebars:


src/qililab/instruments/qblox/qblox_adc_sequencer.py
```diff
@@ -33,7 +33,8 @@ class QbloxADCSequencer(QbloxSequencer):
     scope_store_enabled: bool
     threshold: float
     threshold_rotation: float
-    time_of_flight: int  # nanoseconds
+    # nanoseconds
+    time_of_flight: int
     integration_length: int | None = None
     sequence_timeout: int | None = None  # minutes
     timeout_repetitions: int = 0
```
jordivallsq, line 42: We should make these kind of comments, features of parameters, belonging to a docstring (as Vyron pointed out) instead of defining them simply above.
Because in this example:
```
# nanoseconds
time_of_flight: int
```
gets a bit confusing.

I don't know how to word it but I am sure you can automatize this.

src/qililab/instruments/qblox/qblox_adc_sequencer.py
jordivallsq: also don't know why but it bypassed these minutes here for no apparent reason

src/qililab/instruments/qblox/qblox_adc_sequencer.py
flavie-lebars: That's a bigger change and I think should be done throughout, and a proper docstring to be done.
The scope of this PR is to have a clean git diff.

jordivallsq: Ok

jordivallsq (approved):
The changes look good. Thank you very much!
We need to create a new task for the docstrings

PR #1195, opened 2026-08-19
chore: add time elasped in platform session

flavie-lebars:


src/qililab/platform/platform.py
```diff
@@ -20,6 +20,7 @@
 import ast
 import io
 import re
+import time
```
fedonman: One thing that could be slowing down qililab import is importing  whole modules. Try to refactor to `from A import B,  C` and import what is used. For time, `from time import perf_counter`

src/qililab/platform/platform.py
```diff
@@ -984,6 +985,7 @@ def session(self):
         cleanup_methods = []
         cleanup_errors = []
         execution_error: Exception | None = None
+        start_time = time.time()
```
fedonman: it's better to use `perf_counter()`. `time()` is not reliable.
```suggestion
        start = perf_counter()
```

src/qililab/platform/platform.py
```diff
@@ -1012,6 +1014,9 @@ def session(self):
                     logger.error(f"Error during cleanup: {cleanup_exception}")
                     cleanup_errors.append(cleanup_exception)
 
+            elapsed_time = time.time() - start_time
+            logger.info(f"Platform session took {elapsed_time:.2f} seconds")
```
fedonman: Do we want to just log or return the value too, so it can be used programmatically? I would expect something like:
```
with platform.session() as session:
    ...

print(session.execution_time)
```

I would make the Session an extensible class as more things might be added in the future.

fedonman (changes requested):
Thanks for the ultra quick PR. Left some comments

src/qililab/platform/platform.py
flavie-lebars: well noted thanks, I will modify it here and create a task for the rest of qililab.

src/qililab/platform/platform.py
flavie-lebars: I have changed it, thanks!

src/qililab/platform/platform.py
flavie-lebars: The log was discussed with @NataliaSirera , however I agree with returning the value. It now returns three fields: `execution_time`, `success` and `error`. Let me know what you think.

flavie-lebars:
@NataliaSirera and @jjmartinezQT two points differ from our earlier discussion:
- I ended up starting the timer after `turn_on_instruments`, instead of the beginning of the context manager.
- Following @fedonman suggestion, the value is returned additionally to being logged.

Also FYI, this has been tested on the qblox cluster and works, [test file](https://github.com/qilimanjaro-tech/QHC_hardware_tests/pull/5).

NataliaSirera (approved):
Thank you!!

src/qililab/platform/platform.py
```diff
@@ -947,7 +958,18 @@ def session(self):
...
-            yield
+            start_time = perf_counter()
+            try:
+                yield running_session
+            except Exception as exc:
+                running_session.success = False
+                running_session.error = exc
+                raise
+            else:
+                running_session.success = True
```
fedonman: If `platform.execute_qprogram()` succeeds but `.disconnect()` fails and the user receives an exception, do we consider the execution successful?

fedonman: For me, a successful session is one that runs successfully from start to finish. If any error occurs, including a connection error, it should be stored in .errors.

src/qililab/platform/platform.py
```diff
@@ -947,7 +958,18 @@ def session(self):
...
+            finally:
+                running_session.execution_time = perf_counter() - start_time
+                logger.info(f"Platform session took {running_session.execution_time:.2f} seconds")
```
fedonman, line 981: I find the nested `try-try` a bit weird. Can we try (pun intented) to flatten them? Also, in this case `BaseException` is more correct, as we want to break when `KeyboardInterrupt` event is raised (Ctrl+C). I have a refactor in mind, but touches a bit of the whole method, so I can't comment as suggestion. I could make a commit if that's ok.

fedonman (changes requested):
Thanks Flavie! Left some comments, one is a question we should answer. Other notes relative to this code: We can remove `qililab.exceptions.ExceptionGroup` and do `from builtins import ExceptionGroup` directly, as we are now in 3.11+

src/qililab/platform/platform.py
flavie-lebars: I agree, this is an oversight, I fixed it.

src/qililab/platform/platform.py
flavie-lebars: I have added a commit including all your comments, except this one. I am not sure if you refactor is related to the try-try or not. So this is not fixed for now, but let me know if you want me to fix it.

fedonman: @flavie-lebars I made the refactor and approved the PR. Please review the changes and if you are ok, feel free to merge.

fedonman (approved):
Thank you Flavie! 💯

src/qililab/platform/platform.py
flavie-lebars: @fedonman thank you very much! This is great.
I have one question, is there a reason that clean up catches `Exception`, and not `BaseException`?

fedonman: Good catch, its should be BaseException too. Fixed it.

PR #1197, opened 2026-08-19
chore: remove stale references

flavie-lebars:


jjmartinezQT (approved):
thanks for the effort

PR #1199, opened 2026-08-20
ci: trim the matrices and fix the checks that never fail

### LINKED ISSUE #1198: Trim the CI matrices and fix the checks that never fail
### What needs doing

Run each check once instead of once per Python version, and stop branch protection from naming every matrix leg.

- Cut `code_quality.yml` and `docs.yml` to a single job each.
- Run Tests on the oldest and newest version on PRs, all three on main, and upload coverage once.
- Add a gate job to `tests.yml` so protection can require one context instead of three.
- Give all three workflows a `concurrency` block and `docs.yml` a draft check.
- Make the mdformat step check instead of rewrite, and set the ruff target version from `requires-python`.

### Why

A PR starts 9 jobs, and 6 of them repeat work that does not depend on the interpreter. Code Quality is the clearest case: mypy is pinned to `python_version = "3.13"` in `pyproject.toml` and ruff to `target-version = "py313"`, so all three legs run the same analysis on the same files. Docs builds the same Sphinx output three times. Only Tests really varies per version.

Two steps also cannot fail today. `mdformat .` formats in place rather than checking, so misformatted markdown gets silently rewritten in the runner and the job stays green, and it walks `.venv` while doing it. Nothing cancels superseded runs either, so a second push leaves 9 jobs from the first one running to the end.

fedonman:
Closes #1198.

A PR used to start 9 jobs, and 6 of them repeated work that does not depend on the interpreter: mypy is pinned to `python_version = "3.13"` in `pyproject.toml` and ruff to a target version, so all three Code Quality legs ran the same analysis, and the docs job built the same Sphinx output three times. Both are single jobs now, Tests keeps the oldest and newest version on PRs with the full set on main, and coverage is measured and uploaded once instead of three times. The mdformat step used to format in place, which meant misformatted markdown was silently rewritten in the runner and the job passed anyway, so it now checks instead, skipping the CHANGELOG that towncrier owns. Every workflow also got a concurrency block, since nothing cancelled superseded runs, and the ruff target version now follows `requires-python`. Four jobs on a PR instead of nine.

Branch protection needs its required checks switched to `code-quality`, `docs` and `Tests` when this merges, since the per-version contexts will no longer report.


fedonman:
@flavie-lebars @jjmartinezQT An optimization similar to QiliSDK's https://github.com/qilimanjaro-tech/qilisdk/pull/404 Feel free to propagate the changes to other QHC repositories.

flavie-lebars (approved):
Thanks Vyron! I will make sure it is replicated on the rest of QHC repo.

PR #1201, opened 2026-08-24
qhc-1558-bug-fix-wait-trigger

flavie-lebars:


src/qililab/qprogram/qblox_compiler.py
```diff
@@ -1195,42 +1205,59 @@ def _handle_wait_trigger(self, element: WaitTrigger) -> None:
...
+        # long qpysequence ends up waiting .
+        pending_bus = next((bus for bus in buses if self._buses[bus].upd_param_instruction_pending), None)
+        if pending_bus is not None:
+            duration = self._handle_add_trigger_waits(bus=pending_bus, duration=element.duration, port=port)
+            buses.remove(pending_bus)
+        else:
+            duration = element.duration
+
+        for bus in buses:
+            self._handle_add_trigger_waits(bus=bus, duration=duration, port=port)
```
fedonman: `WaitTriggerUpdParam` puts its `upd_param 4` before the trigger, not after. So the bus that flushes the update spends those 4 ns not yet watching the trigger line, and once the trigger arrives it carries on 4 ns before the other buses.

```python
from qililab.qprogram import QProgram
from qililab.qprogram.qblox_compiler import QbloxCompiler

qp = QProgram()
qp.wait(bus="drive", duration=100)
qp.wait(bus="readout", duration=100)
qp.set_frequency(bus="drive", frequency=1e6)
qp.wait_trigger(bus="drive", duration=100, port=1)

seqs = QbloxCompiler().compile(qprogram=qp, ext_trigger=True).sequences
print(repr(seqs["drive"]._program))    # upd_param 4 ; wait_trigger 1, 96   -> carries on at Ttrig + 96
print(repr(seqs["readout"]._program))  # wait_trigger 1, 100                -> carries on at Ttrig + 100
```

Same 4 ns gap on the long path: with `duration=70000`, drive gets `wait_trigger 1, 65535` + `wait 4461` and readout gets `wait_trigger 1, 65535` + `wait 4465`.

Flushing the update before the sync fixes both.
```suggestion
        # Flush any pending upd_param before the sync, so every bus reaches the trigger aligned and
        # waits on it with the same plain wait_trigger.
        for bus in self._buses:
            if self._buses[bus].upd_param_instruction_pending:
                self._buses[bus].qpy_block_stack[-1].add(component=QPyInstructions.UpdParam(duration=INST_MIN_WAIT))
                self._buses[bus].static_duration += INST_MIN_WAIT
                self._buses[bus].duration_since_sync += INST_MIN_WAIT
                self._buses[bus].upd_param_instruction_pending = False
                self._buses[bus].marked_for_sync = True

        self._handle_sync(element=Sync(buses=None), delay=True)

        port = element.port if element.port else EXT_TRIGGER_ADDRESS
        for bus in self._buses:
            self._handle_add_trigger_waits(bus=bus, duration=element.duration, port=port)
```

fedonman (changes requested):
Thanks a lot Flavie! I think I found a single bug. Otherwise looks great!

src/qililab/qprogram/qblox_compiler.py
flavie-lebars: ah oups, nice catch, thanks!

fedonman (approved):
Thanks Flavie! 🚀 

Image

PR #1203, opened 2026-09-03
[QHC-1594] Fix `InstrumentFactory` 's and `InstrumentControllerFactory` 's  register wrapper to return the correct typing.

elygoner:


flavie-lebars (approved):
Image

PR #1205, opened 2026-09-07
Qhc 1542 crosstalk units change, from Phi_0/V to pH

jordivallsq:
Crosstalk matrices are now interpreted in units of **pH** (mutual inductance) rather than Φ₀/V. `CrosstalkMatrix.flux_to_bias` (and the non-linear override) and `FluxVector.set_crosstalk_from_bias` convert the stored pH matrix to Φ₀/V using each drive line's resistance before applying it.
Every flux line involved must have a resistance set, a `ValueError` is raised otherwise.

- New method `CrosstalkMatrix.in_phi0_per_volt()` returning an equivalent matrix expressed in Φ₀/V.
- New helpers `convert_phi0_per_volt_to_pH`, `convert_pH_to_phi0_per_volt` and the `PHI_0_WB` constant (magnetic flux quantum, in Wb) in `qililab.qprogram.crosstalk_matrix`.


src/qililab/qprogram/crosstalk_matrix.py
elygoner, line 88: For what is said in the changelog, resistances can no longer be None, change the typing and remove the then-unnecessary cast on line 222.

elygoner, line 255: Is this correct?, because now you have an input in Φ₀ that multiplies a pH/pH matrix and gets V. I don't think that is right, don't we have to use the resistences to lower the Φ₀ to pH before doing the calulation. Isn't this also true for the Non linear calculations?

elygoner (changes requested):
I don't think the changes to the implementatin are as easy as this PR makes them seem. Can you confirm my main question before I continue with the review please?

src/qililab/qprogram/crosstalk_matrix.py
jordivallsq: They can no longer be none but if someone leave an empty field it sets them as None (this way it can be filled one by one for example. Therefore, the typing works

jordivallsq: This line transforms the matrix into Phi_0/V (just above your comment)
`inverse = self.in_phi0_per_volt().inverse()`

jordivallsq:
@flavie-lebars has shared two very relevant points:
1. Existing calibration data will raise ValueError at runtime, with no migration path: [changes/1205.feature.md:1](vscode-webview://1o6jv1pp0udulbjrbgujci8u7id21ek3036s2qgsse5p3dt7u53h/qililab/changes/1205.feature.md#L1), [crosstalk_matrix.py:227](vscode-webview://1o6jv1pp0udulbjrbgujci8u7id21ek3036s2qgsse5p3dt7u53h/qililab/src/qililab/qprogram/crosstalk_matrix.py#L227)
resistances was previously dead/unused on CrosstalkMatrix; flux_to_bias now requires it and raises for any missing line. Any pre-existing calibration file (built under the old Φ₀/V convention) has resistances=None for every bus and will start raising ValueError("Missing resistance for line '...'") the first time a FLUX parameter is set or a QProgram compiles (not at load time, but deep in execution). There's no runcard/YAML field for resistances, so the only fix is a code-level set_resistances() call.
2. Satisfying that error doesn't make the data correct, it just hides the real problem, same file
Once a user adds resistances to silence the exception, flux_to_bias proceeds assuming the stored matrix values are already in pH. But existing values were calibrated in Φ₀/V. The changelog documents the two new conversion helpers but never tells users their existing numbers need to be run through convert_phi0_per_volt_to_pH first. Result: no exception, but a silently wrong bias voltage applied to hardware. There's also no version marker in the YAML serialization to distinguish old- vs new-format calibration files.

I am thinking how to handle them but short answer: Oscar and I were aware of these limitations but due to the need of this change we decided to proceed while knowing that some old calibration files will be deprecated.
I will think of a proper reply and solution to those issues.

tests/qprogram/test_qblox_compiler.py
```diff
@@ -1019,7 +1029,7 @@ def variable_expression_two_gains() -> QProgram:
 @pytest.fixture(name="calibration_crosstalk")
 def fixture_calibration_crosstalk() -> Calibration:
     inverse_xtalk_array = np.linalg.inv([[1, 0.5], [0.5, 1]])
-    crosstalk = CrosstalkMatrix().from_array(["flux1", "flux2"], inverse_xtalk_array)
+    crosstalk = _crosstalk_ph(inverse_xtalk_array)
```
elygoner, line 1032: With how much this exact lines repeat, they should be a function or fixture

tests/qprogram/test_qdac_compiler.py
```diff
@@ -423,6 +426,7 @@ def test_crosstalk_compensation(self, qdac: QDevilQDac2, flux1: Bus, flux2: Bus)
         crosstalk = CrosstalkMatrix.from_buses(
             buses={"flux1": {"flux1": 1.0, "flux2": 0.5}, "flux2": {"flux1": 0.1, "flux2": 1.0}}
         )
+        crosstalk.set_resistances({"flux1": _UNIT_RESISTANCE, "flux2": _UNIT_RESISTANCE})
```
elygoner: same with this repeated  lines.

tests/qprogram/test_qprogram.py
```diff
@@ -669,6 +672,7 @@ def test_with_crosstalk_multi_variable_offset(self):
         # Build a 2x2 crosstalk matrix between flux_bus_0 and flux_bus_1
         inverse_xtalk_array = np.linalg.inv([[1, 0.5], [0.5, 1]])
         crosstalk = CrosstalkMatrix().from_array(["flux1", "flux2"], inverse_xtalk_array)
+        crosstalk.set_resistances({"flux1": _UNIT_RESISTANCE, "flux2": _UNIT_RESISTANCE})
```
elygoner: same with this repeated lines

src/qililab/qprogram/crosstalk_matrix.py
elygoner: I don't agree, if you have any resistance as none anything you try to do with the CrosstalkMatrix will error, so resistances shouldn't allow None.

elygoner (changes requested):
Please adres the comments.

src/qililab/qprogram/crosstalk_matrix.py
jordivallsq: If I do that any current code that sets all frequencies sequentially will crash, by leaving it like this it crashed only. If someone forgets a resistance it will crash when using the crosstalk (setting it of starting the experiment, before any interaction with the instruments other than connecting), same as anyone not setting the resistances at all.

What I can do with that is improve the error raising to specify which resistance is missing

tests/qprogram/test_qblox_compiler.py
jordivallsq: it is a function

tests/qprogram/test_qprogram.py
jordivallsq: fixed

tests/qprogram/test_qdac_compiler.py
jordivallsq: fixed

src/qililab/qprogram/crosstalk_matrix.py
elygoner: Cant you do a warning saying what resistances are missing as a list:
```python
warn(f"Resistances for the buses: {tuple(bus for bus in self._sorted_buses() if self.resistances.get(bus) is None)}. \n please add the missing resistances or CrossatalkMatrix will fail upon ay operation")
```

src/qililab/qprogram/crosstalk_matrix.py
```diff
@@ -29,15 +29,15 @@
 def _rescale_by_resistance(
     matrix: dict[str, dict[str, float]], flux_line_resistances_ohms: dict[str, float], *, invert: bool
 ) -> dict[str, dict[str, float]]:
+    missing = sorted({col for cols in matrix.values() for col in cols if flux_line_resistances_ohms.get(col) is None})
```
elygoner: Use the bus sorter

elygoner (approved):
Good enough. I still don't like how the CrostalkMatrix works in general, but fixing that isn't the objective of this PR. I left a comment.

src/qililab/qprogram/crosstalk_matrix.py
jordivallsq: added the warning

src/qililab/qprogram/crosstalk_matrix.py
jordivallsq: done

jordivallsq:
Marked on hold until Oscar gives the OK (after test in the chip). This change will make the change of units mandatory as it uses pH units by default

PR #1206, opened 2026-09-08
[QuickFix] Make db calibration fitting path optional and defaulting to None

jordivallsq:
Made the `path` argument optional in `DatabaseManager.add_calibration_fitting`, so fitting parameters can be attached to an autocalibration measurement without providing a fitting results path.

Seqtante's path is not mandatory while writing in the database.


elygoner (approved):
Image

PR #1208, opened 2026-09-09
[QHC-1609] Add one way numpy scalar serialization

elygoner:
I have repourposed the `compare_platforms` i added to compare platform serialization and renamed it to compare_objects since it is general


flavie-lebars (changes requested):
Image

Add a bugfix in the changelog please.

flavie-lebars (approved):
Image

PR #1210, opened 2026-09-10
Qhc 1593 bug wait trigger on a non default port never receives a trigger event

jordivallsq:
Fixed `wait_trigger` silently compiling to an unreachable trigger address on Qblox. Previously the `port` argument was forwarded straight to the emitted `WaitTrigger`/`LongWaitTrigger` instruction, but the only signals published to the Qblox trigger network are the cluster's external trigger input (mapped to `EXT_TRIGGER_ADDRESS = 15`) and, for `measure_reset`, a QRM's thresholded-acquisition result. Passing any other address (e.g. `port=7`) produced a program that compiled cleanly yet blocked forever on hardware, because nothing ever drove that address.

`QbloxCompiler._handle_wait_trigger` now always waits on `EXT_TRIGGER_ADDRESS`, ignoring the `port` argument on Qblox buses. The `port` argument remains available on `QProgram.wait_trigger` and the `WaitTrigger` operation for the QDACII compiler, which uses it to select an external input-trigger port.

`EXT_TRIGGER_ADDRESS` was previously duplicated as a module-level literal in both `QbloxClusterController` (which publishes to it) and `QbloxCompiler` (which waits on it), so the two could silently diverge — precisely the failure mode of this bug. It is now defined once in `QBLOXCONSTANTS.EXT_TRIGGER_ADDRESS` and imported by both.


jordivallsq:
chnagelog for PR 1206 added here by mistake as it was not added in previous PR merge (the note is minimal and suppose no changes to the actual PR)

flavie-lebars (approved):
Image

PR #1212, opened 2026-09-14
Qhc 1610 set bias to zero does not correctly update qlparameterflux

jordivallsq:
Fixed `Platform.set_bias_to_zero` leaving the tracked `Parameter.FLUX` value stale. The instrument bias was correctly driven to 0 V/A, but the cached `flux_parameter` was never updated, so `get_parameter(alias, Parameter.FLUX)` kept returning the previous (non-zero) value. `set_bias_to_zero` now updates the tracked `Parameter.FLUX` for each bus to the flux that corresponds to zero bias.


jordivallsq:
successfully tested on HW:
https://github.com/qilimanjaro-tech/QHC_hardware_tests/blob/ea4624d834dde48785fc5b93b492b469970fbd2c/tests/qililab/test_PR_1212.ipynb

tests/platform/test_platform.py
```diff
@@ -410,6 +410,35 @@ def test_set_bias_to_zero(self, platform: Platform):
...
+
+        platform.set_bias_to_zero()
+
+        # Instrument output is 0 and the tracked flux is back in sync.
+        assert platform.get_parameter(alias="drive_line_q0_bus", parameter=Parameter.OFFSET_OUT0) == 0.0
+        assert platform.get_parameter(alias="drive_line_q0_bus", parameter=Parameter.FLUX) == 0.0
+
+    def test_set_bias_to_zero_with_flux_offsets_keeps_voltage_zero(self, platform: Platform):
+        """With non-zero flux offsets, voltage must stay 0 and FLUX must track the zero-bias flux."""
+        crosstalk_matrix = CrosstalkMatrix.from_buses(buses={"drive_line_q0_bus": {"drive_line_q0_bus": 0.1}})
```
elygoner: This line is hilarious to me. I understand it, but it might be worth it do test a bit more complex case

elygoner (approved):
Image

tests/platform/test_platform.py
jordivallsq: I left it as it is because it checks the issue at hand but I increased the number of buses to make sense

PR #1213, opened 2026-09-17
Quickfix crosstalk flux vector's play not restarting correctly

jordivallsq:
Fixed crosstalk compensation (`QProgram.with_crosstalk_qblox`) using the wrong flux vector when a flux bus is reused within a program. A repeated `play` on a bus was compensated as if the buses played earlier were still active, distorting the pulse. A reused `play` is now treated as an independent pulse and compensated on its own. Offsets and gains, being persistent DC settings, are instead kept across waits and loops, so the compensation of later elements correctly accounts for the still-active bias of the other buses (e.g. setting one flux bus and then sweeping another in a loop keeps the first bus at its target flux).


jordivallsq:
Tested and validated with @Oriol-DeMiguel

jjmartinezQT (approved):
Good work.

PR #1215, opened 2026-09-18
Qhc 1612 database save flux offsets through stream array

jordivallsq:
Added `flux_offsets` to the `Measurement` database table, saved automatically by `StreamArray` when a calibration with a crosstalk matrix is available. The stored flux vector is the sum of the platform's `flux_vector` and the flux defined by the QProgram's constant `set_offset` calls on crosstalk buses. Only buses offset exactly once by a scalar value are taken from the QProgram; swept offsets and buses offset more than once are ignored.


src/qililab/result/stream_results.py
```diff
@@ -255,6 +259,45 @@ def _get_offsets(self) -> dict[str, float] | None:
...
+        crosstalk_buses = set(self.calibration.crosstalk_matrix.matrix.keys())
+
+        offsets: dict[str, list[float | Variable]] = {}
+
+        def traverse(block: Block) -> None:
+            for element in block.elements:
+                if isinstance(element, SetOffset) and element.bus in crosstalk_buses:
+                    offsets.setdefault(element.bus, []).append(element.offset_path0)
+                elif isinstance(element, Block):
+                    traverse(element)
```
elygoner, line 290: On the QDAC... is the AWS flux also additive?

src/qililab/result/stream_results.py
```diff
@@ -255,6 +259,45 @@ def _get_offsets(self) -> dict[str, float] | None:
...
+
+        traverse(self.qprogram._body)  # type: ignore[union-attr]
```
elygoner: What is this type ignore doing here, the `StructuredProgram._body` is typed as `Block`
```python
class StructuredProgram:
    def __init__(self) -> None:
        self._body: Block = Block()
        ...
```

tests/result/test_stream_results.py
```diff
@@ -619,6 +620,127 @@ def test_add_fitting_autocalibration(self, stream_array: StreamArray):
...
+        stream_array.qprogram = _qprogram_with_offsets([("flux_q0", 0.15), ("flux_q1", 0)])
+
+        assert stream_array._get_qprogram_flux_vector() == {"flux_q0": 0.15, "flux_q1": 0.0}
+
+    def test_qprogram_flux_vector_reads_nested_blocks(self, stream_array: StreamArray):
+        """Offsets set inside nested blocks (loops) are collected."""
+        stream_array.calibration = _calibration_with_crosstalk(["flux_q0", "flux_q1"])
+        qprogram = QProgram()
+        qprogram.set_offset("flux_q0", 0.15)
+        with qprogram.for_loop(variable=qprogram.variable("t", Domain.Time), start=0, stop=4, step=1):
```
elygoner: use `Averages` instead of a `ForLoop` it is also a block and has way less parameters.

src/qililab/result/stream_results.py
```diff
@@ -255,6 +259,45 @@ def _get_offsets(self) -> dict[str, float] | None:
...
+        return flux_offsets
+
+    def _get_qprogram_flux_vector(self) -> dict[str, float]:
+        """Flux vector from the QProgram's constant ``set_offset`` calls on crosstalk buses.
+
+        Returns empty if any bus is offset more than once; swept offsets are skipped.
+        """
+        if not self.calibration or not self.calibration.crosstalk_matrix:
+            return {}
+        crosstalk_buses = set(self.calibration.crosstalk_matrix.matrix.keys())
```
elygoner, line 281: Why do you only get the crosstalk buses?
you can just `StructuredProgram._buses`

elygoner (changes requested):
As I said, I don't like saving this to the database because it is just saving information that your database already contains. I would much rather have a function that lets you get it on `Measurement`.

src/qililab/result/stream_results.py
jordivallsq: yes, you can set it as set_offset (although it works exactly as platform set_parameter, it was created for consistency)

src/qililab/result/stream_results.py
jordivallsq: because this way any other bus used to set offset at the beginning of a qprogram (Readout or drive) will not be tangled in this function

tests/result/test_stream_results.py
jordivallsq: applied

src/qililab/result/stream_results.py
jordivallsq: applied

src/qililab/result/stream_results.py
elygoner: Then the answer is No, it is not additive. Is that taken into account?

src/qililab/result/stream_results.py
elygoner: still you are excluding buses.

jordivallsq:
I tried to implement it inside Measurement but the platform save does not hold the flux_vector, and therefore we should save it somewhere otherwise, implying that we had to change the platform's to_dict or alter the database.

elygoner (approved):
LGTM

PR #1216, opened 2026-09-18
Qhc 1613 set a trigger to enabledisable non linear crosstalk terms

jordivallsq:
Added a toggle to enable or disable the nonlinear crosstalk correction terms of a `NonLinearCrosstalkMatrix` without discarding its stored parameters.

- `NonLinearCrosstalkMatrix.set_non_linear(enabled=True)` flips the new `non_linear_enabled` flag. While disabled, `flux_to_bias` (and `get_non_linear_flux_terms`) skip the non-linear corrections, so the matrix behaves as a purely linear crosstalk matrix.
- `Calibration.set_non_linear_crosstalk(enabled=True)` toggles it on the stored crosstalk matrix. It raises `ValueError` if no crosstalk matrix is set, and logs a warning (leaving the matrix untouched) if the stored matrix is a linear `CrosstalkMatrix`.


changes/1216.feature.md
```diff
@@ -0,0 +1,4 @@
+Added a toggle to enable or disable the nonlinear crosstalk correction terms of a `NonLinearCrosstalkMatrix` without discarding its stored parameters.
```
jjmartinezQT, line 1: This code do not act on the AC crosstalk matrix, only in the DC one, is it intended or the AC one needs to be subjected to the toggle as well?

jjmartinezQT (changes requested):

changes/1216.feature.md
jordivallsq: Very good point, it was intended for the DC crosstalk but it only makes sense to add this in the AC crosstalk as well, I will implement the change

jordivallsq: added AC crossatlk in the trigger

